PR #8, opened 2018-05-18
Added generate_design_matrix function

Vaibhavdixit02:


src/morris_sensitivity.jl
```diff
@@ -0,0 +1,28 @@
+function generate_design_matrix(prob::DEProblem,param_bounds,p)
+    init_param_vector = prob.p
+    B = Array{Array{Float64}}(length(prob.p)+1)
+    r = Float64[]
+    for i in 1:length(param_bounds)
+        dels = [j/p for j in param_bounds[i][1]:param_bounds[i][2]]
+        del_i = rand(dels)
+        push!(r,init_param_vector[i]+del_i)
```
ChrisRackauckas: shouldn't these be del changes off of the previous vector, so `r[i-1]`?

src/morris_sensitivity.jl
```diff
@@ -0,0 +1,28 @@
...
+        # println(r)
+    end
+
+    for i in 1:(length(prob.p)+1)
+        j = rand(1:length(prob.p))
+        r[j] = r[j]+ 1/p
```
ChrisRackauckas: what's this part doing?

src/morris_sensitivity.jl
```diff
@@ -0,0 +1,28 @@
...
+    init_param_vector = prob.p
+    B = Array{Array{Float64}}(length(prob.p)+1)
+    r = Float64[]
+    for i in 1:length(param_bounds)
+        dels = [j/p for j in param_bounds[i][1]:param_bounds[i][2]]
+        del_i = rand(dels)
+        push!(r,init_param_vector[i]+del_i)
+        # println(r)
+    end
+
```
ChrisRackauckas: isn't r a sufficient representation of B?

src/morris_sensitivity.jl
Vaibhavdixit02: Don't get this?

src/morris_sensitivity.jl
Vaibhavdixit02: For the same `r` we can have different trajectories, so I don't think so?

src/morris_sensitivity.jl
Vaibhavdixit02: This is the "select X(3) such that it differs from X(2) for only one component j ̸= i by ±∆" step, should add 1/-1 there actually.

src/morris_sensitivity.jl
Vaibhavdixit02: Oh I got it now, no `del` are the unit of change we can do to each parameter, so say for some parameter in the `n`th vector of parameters the difference from `n+1`th would be `del` so that isn't dependent on `r` in any way, this should perhaps be the same for all parameters though, I was doubtful about that.

src/morris_sensitivity.jl
```diff
@@ -0,0 +1,38 @@
+using DifferentialEquations
```
ChrisRackauckas: shouldn't using the whole thing

src/morris_sensitivity.jl
```diff
@@ -0,0 +1,38 @@
...
+        end
+        if flag == 0
+            all_idxs[i] = copy(indices)
+            i = i+1
+        end
+    end
+
+    B = Array{Array{Float64}}(k)
+    for j in 1:k
+        cur_p = [ps[u][Int(all_idxs[j][u])] for u in 1:length(p_range)]
```
ChrisRackauckas: Is the `Int` necessary?

src/morris_sensitivity.jl
Vaibhavdixit02: Yeah, was there for the example.

src/morris_sensitivity.jl
```diff
@@ -0,0 +1,38 @@
...
+    i = 2
+    while i <= k
+        flag = 0
+        indices[rand(1:length(p_range))] += (rand() < 0.5 ? -1 : 1)
+        for j in indices
+            if j > k || j < 1.0
+                flag = 1
+            end
+        end
+        if flag == 0
```
ChrisRackauckas: is this how rejection is supposed to be done?

src/morris_sensitivity.jl
Vaibhavdixit02: Yeah it was throwing invalid index error.

src/morris_sensitivity.jl
Vaibhavdixit02: There wasn't any method mentioned in the literature.

src/morris_sensitivity.jl
ChrisRackauckas: just use the packages you need.

src/morris_sensitivity.jl
ChrisRackauckas: I guess it's safe. But it's because you made `indices` a float array. I guess it's better to have "safety mode on".

src/morris_sensitivity.jl
ChrisRackauckas: I see, so you're just rejecting any step that goes over the boundary? Thinking about it more, it's unbiased.

ChrisRackauckas:
Just fix up the test setup `using` and it's good.

PR #11, opened 2018-05-22
Changes in generating matrix and added sample_matrices

Vaibhavdixit02:


src/morris_sensitivity.jl
```diff
@@ -1,15 +1,23 @@
-function generate_design_matrix(prob::DEProblem,p_range,p_steps,k = 10)
+struct Matrix
```
ChrisRackauckas: probably not a good idea to shadow this

src/morris_sensitivity.jl
```diff
@@ -1,15 +1,23 @@
...
+    mat::Array{Array{Float64}}
+    spread::Float64
```
ChrisRackauckas: I don't think there's a need to store it in here.

src/morris_sensitivity.jl
ChrisRackauckas: It probably would be easier to just use a VectorOfArray from RecursiveArrayTools? Then you'll get all of the indexing and conversion features for free.

src/morris_sensitivity.jl
```diff
@@ -1,15 +1,23 @@
...
-    all_idxs[1] = copy(indices)
-    i = 2
+    i = 1
     while i <= k
         flag = 0
         indices[rand(1:length(p_range))] += (rand() < 0.5 ? -1 : 1)
-        for j in indices
-            if j > k || j < 1.0
+        for j in 1:length(indices)
+            if indices[j] > p_steps[j]
```
ChrisRackauckas: This is making it way too complicated. You only change one index each step. You only need to check if the one index went out of bounds, and if it did, move it 2 in the other direction.

src/morris_sensitivity.jl
Vaibhavdixit02: This is actually to store the matrix and the spread, I didn't understand what you meant by shadow here?

src/morris_sensitivity.jl
```diff
@@ -1,15 +1,23 @@
...
+end
+
+function generate_design_matrix(prob::DEProblem,p_range,p_steps;k = 10)
     ps = [linspace(p_range[i][1],p_range[i][2],p_steps[i]) for i in 1:length(p_range)]
     indices = [rand(1:i) for i in p_steps]
     all_idxs = Vector{typeof(indices)}(k)
-    all_idxs[1] = copy(indices)
-    i = 2
+    i = 1
     while i <= k
```
ChrisRackauckas: If the index reflecting is done right, this should do one step each time, so it can be a for loop

src/morris_sensitivity.jl
ChrisRackauckas: Shadowing is when you override a name from Base, making it unavailable without prefixing. `Matrix` is defined and exported in Base Julia. This won't warn, but it will error if someone tries to use `Matrix` the normal way.

src/morris_sensitivity.jl
```diff
@@ -24,4 +32,24 @@ function generate_design_matrix(prob::DEProblem,p_range,p_steps,k = 10)
         B[j] = cur_p
     end
     B
+end
+
+function calculate_spread(matrix::Array{Array{Float64}})
```
ChrisRackauckas: Just leave this duck typed. It'll fail anyways since it should be `Array{<:Array{Float64}}` since `Array` is abstract without a dimension.

src/morris_sensitivity.jl
```diff
@@ -1,21 +1,23 @@
-function generate_design_matrix(prob::DEProblem,p_range,p_steps,k = 10)
+struct MatSpread
+    mat::Array{Array{Float64}}
```
ChrisRackauckas: This needs dimensions to be inferred.

PR #12, opened 2018-05-24
Added MorrisGlobalSensitivity

Vaibhavdixit02:


src/morris_sensitivity.jl
```diff
@@ -36,14 +36,45 @@ function calculate_spread(matrix)
...
+            for k in 1:length(del)
+                if abs(del[k]) > 0
+                    change_index = k
+                    break
+                end
+            end
+            del = sum(del)
+            prob1 = remake(prob;p=i[j+1])
+            prob2 = remake(prob;p=i[j])
+            y1 = solve(prob1,alg;saveat=t)
```
ChrisRackauckas: you don't need to solve twice since the previous one has already been done.

src/morris_sensitivity.jl
```diff
@@ -36,14 +36,45 @@ function calculate_spread(matrix)
...
+            push!(effects[change_index],elem_effect)
+        end
+    end
+    means = Float64[]
+    variances = Float64[]
+    for i in effects
+        push!(means,mean(i))
+        push!(variances,var(i))
+    end
+    means,variances
```
ChrisRackauckas: return a struct with means, variances, effects.

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,29 @@
+{
+    "configurations": [
```
ChrisRackauckas: ?

ChrisRackauckas:
add tests and its good to go.

test/morrismethod.jl
```diff
@@ -6,6 +6,7 @@ f = @ode_def LotkaVolterra begin
 end a b c
   
 p = [1.5,1.0,3.0]
+t = collect(linspace(0,10,200))
 prob = ODEProblem(f,[1.0;1.0],(0.0,10.0),p)
-m = sample_matrices(prob,[[0,10],[0,10],[0,10]],[10,10,10],simulations=100)
+m = MorrisGlobalSensitivity(prob,Tsit5(),t,[[1,2],[1,2],[2,4]],[10,10,10],simulations=100)
 println(m)
```
Vaibhavdixit02: This is the test?

ChrisRackauckas: do we have some indication that it's right?

Vaibhavdixit02:
https://julialang.slack.com/archives/D8LVBCHBP/p1527212971000035?thread_ts=1527212775.000246&cid=D8LVBCHBP I misunderstood this.

Vaibhavdixit02:
Can you clear up what you meant there?

ChrisRackauckas:
Instead of embedding that the functionality has to be used for solving an ODE and getting its timeseries, you can let the user give a function that is p -> y(p) that defaults as p -> timeseries of ODE solution

Vaibhavdixit02:
I am sorry I can't quite picture how that would be, I don't think it's possible the ODE is quite inherently involved in the implementation here, maybe some mock example of how you expect it's interface should be would help me clear it up.

ChrisRackauckas:
ODEs are not involved in the implementation at all. The ODE part in the loop is just:

```julia
prob1 = remake(prob;p=i[j+1]) 
y1 = solve(prob1,alg;saveat=t)
```

But that's really a function from parameters to an array.

test/morrismethod.jl
```diff
@@ -4,24 +4,12 @@ A = reshape([1,0,2,3],2,2)
 function f(p)
     A*p
 end
-m = DiffEqSensitivity.morris_sensitivity(f,[[1,5],[1,5]],[10,10],k =1500,simulations=1000,r=150)
+m = DiffEqSensitivity.morris_sensitivity(f,[[1,5],[1,5]],[10,10],length_of_trajectory=1500,simulations=1000,number_of_trajectories=150)
```
ChrisRackauckas: then what's simulations?

PR #14, opened 2018-06-15
Sobol Method implementation

Vaibhavdixit02:


test/chemical_model.jl
```diff
@@ -0,0 +1,20 @@
+using DiffEqSensitivity, OrdinaryDiffEq, ParameterizedFunctions
```
ChrisRackauckas: This should go to diffeqproblemlibrary (with the analytical solution)

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,39 @@
...
+    f = function (p)
+        prob1 = remake(prob;p=p)
+        Array(solve(prob1,alg;saveat=t))
+    end
+
+    calc_mean_var(f,p_range,p_steps,N)
+end
+
+function calc_mean_var(f,p_range,p_steps,N)
+    p_values = [collect(linspace(p_range[i][1],p_range[i][2],p_steps[i])) for i in 1:length(p_range)]
```
ChrisRackauckas: you shouldn't need to build a set of parameter values

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,39 @@
...
+    p = [rand(p_values[j]) for j in 1:length(p_values)]
```
ChrisRackauckas: should be continuous variation

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,39 @@
...
+        y0 = [[0.0 for j in 1:size(y1)[1]] for i in 1:size(y1)[2]]
+        v = [[0.0 for j in 1:size(y1)[1]] for i in 1:size(y1)[2]]
+    else
+        y0 = [0.0 for j in 1:size(y1)[1]]
+        v = [0.0 for j in 1:size(y1)[1]]
+    end
+    for i in 1:N
+        p = [rand(p_values[j]) for j in 1:length(p_values)]
+        y1 = f(p)
+        if length(size(y1)) != 1
```
ChrisRackauckas: why?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,39 @@
...
+                v[j] .+= (y1[j]).^2
+            end
+        else
+            y0 .+= y1
+            v .+= y1.^2
+        end
+    end
+    y0 = y0./N
+    y0_sq = [i.^2 for i in y0]
+    v = v./N .- y0_sq
```
ChrisRackauckas: v's not right. You never do the second sample.

test/chemical_model.jl
Vaibhavdixit02: Will move it once the method is ready, currently using it as quick tests for implementation.

Vaibhavdixit02:
@ChrisRackauckas ping.

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,109 @@
+function give_rand_y(f,p_range,p_fixed=nothing,indices=nothing)
+    if p_fixed == nothing
+        p = [rand(p_range[j][1]:1e-6:p_range[j][2]) for j in 1:length(p_range)]
```
ChrisRackauckas: why are these choices discrete?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,109 @@
...
+    end
+    for i in 1:N
+        y1 = give_rand_y(f,p_range)
+        y0 .+= y1
+        y1 = give_rand_y(f,p_range)
+        v .+= y1.^2
+    end
+    y0 = y0./N
+    y0_sq = [i.^2 for i in y0]
+    v = v./N .- y0_sq
```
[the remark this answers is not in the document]
Vaibhavdixit02: Yeah that is how it is defined, this calculation is mentioned explicitly in the thesis.

src/sobol_sensitivity.jl
Vaibhavdixit02: I expected `rand(p_range[j][1]:p_range[j][2])` to work as a continuous interval but it wasn't correct, what would be the syntax for that?

ChrisRackauckas: `(p_range[j][2]-p_range[j][1])*rand() + p_range[j][1]`

src/sobol_sensitivity.jl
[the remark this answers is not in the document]
ChrisRackauckas: okay.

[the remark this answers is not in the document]
Vaibhavdixit02: But this is giving significantly poor results (variations in the output in the order of `1e-1`) in my runs locally, should we change it?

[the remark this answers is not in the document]
Vaibhavdixit02: With `N` being of the order `1e5`

ChrisRackauckas:
Which paper are you working from?

Vaibhavdixit02:
@ChrisRackauckas I primarily used the UCL thesis as reference.

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+            else
+                p[i] = (p_range[i][2] -p_range[i][1])*rand() + p_range[i][1]
+            end
+        end
+    end
+    p
+end
+
+function calc_mean_var(f,p_range,N)
+    y1 = give_rand_y(f,p_range)
```
ChrisRackauckas: why is this here?

ChrisRackauckas: where is `give_rand_y` defined?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+        y0 = reshape(y0,size(y1)[1],size(y1)[2])
+        v = reshape(v,size(y1)[1],size(y1)[2])
+    end
+    for i in 1:N
+        y1 = give_rand_y(f,p_range)
+        y0 .+= y1
+        y1 = give_rand_y(f,p_range)
+        v .+= y1.^2
+    end
+    y0 = y0./N
```
ChrisRackauckas: `@.` this whole thing.

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+    y0_sq = [i.^2 for i in y0]
+    v = v./N .- y0_sq
```
ChrisRackauckas: `@.` this whole thing.

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+    y0 = [0.0 for j in 1:length(y1)]
+    v = [0.0 for j in 1:length(y1)]
+    if length(size(y1)) != 1
+        y0 = reshape(y0,size(y1)[1],size(y1)[2])
+        v = reshape(v,size(y1)[1],size(y1)[2])
+    end
+    for i in 1:N
+        y1 = give_rand_y(f,p_range)
+        y0 .+= y1
+        y1 = give_rand_y(f,p_range)
```
ChrisRackauckas: why use a different one?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+    y0 = y0./N
+    y0_sq = [i.^2 for i in y0]
+    v = v./N .- y0_sq
+    y0,v
+end
+
+function first_order_var(f,p_range,N,y0)
+    ys = []
+    for i in 1:length(p_range)
+        y = [0.0 for i in 1:length(y0)]
```
ChrisRackauckas: `zeros(y0)`?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+                p[i] = (p_range[i][2] -p_range[i][1])*rand() + p_range[i][1]
+            end
+        end
+    end
+    p
+end
+
+function calc_mean_var(f,p_range,N)
+    y1 = give_rand_y(f,p_range)
+    y0 = [0.0 for j in 1:length(y1)]
```
ChrisRackauckas: `zeros(y1)`?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
+function give_rand_p(f,p_range,p_fixed=nothing,indices=nothing)
+    if p_fixed == nothing
+        p = [(p_range[j][2] -p_range[j][1])*rand() + p_range[j][1] for j in 1:length(p_range)]
+    else
+        p = [0.0 for i in p_range]
```
ChrisRackauckas: `zeros(p_range)`?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+    ys = []
+    for i in 1:length(p_range)
+        y = [0.0 for i in 1:length(y0)]
+        if length(size(y0)) != 1
+            y = reshape(y,size(y0)[1],size(y0)[2])
+        end
+        for j in 1:N
+            p2 = give_rand_p(f,p_range)
+            p1 = give_rand_p(f,p_range,[p2[i]],[i])
+            yer = Array(f(p1)) .* Array(f(p2))
```
ChrisRackauckas: the mean/var calculations should utilize these same `f` calls. It should just be one function with some options to switch the sampling on/off for second order moments, but you can re-use a lot of computations here.

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,144 @@
...
+            y .+= yer
+        end
+        y = y./N - y0.^2
+        push!(ys,copy(y))
+    end
+    ys
+end
+
+function sobol_sensitivity(f,p_range,N,order=0)
```
ChrisRackauckas: default to second order.

ChrisRackauckas: no. The total effects are far too expensive in most cases so they aren't really ever used. Sobol is useful over things like Morris since it can get the second order cases, so it's usually used because of its ability to get the second order interactions only. The total variance estimate then gives you a measure for how much the higher order interactions matter, which is usually low and justifies only doing the second order.

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,143 @@
...
+    ys
+end
+
+function second_order_var(f,p_range,N,y0)
+    ys = []
+    for i in 1:length(p_range)
+        for j in i+1:length(p_range)
+            y = zeros(length(y0))
+            if length(size(y0)) != 1
+                y = reshape(y,size(y0)[1],size(y0)[2])
```
ChrisRackauckas: why is this needed?

src/sobol_sensitivity.jl
```diff
@@ -0,0 +1,143 @@
...
+            @. y += yer
+        end
+        y = @. y/N - y0^2
+        push!(ys,copy(y))
+    end
+    ys
+end
+
+function second_order_var(f,p_range,N,y0)
+    ys = []
```
ChrisRackauckas: why not make this a matrix of the right size?

ChrisRackauckas:
Just needs a few minor changes.

ChrisRackauckas:
I'm merging since this is pretty much complete and improvements can happen in a further PR.

src/sobol_sensitivity.jl
Vaibhavdixit02: It was for the consistency across first order, second order and total order implementations, each would have to be initialised to be of different sizes that would require some calculation. Does this affect performance, I guess fixed size matrix will offer some sort of compiler optimisation, right?

src/sobol_sensitivity.jl
Vaibhavdixit02: For the solutions of DiffEqs?

src/sobol_sensitivity.jl
ChrisRackauckas: This effects performance in two ways. One, growing one-by-one is slower than making it once (though it's not so bad). But two, this is completely untyped so any access is not type inferred. If `f` is costly enough this doesn't matter though, but we might as well try.

src/sobol_sensitivity.jl
ChrisRackauckas: if you do `zeros(y0)` then you don't need to reshape. Why did you do `zeros` of the length?

PR #16, opened 2018-06-25
Added standard regression and Pearson correlation coefficient

Vaibhavdixit02:


src/regression_sensitivity.jl
```diff
@@ -0,0 +1,100 @@
...
+        params = []
+        for j in 1:length(p_range)
+            if i != j
+                push!(params,p_fixed[j])
+            else
+                push!(params,0)
+            end
+        end
+        xis = []
+        yis = Array{Float64}[]
```
ChrisRackauckas: this typing can be incorrect.

src/regression_sensitivity.jl
```diff
@@ -0,0 +1,100 @@
...
+        srcs = []
+        params = []
+        for j in 1:length(p_range)
+            if i != j
+                push!(params,p_fixed[j])
+            else
+                push!(params,0)
+            end
+        end
+        xis = []
```
ChrisRackauckas: This is untyped

src/regression_sensitivity.jl
```diff
@@ -0,0 +1,100 @@
...
+            formula *= "x$j +"
+        end
+        formula = formula[1:end-1] * ")"
+        ols_x = lm(eval(parse(formula)),df)
+        x_cap = x_lm .- predict(ols_x)
+        y_caps = []
+        for k in 1:size(yis)[1]
+            y_cps = []
+            for j in 2:size(yis)[2]
+                df_arr = Array{Float64}[Float64[] for i in 1:length(yis[k,j,:])]
```
ChrisRackauckas: Again many assumptions here, and this isn't strictly typed anyways.

ChrisRackauckas:
Needs tests. Test against a linear function and make sure it gets the analytical values.

src/regression_sensitivity.jl
Vaibhavdixit02: The typing was required for the regression model 😕

src/regression_sensitivity.jl
Vaibhavdixit02: There were problems in the regression model creation without the typing.

ChrisRackauckas:
I'll merge and we can continue when you're ready.

PR #17, opened 2018-07-04
Changes in matrix initialization and test file

Vaibhavdixit02:


ChrisRackauckas:
I updated the test structure (I think the previous PR forgot to actually add the tests to the runtests? I missed that haha). Now it runs but with an error. I'll push out the post which says this is released anyways since you're really close.

Vaibhavdixit02:
Yeah, really sorry about that, it slipped out of my attention too!

PR #27, opened 2018-10-03
Adjust buffer length

YingboMa:


ChrisRackauckas:
Can you add a test for what's finding these sizing issues? I assume it's something with more parameters than states?

PR #28, opened 2018-10-09
RFC: Jacobian vector product via seeding

YingboMa:


src/local_sensitivity.jl
```diff
@@ -70,11 +80,15 @@ function ODELocalSensitivityProblem(f::DiffEqBase.AbstractODEFunction,u0,
                                     alg = SensitivityAlg();
                                     callback=CallbackSet(),mass_matrix=I)
   isinplace = DiffEqBase.isinplace(f)
+  isautojacvec = get_jacvec(alg)
```
ChrisRackauckas: This is dependent on alg? I thought this would be a user-settable switch which defaults to true. We should always try to use the AD seeded version here since it should be much more efficient.

src/local_sensitivity.jl
```diff
@@ -70,11 +80,15 @@ function ODELocalSensitivityProblem(f::DiffEqBase.AbstractODEFunction,u0,
...
   p == nothing && error("You must have parameters to use parameter sensitivity calculations!")
-  uf = DiffEqDiffTools.UJacobianWrapper(f,tspan[1],p)
+  uf = isautojacvec ? nothing : DiffEqDiffTools.UJacobianWrapper(f,tspan[1],p)
   pf = DiffEqDiffTools.ParamJacobianWrapper(f,tspan[1],copy(u0))
   if DiffEqBase.has_jac(f)
     jac_config = nothing
+  elseif isautojacvec
+    jac_config = ForwardDiff.Dual{:___jac_tag}.(u0,u0)
+    jac_config = jac_config, similar(jac_config)
```
ChrisRackauckas: ?????

src/local_sensitivity.jl
ChrisRackauckas: Oh nevermind, I see where it's passed

src/derivative_wrappers.jl
```diff
@@ -1,6 +1,6 @@
-struct SensitivityAlg{CS,AD,FDT} <: DiffEqBase.DEAlgorithm end
-Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central})
-  SensitivityAlg{chunk_size,autodiff,typeof(diff_type)}()
+struct SensitivityAlg{CS,AD,FDT,Jv} <: DiffEqBase.DEAlgorithm end
+Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central},autojacvec=false)
```
ChrisRackauckas: I think we should default to true here? I think the timings will show you almost always want to do this.

ChrisRackauckas:
Yeah that looks just about right. Make sure to do the one for adjoints as well.

YingboMa:
For consistency sake, I named `isautojacvec` for both forward and reverse sensitivity analysis, but it might be very confusing. Maybe we should call it `isautopullback`? @ChrisRackauckas

YingboMa:
Currently, `vec'Jac` via seeding isn't optimized. Hopefully, when I look deeper into `ReverseDiff.jl`, I can come with a way to cache the tape. @ChrisRackauckas could you review this PR? I will make another PR to optimize adjoint sensitivity analysis.

ChrisRackauckas:
Looks good but tests fail?

YingboMa:
That is because of https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/pull/40

src/adjoint_sensitivity.jl
```diff
@@ -139,7 +150,8 @@ function AdjointSensitivityIntegrand(sol,adj_sol,alg=SensitivityAlg())
   p = sol.prob.p
   y = similar(sol.prob.u0)
   λ = similar(adj_sol.prob.u0)
-  pf = DiffEqDiffTools.ParamJacobianWrapper(f,tspan[1],copy(y))
+  # we need to alias `y`
```
ChrisRackauckas: why?

YingboMa: Because we need the current `y` to compute Jacobian w.r.t. params, and we never write into `y` inside `ParamJacobianWrapper` so it is fine to alias.

PR #30, opened 2018-11-20
Optimize adjoint sensitivity with v'J seeding

YingboMa:
HUGE thanks to @jrevels, his following two gists made this possible.
https://gist.github.com/jrevels/c165ed338cc7159085238aa54a763fe2
https://gist.github.com/jrevels/664e2926c01abb15ac6d92fd4a4788c8


ChrisRackauckas:
Does this constrain what derivative functions can be used? We'll need to document that.

ChrisRackauckas:
Make a note in the paper that without compiling the tape it wasn't nearly as efficient, and just grabbing the traditional Jacobian was more efficient.

PR #34, opened 2018-12-04
Integrate the integrand within the adjoint ODE

YingboMa:


YingboMa:
The failure on Julia 1.0 was due to timeout, but tests passed on Julia nightly.

ChrisRackauckas:
oh yes, missed that. I am not sure why the tests have been doing that...

PR #37, opened 2019-01-11
Add backsolve adjoint

YingboMa:


src/derivative_wrappers.jl
```diff
@@ -1,6 +1,7 @@
-struct SensitivityAlg{CS,AD,FDT,Jv,Quad} <: DiffEqBase.DEAlgorithm end
-Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central},autojacvec=true,quad=true)
-  SensitivityAlg{chunk_size,autodiff,typeof(diff_type),autojacvec,quad}()
+struct SensitivityAlg{CS,AD,FDT,Jv,Quad,BS} <: DiffEqBase.DEAlgorithm end
+Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central},autojacvec=true,quad=true,backsolve=false)
+  backsolve && (quad = false)
+  SensitivityAlg{chunk_size,autodiff,typeof(diff_type),autojacvec,quad,backsolve}()
```
ChrisRackauckas: Does backsolve need to be type information?

src/adjoint_sensitivity.jl
```diff
@@ -45,17 +45,28 @@ end
...
 function (S::ODEAdjointSensitvityFunction)(du,u,p,t)
+  idx = length(S.y)
   y = S.y
-  S.sol(y,t)
-  if isquad(S.alg)
-    λ     = u
-    dλ    = du
-  else
-    idx   = length(y)
+  if isbcksol(S.alg)
```
ChrisRackauckas: Maybe clean this up through dispatch? Or is there a lot of shared code?

YingboMa: Yeah, and even if I use dispatch, I still need to check `isbcksol` and `isquad` to assign variables.

src/derivative_wrappers.jl
YingboMa: No, it doesn't. I will move `autojacvec,quad,backsolve` out.

PR #38, opened 2019-01-17
using Flux

MikeInnes:
This is an experimental branch using Flux's AD to derive the integrand in the adjoint method. The main aim is that we can eventually support Flux's style of implicit parameters, and therefore support e.g. neural nets inside of differential equations.

There are a couple of tangential benefits: it's significantly simpler, since we can just call `gradient` rather than passing configs and function wrappers around, and as a result of that I accidentally fixed a bug in which functions depending on `t` were derived incorrectly. It also reveals that we're (probably) taking gradients twice redundantly, so fixing that should get us a nice speedup.

I've also included some simple typo fixes etc. but can split those out if need be.


ChrisRackauckas:
>It also reveals that we're (probably) taking gradients twice redundantly, so fixing that should get us a nice speedup.

How so? Before it was just doing the gradient with respect to the parameters, and the Jacobian w.r.t. the state.

ChrisRackauckas:
Can we run the benchmarks of https://github.com/UMCTM/sensitivity_analysis_paper to see how this differs? With ReverseDiff it was inefficient to not do the tape compilation here.

src/adjoint_sensitivity.jl
```diff
@@ -261,8 +252,12 @@ function (S::AdjointSensitivityIntegrand)(out,t)
     end
     mul!(out',λ,S.pJ)
   else
-    tape = S.paramjac_config
-    vecjacobian!(out, λ, tape, y, S.p)
+    _, back = Tracker.forward(y, S.p) do u, p
+      out_ = map(zero, u)
```
ChrisRackauckas: why not make this non-allocating? For small `u` this could be an issue?

MikeInnes: Not sure what you mean; you want `out_` to be preallocated somewhere? FWIW, this is exactly what the ReverseDiff version was doing also, the allocation of `out_` was just hidden inside the `UDerivativeWrapper`.

ChrisRackauckas: I don't think it would've used UDerivativeWrapper, just UJacobianWrapper which has an in-place form of the function? But I guess the issue is that ReverseDiff couldn't use the non-allocating form? https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/blob/master/src/derivative_wrappers.jl#L88 @YingboMa that would explain why it was bad on smaller models.

Manifest.toml
```diff
@@ -0,0 +1,426 @@
+# This file is machine-generated - editing it directly is not advised
```
ChrisRackauckas: We're still using REQUIRE until METADATA/attobot updates.

src/DiffEqSensitivity.jl
```diff
@@ -2,7 +2,7 @@ __precompile__()
 
 module DiffEqSensitivity
 
-using DiffEqBase, Compat, ForwardDiff, ReverseDiff, DiffEqDiffTools, Statistics
+using DiffEqBase, Compat, ForwardDiff, ReverseDiff, Flux, DiffEqDiffTools, Statistics
```
ChrisRackauckas: Is ReverseDiff needed anymore with this PR?

src/adjoint_sensitivity.jl
```diff
@@ -79,8 +80,12 @@ function (S::ODEAdjointSensitvityFunction)(du,u,p,t)
     end
     mul!(dλ,λ,S.J)
   else
-    tape = S.jac_config
-    vecjacobian!(dλ, λ, tape, y)
```
ChrisRackauckas: `vecjacobian!` is dead code now?

MikeInnes:
In the original code, you use `vecjacobian!` to calculate `df/du`, then separately to calculate `df/du` and `df/dp` (though the `df/du` is ignored). Where `t` is the same you can get both in one shot.

Benchmarks would be useful; but I want to point out that aside from basic correctness issues, using pre-allocation adds undocumented constraints on what `f` can look while getting correct gradients, so perf isn't the only consideration here. Ideally preallocation would use something like Alloc.jl, which is orthogonal to AD and can work with any of them.

Manifest.toml
MikeInnes: I haven't gotten rid of the REQUIRE file, and they aren't mutually exclusive. But the manifest is useful for getting a known good version of things if I end up on e.g. master branches of Flux. Happy to remove before we merge anything if you prefer, though.

ChrisRackauckas: This just leads to issues because CI doesn't actually test the REQUIRE file. For example, in this PR currently Flux isn't in the REQUIRE which should error, but since it uses the Project/Manifest this issue goes undetected. So it's safer to just not use this at all until it's actually what's used by METADATA.

src/adjoint_sensitivity.jl
MikeInnes: Sure, whichever one – https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/blob/2dc3c201ad0b1c568ce204f409ec2806000b32e4/src/function_wrappers.jl#L16

I expect it's redundant given that Flux has to allocate anyway. But we can benchmark at some point.

ChrisRackauckas: Oh, I thought ReverseDiff would try to use the non-allocating form https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/blob/2dc3c201ad0b1c568ce204f409ec2806000b32e4/src/function_wrappers.jl#L15 ...

src/DiffEqSensitivity.jl
MikeInnes: I haven't excised it fully yet, since I don't have test cases for all the places where it's used. But we can work towards that.

ChrisRackauckas: It was only used for `vecjacobian!`, i.e. for `autovecjac` via seeding. If we're doing this with Flux now, there shouldn't be any more ReverseDiff other than in the tests.

MikeInnes: No, wait, I wasn't going crazy: https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/blob/f7e5cd179d9b4e9166ab5051ad1f2004500bdc77/src/adjoint_sensitivity.jl#L113

ChrisRackauckas: why not replace that one with a flux version as well?

MikeInnes: Sure, I have no issue doing that, it just means there's more than simply deleting code left to be done.

ChrisRackauckas: This shouldn't be very different from the other 2?

ChrisRackauckas:
@YingboMa can you run the benchmarks on this?

ChrisRackauckas:
Actually, we probably don't need to run big benchmarks on it, just a simple LV comparison should do fine.

ChrisRackauckas:
```julia
using DiffEqSensitivity,OrdinaryDiffEq, ParameterizedFunctions

f = @ode_def begin
  dx = a*x - b*x*y
  dy = -c*y + x*y
end a b c

p = [1.5,1.0,3.0]
prob = ODEProblem(f,[1.0;1.0],(0.0,10.0),p)
sol = solve(prob,Vern9(),abstol=1e-14,reltol=1e-14)

t = 0.0:0.5:10.0 # TODO: Add end point handling for callback
# g(t,u,i) = (1-u)^2/2, L2 away from 1
function dg(out,u,p,t,i)
  (out.=2.0.-u)
end

using BenchmarkTools
@btime adjoint_sensitivities(sol,Vern9(),dg,t,abstol=1e-14,
                             reltol=1e-14,iabstol=1e-14,ireltol=1e-12)
```

```
Flux
51.791 ms (611491 allocations: 12.83 MiB)
47.276 ms (611491 allocations: 12.83 MiB)

ReverseDiff
45.876 ms (611494 allocations: 12.83 MiB)
45.949 ms (611494 allocations: 12.83 MiB)
```

ReverseDiff is a little bit faster here, but not enough to justify the requirement of tape compilation. So we'll go with the change and I'm sure Flux (and soon Zygote) will continue to improve this anyways.

PR #39, opened 2019-01-18
Don't pass adjoint state vector to a solver

YingboMa:
This PR aims to make adjoint sensitivity analysis play well with stiff
solvers.


ChrisRackauckas:
how did it work before?

YingboMa:
https://github.com/UMCTM/sensitivity_analysis_paper/blob/master/script/adjoint_sensitivity_analysis_bench.jl#L5 :P

PR #41, opened 2019-01-21
Make it twice as fast

MikeInnes:
We currently get gradients for two different function parameters in two different calls, so this avoids more or less 50% redundant work in the adjoint's bottleneck. In particular, it's now faster than the quad method on a simple benchmark.

```julia
julia> @btime adjoint_sensitivities(sol,Vern9(),dg,t,sensealg=SensitivityAlg(quad=true));
  24.132 ms (415273 allocations: 14.76 MiB)

julia> @btime adjoint_sensitivities(sol,Vern9(),dg,t,sensealg=SensitivityAlg(quad=false));
  14.299 ms (292506 allocations: 10.66 MiB)
```

It'd be worth trying some other benchmarks, though I'd expect that anything more complex than this is only going to be doing more work in AD, making the redundant work needed by a second integration not worthwhile.


ChrisRackauckas:
@YingboMa

PR #42, opened 2019-01-21
More efficient backsolve

MikeInnes:
Similar to #41 I'm reusing AD results to avoid repeated calculation; in this case, we make the more efficient `backsolve` method approximately free. I also took the liberty of changing the default settings, since I think these are more in line with what people expect from the adjoint method (e.g. constant memory), and they are also now the fastest on my test cases. If there are reasons to avoid that we can always revert it and set different defaults for DiffEqFlux though.

### DIFF
--- a/src/adjoint_sensitivity.jl
+++ b/src/adjoint_sensitivity.jl
@@ -49,6 +49,8 @@ end
 function (S::ODEAdjointSensitivityFunction)(du,u,p,t)
   idx = length(S.y)
   y = S.y
+  isautojacvec = DiffEqBase.has_jac(S.f) ? false : get_jacvec(S.alg)
+
   if isbcksol(S.alg)
     λ     = @view u[1:idx]
     dλ    = @view du[1:idx]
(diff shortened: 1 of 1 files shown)


src/derivative_wrappers.jl
```diff
@@ -3,7 +3,7 @@ struct SensitivityAlg{CS,AD,FDT} <: DiffEqBase.DEAlgorithm
   quad::Bool
   backsolve::Bool
 end
-Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central},autojacvec=autodiff,quad=true,backsolve=false)
+Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central},autojacvec=autodiff,quad=false,backsolve=true)
```
ChrisRackauckas: `backsolve` can't be the default because it's not technically correct. It's unstable. Besides, no one (except ML people as of 1 month ago) would expect it to be this version.

MikeInnes: Sure. What about `quad` – is there any reason to keep that around at this point?

MikeInnes:
Looks like the tests need updating before we change `quad` anyway, so I just removed that for now.

PR #43, opened 2019-01-22
Add checkpointing scheme to adjoint sensitivity analysis

YingboMa:


src/derivative_wrappers.jl
```diff
@@ -2,10 +2,13 @@ struct SensitivityAlg{CS,AD,FDT} <: DiffEqBase.DEAlgorithm
   autojacvec::Bool
   quad::Bool
   backsolve::Bool
+  checkpointing::Bool
 end
-Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central},autojacvec=autodiff,quad=true,backsolve=false)
+Base.@pure function SensitivityAlg(;chunk_size=0,autodiff=true,diff_type=Val{:central},
```
ChrisRackauckas: We should default to interpolating if `sol.dense`, otherwise default to checkpointing.

test/adjoint.jl
```diff
@@ -36,6 +36,8 @@ easy_res4 = adjoint_sensitivities(solb,Vern9(),dg,t,abstol=1e-14,
                                   reltol=1e-14,iabstol=1e-14,ireltol=1e-12,sensealg=SensitivityAlg(backsolve=true))
 easy_res5 = adjoint_sensitivities(sol,Kvaerno5(nlsolve=NLAnderson(), smooth_est=false),dg,t,abstol=1e-14,
                                  reltol=1e-14,iabstol=1e-14,ireltol=1e-12,sensealg=SensitivityAlg(backsolve=true))
+easy_res6 = adjoint_sensitivities(solb,Vern9(),dg,t,abstol=1e-14,
```
ChrisRackauckas: should stress this test more. Make it only have a checkpoint at 0 and 5 or something like that.

PR #44, opened 2019-02-05
Gradients of u0

MikeInnes:
Not the most elegant interface but it'll get the job done.

### DIFF
--- a/src/adjoint_sensitivity.jl
+++ b/src/adjoint_sensitivity.jl
@@ -307,6 +307,17 @@ function (S::AdjointSensitivityIntegrand)(t)
   S(out,t)
 end
 
+function adjoint_sensitivities_u0(sol,alg,g,t=nothing,dg=nothing;
+                                  abstol=1e-6,reltol=1e-3,
+                                  iabstol=abstol, ireltol=reltol,sensealg=SensitivityAlg(checkpointing=!sol.dense,quad=false),
+                                  checkpoints=sol.t,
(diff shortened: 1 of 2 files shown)


src/adjoint_sensitivity.jl
```diff
@@ -307,6 +307,17 @@ function (S::AdjointSensitivityIntegrand)(t)
...
 
+function adjoint_sensitivities_u0(sol,alg,g,t=nothing,dg=nothing;
+                                  abstol=1e-6,reltol=1e-3,
+                                  iabstol=abstol, ireltol=reltol,sensealg=SensitivityAlg(checkpointing=!sol.dense,quad=false),
+                                  checkpoints=sol.t,
+                                  kwargs...)
+  isquad(sensealg) && error("Can't get sensitivities of u0 with quadrature.")
+  adj_prob = ODEAdjointProblem(sol,g,t,dg,sensealg,checkpoints=checkpoints)
+  adj_sol = solve(adj_prob,alg;abstol=abstol,reltol=reltol,save_everystep=false,kwargs...)
+  -adj_sol[end][1:length(sol.prob.u0)],
```
ChrisRackauckas: Does this need some special input?

MikeInnes: No, the solver always calculates du0 but we ignore it usually.

MikeInnes:
Just in case CI times out again, this does pass locally.

PR #49, opened 2019-03-28
Fix #47

### LINKED ISSUE #47: Can't fix parameter in morris
Sensitivity analysis (Morris) does not allows for parameter fixing -
could also be an issue in other global methods, but I have not tried yet

param_range=[[0,5],[2,4],[3,3]]
param_step=[10,10,1] # didn'r work with [10,10,0] either or [10,10,2] either

BoundsError: attempt to access 1-element StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}} at index [2]

Thanks

Vaibhavdixit02:


Vaibhavdixit02:
@ChrisRackauckas are the LSA tests supposed to be failing? I don't think I made any changes to the LSA code in this

ChrisRackauckas:
merged

PR #56, opened 2019-04-10
Added new method DGSM_Crossed for Global Sensitivity Analysis

siddharthlal25:
This method evaluates the E(c), E(c^2) and E(|c|) where c is 2nd order partial derivative of f w.r.t. x_i and x_j , for all values of i,j given i and j are not equal.


ChrisRackauckas:
These cross terms should probably just be an option in the DGSM method: `crossed=true` turns on this calculation and adds them, otherwise they are nothing.

src/DGSM.jl
```diff
@@ -18,8 +24,11 @@ The inputs to the function DGSM are as follows:
     Array of distribution of respective variables
     Eg- dist = [Normal(5,6),Uniform(2,3)]
 	for two variables
+4.crossed:
+    A string(True/False) which act as indicator for computation of DGSM crossed indices
+    Eg- a True value over there will lead to evauation of crossed indices
 """
-function DGSM(f,samples::Int,distr::AbstractArray)
+function DGSM(f,samples::Int64,distr::AbstractArray, crossed::String)
```
ChrisRackauckas: crossed=false should be a keyword argument and a boolean.

ChrisRackauckas: samples should be an `Int`

siddharthlal25: I have made the above changes

src/DGSM.jl
```diff
@@ -39,11 +45,75 @@ function DGSM(f,samples::Int,distr::AbstractArray)
...
     
-    DGSM_Vi = [DGSM(mean(dfdx[:,i]),mean(abs.(dfdx[:,i])),mean(dfdx[:,i].^2)) for i in 1:k]
+    cross1 = zeros(Float64,0,0)
+    cross2 = zeros(Float64,0,0)
+    cross3 = zeros(Float64,0,0)
     
-    #This function finally returns an array of structures, consisting a, absa and asq
-    #respectively for the k independent parameters
+    if crossed == true
+        cross = DGSM_Crossed(f,samples,distr)
```
ChrisRackauckas: just inline it here.

src/DGSM.jl
```diff
@@ -1,10 +1,13 @@
-
 mutable struct DGSM
-    a::Float64
-    absa::Float64
-    asq::Float64
+    a::Array
```
ChrisRackauckas: these aren't type stable

src/DGSM.jl
```diff
@@ -1,10 +1,13 @@
-
 mutable struct DGSM
```
Vaibhavdixit02: It would make more sense to name it as `DGSM_res` or `Res_DGSM` to indicate the struct is for the returning the result, I got confused and thought this was going to be used to construct the problem

src/DGSM.jl
```diff
@@ -39,11 +48,44 @@ function DGSM(f,samples::Int,distr::AbstractArray)
...
+                crossedsq[a + (b-1)*k] = crossedsq[b + (a-1)*k]
+                abscrossed[b + (a-1)*k] = mean(abs(dfdxdy[i][b + (a-1)*k]) for i in 1:samples)
+                abscrossed[a + (b-1)*k] = abscrossed[b + (a-1)*k]
+            end
+        end
+        
+    else
+        crossed = zeros(Float64,0,0)
+        crossedsq = zeros(Float64,0,0)
+        abscrossed = zeros(Float64,0,0)
```
Vaibhavdixit02: Instead just use `DGSM(a, absa, asq, nothing, nothing, nothing)` and fix the type constraint in the struct as well..

src/DiffEqSensitivity.jl
```diff
@@ -22,7 +23,7 @@ export ODELocalSensitivityFunction, ODELocalSensitivityProblem, SensitivityFunct
        ODEAdjointSensitivityProblem, ODEAdjointProblem, AdjointSensitivityIntegrand,
        adjoint_sensitivities, adjoint_sensitivities_u0,
        morris_sensitivity, MorrisSensitivity, sobol_sensitivity, 
-       SensitivityAlg, regression_sensitivity, DGSM
+       SensitivityAlg, regression_sensitivity, DGSM, DGSM_res
```
Vaibhavdixit02: Why does this need to be exported?

siddharthlal25: Removed it from export

PR #59, opened 2019-04-14
added sigma DGSM

siddharthlal25:
Based on paper- Derivative based global sensitivity measures by Sergeı̈ Kucherenko and Bertrand Iooss. 
Link - https://arxiv.org/pdf/1412.2619.pdf

### DIFF
--- a/src/DGSM.jl
+++ b/src/DGSM.jl
@@ -56,6 +56,8 @@ function DGSM(f,samples::Int,distr::AbstractArray, crossed::Bool = false)
     
     sigma = zeros(Float64,k)
 
+    #Evaluating sigma_i for all input parameters
+
     for i in 1:k
         for j in 1:samples
             sigma[i] += 0.5*(XX[j][i])*(1-XX[j][i])*dfdx[j + (i-1)*samples]^2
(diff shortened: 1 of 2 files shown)


src/DGSM.jl
```diff
@@ -53,7 +54,14 @@ function DGSM(f,samples::Int,distr::AbstractArray, crossed::Bool = false)
     asq = [mean(dfdx[:,i].^2) for i in 1:k]
     absa = [mean(abs.(dfdx[:,i])) for i in 1:k]
     
-    
+    sigma = zeros(Float64,k)
+
+    for i in 1:k
+        for j in 1:samples
+            sigma[i] += 0.5*(XX[j][i])*(1-XX[j][i])*dfdx[j + (i-1)*samples]^2
```
ChrisRackauckas: is this supposed to be xi equation 5?

siddharthlal25: Yes

ChrisRackauckas: Why was the name changed? This quantity should definitely be described in docs

siddharthlal25: Which docs are you referring to?

siddharthlal25:
> what resources calls it sigma? The paper you mention calls it xi. This is something that is obscure so it needs some extra info

Screenshot from 2019-04-15 11-18-29

And my referred equation is:
Screenshot from 2019-04-15 01-19-22

PR #71, opened 2019-05-03
Zygote-based vjps

ChrisRackauckas:


MikeInnes:
FWIW it might be a good idea to swap put the `if` outside the call to `forward` here (you'd end up with a different `forward` in the in-place and out-of-place cases).

Control flow inside `gradient` isn't an inherent problem, but if it's removed then Zygote will be able to compile allocation-free code in some cases.

ChrisRackauckas:
Yeah, that should be done. FWIW, it'll probably always be optimized away, so it's not a big deal.

MikeInnes:
Zygote instruments control flow (e.g. recording pullbacks on a stack) so when differentiating you end up with more complex code that's harder to just optimise away, even if it can be done in principle.

ChrisRackauckas:
I see. Then yeah, it would be a good idea to move it out.

ChrisRackauckas:
Passes with Zygote master 🏁

ChrisRackauckas:
nope looks like it fails

ChrisRackauckas:
@YingboMa what branch of Zygote were you testing on?

ChrisRackauckas:
The remaining issue is just https://github.com/FluxML/Zygote.jl/issues/254

MikeInnes:
I've pushed a manifest and test.jl so that your test case is at least somewhat reproducible. Unfortunately, it appears to be working fine for me (no error) so you'll have to tweak it to show the bug.

ChrisRackauckas:
the weird errors were traced back to be specific to https://github.com/FluxML/Zygote.jl/pull/360, so we'll just ignore that. But then `du .= 0` fails because of no `copyto!`, and if one writes out the loop then you get `nothing` out from the backpass on https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/blob/81b2438ab7c04c695f32363af26980e0042a2c7d/test/adjoint.jl#L158-L165

ChrisRackauckas:
Supersceded by https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/159

PR #75, opened 2019-05-13
Wrong elem_effect assignment if effects is not preallocated

cg1234:
In case the effects array is not preallocated the change_index doesn’t match to the position of the requested parameter index. The array will instead grow dynamically until the max parameter index is reached…
Additionally the y1 assignment at line 70 is not used (will be overriden on line 73).

### DIFF
--- a/src/morris_sensitivity.jl
+++ b/src/morris_sensitivity.jl
@@ -67,8 +67,7 @@ end
 
 function morris_sensitivity(f,p_range,p_steps;relative_scale=false,kwargs...)
     design_matrices = sample_matrices(p_range,p_steps;kwargs...)
-    y1 = f(design_matrices[1][1])
-    effects = [[]]
+    effects = [[] for i=1:length(p)]
     for i in design_matrices
         y1 = f(i[1])
         for j in 1:length(i)-1
@@ -92,10 +91,8 @@ function morris_sensitivity(f,p_range,p_steps;relative_scale=false,kwargs...)
                     elem_effect = @. abs((y1-y2)/(y1*del))
                 end
             end
-            if length(effects) >= change_index && change_index > 0 
+            if change_index > 0
                 push!(effects[change_index],elem_effect)
-            elseif change_index > 0
-                push!(effects,[elem_effect])
             end
         end
     end


src/morris_sensitivity.jl
```diff
@@ -92,10 +91,8 @@ function morris_sensitivity(f,p_range,p_steps;relative_scale=false,kwargs...)
                     elem_effect = @. abs((y1-y2)/(y1*del))
                 end
             end
-            if length(effects) >= change_index && change_index > 0 
+            if change_index > 0
```
ChrisRackauckas: this can go out of bounds?

src/morris_sensitivity.jl
```diff
@@ -67,8 +67,7 @@ end
 
 function morris_sensitivity(f,p_range,p_steps;relative_scale=false,kwargs...)
     design_matrices = sample_matrices(p_range,p_steps;kwargs...)
-    y1 = f(design_matrices[1][1])
-    effects = [[]]
+    effects = [[] for i=1:length(p)]
```
ChrisRackauckas: This is not longer type strict. That may not be an issue though.

Vaibhavdixit02: This used to be something like this (https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/commit/5f32fa05999a9f8bcac97a705d7e5f0726489c5b#diff-d2cff749b323379a3ae89b98849b4d6bL71) but this preallocation doesn't allow for fixed parameters

src/morris_sensitivity.jl
```diff
@@ -67,8 +67,7 @@ end
 
 function morris_sensitivity(f,p_range,p_steps;relative_scale=false,kwargs...)
     design_matrices = sample_matrices(p_range,p_steps;kwargs...)
-    y1 = f(design_matrices[1][1])
```
Vaibhavdixit02: Yes this is not required anymore.

Vaibhavdixit02:
I am not sure I understood why tfere would be wrong elem_effect assignment if effects is not preallocated as we are getting the index to push to in the `effects` vector and then pushing.

cg1234:
> I am not sure I understood why tfere would be wrong elem_effect assignment if effects is not preallocated as we are getting the index to push to in the `effects` array and then pushing.

If I understand correctly the effects array is at beginning empty. Now the loop over the design_matrices starts with any random matrix which holds the also randomly changed parameter. The current change_index is calculated but the length(effects) may be smaller than change_index (because it is randomly changed). In this case the elseif change_index > 0 will match and the elem_effect will just pushed to the end of the effects array (without assigning it to the real change_index position).
I thought the effects array holds the element effects for each changed parameter which was changed. Please correct me if I'm wrong!

Vaibhavdixit02:
Yes you are right. I have fixed it in #76 as I couldn't push to your branch. Thank you for the catch!

cg1234:
Great, thanks a lot!

Von meinem iPad gesendet

PR #87, opened 2019-07-10
add downstream testing

ChrisRackauckas:


ChrisRackauckas:
@YingboMa I've added the downstream tests so it's easier to track.

PR #90, opened 2019-08-03
Some changes to Sobol

Vaibhavdixit02:


ChrisRackauckas:
add a test for what got fixed.

Vaibhavdixit02:
There isn't really any change in the output, it is just a bit of re-writing the trajectory as the standard way instead of the arbitrary way I was doing earlier. I think we can just merge this for now

ChrisRackauckas:
Looks like it has a test failure

Vaibhavdixit02:
This also has performance regression, I am going to close this.

PR #91, opened 2019-08-19
Minor fix

Vaibhavdixit02:


src/sobol_sensitivity.jl
```diff
@@ -38,11 +38,11 @@ function first_order_var(f,p_range,N,y0,v)
         for j in 1:N
             p2 = give_rand_p(p_range)
             p1 = give_rand_p(p_range,[p2[i]],[i])
-            yer =  Array(f(p1)) .* Array(f(p2))
+            p3 = give_rand_p(p_range,p1[1:end .!= i],[k for k in 1:length(p_range) if k != i])
+            yer =  Array(f(p2)) .* (Array(f(p1)) .- Array(f(p3)))
```
ChrisRackauckas: we might want to require that `f` returns something braodcastable instead.

src/sobol_sensitivity.jl
```diff
@@ -58,22 +58,23 @@ function second_order_var(f,p_range,N,y0,v)
             y = zero(y0)
             for k in 1:N
                 p2 = give_rand_p(p_range)
-                p1 = give_rand_p(p_range,[p2[i],p2[j]],[i,j])
-                y .+=  Array(f(p1)) .* Array(f(p2))
+                p1 = give_rand_p(p_range,p2[1:end .!= i],[k for k in 1:length(p_range) if k != i])
+                p3 = give_rand_p(p_range,p2[1:end .!= j],[k for k in 1:length(p_range) if k != j])
```
ChrisRackauckas: p3 isn't used?

Vaibhavdixit02:
@ChrisRackauckas LGTM?

Vaibhavdixit02:
>Let's keep working on the interface here. Instead of manually doing Array, why not require a broadcastable? Or should it be a Vector? We shouldn't be relying on an internal conversion in the method to work.

I hadn't tried removing it. It was just a legacy. It works now. I'll push it.

>and I would like to see if we could get NamedTuple outputs working.

I think this would be easier handling it in Pumas. With the TransformVariables stuff. I am already allowing passing the parameters as the named tuple so I think the post processing of the output there would be better.

PR #101, opened 2019-09-26
eFAST implementation

Vaibhavdixit02:


Vaibhavdixit02:
https://www.researchgate.net/publication/243587760_A_Quantitative_Model-Independent_Method_for_Global_Sensitivity_Analysis_of_Model_Output

https://github.com/cran/sensitivity/blob/master/R/fast99.R

https://github.com/SALib/SALib/blob/master/src/SALib/analyze/fast.py

src/eFAST.jl
```diff
@@ -0,0 +1,65 @@
+Base.@kwdef struct eFAST
+    num_harmonics::Int=4
```
ChrisRackauckas: what's this parameter?

src/eFAST.jl
```diff
@@ -0,0 +1,65 @@
...
+    s = collect((2*pi/n) * (0:n-1))
+    ps = zeros(num_params,n*num_params)
+
+    for i in 1:num_params
+        omega_temp[i] = omega[1]
+        omega_temp[[k for k in 1:num_params if k != i]] = omega[2:end]
+        l = collect((i-1)*n+1:i*n)
+        phi = 2*pi*rand()
+        for j in 1:num_params
+            x =  0.5 .+ (1/pi) .*(asin.(sin.(omega_temp[j]*s .+ phi)))
```
ChrisRackauckas: `asin(sin` is just to take the mod?

Vaibhavdixit02: Not always? 
```
julia> asin(sin(10))
-0.5752220392306202

julia> asin(sin(pi/2))
1.5707963267948966
```

ChrisRackauckas: looks like that's a mod 2pi? If not, what's it doing? It's a fairly odd piece of code.

src/eFAST.jl
```diff
@@ -0,0 +1,65 @@
...
+            ps[j,l] .= quantile.(Uniform(p_range[j][1], p_range[j][2]),x)
```
ChrisRackauckas: why quantile, and why broadcast?

src/eFAST.jl
Vaibhavdixit02: The number of harmonics to sum in the Fourier series decomposition

src/eFAST.jl
Vaibhavdixit02: Oh you mean modulus, sorry, I understood it as `abs`. Yes then it's the same thing.

src/eFAST.jl
```diff
@@ -0,0 +1,65 @@
...
+
+struct eFASTResult{T1}
+    first_order::T1
+    total_order::T1
+end
+
+function gsa(f,method::eFAST,p_range::AbstractVector,n::Int=1000;batch=false)
+    @unpack num_harmonics = method
+    num_params = length(p_range)
+    omega = [floor((n-1)/(2*num_harmonics))]
```
ChrisRackauckas: where is this choice coming from?

PR #105, opened 2019-10-07
Quadrature Sobol

Vaibhavdixit02:


src/DiffEqSensitivity.jl
```diff
@@ -18,13 +18,14 @@ include("morris_sensitivity.jl")
...
 include("DGSM.jl")
+include("eFAST.jl")
 
 export extract_local_sensitivities
 
 export ODELocalSensitivityFunction, ODELocalSensitivityProblem, SensitivityFunction,
        ODEAdjointSensitivityProblem, ODEAdjointProblem, AdjointSensitivityIntegrand,
        adjoint_sensitivities, adjoint_sensitivities_u0, Sobol, Morris, gsa,
-       SensitivityAlg, regression_sensitivity, DGSM
+       SensitivityAlg, regression_sensitivity, DGSM, fast, SobolQuad
```
ChrisRackauckas: don't export `fast`. That'll pollute.

src/DiffEqSensitivity.jl
```diff
@@ -18,13 +18,14 @@ include("morris_sensitivity.jl")
 include("sobol_sensitivity.jl")
 include("regression_sensitivity.jl")
 include("DGSM.jl")
+include("eFAST.jl")
```
ChrisRackauckas: is this from the wrong branch?

Vaibhavdixit02: Yes, just this file though. Will fix in next commit.

src/sobol_sensitivity.jl
```diff
@@ -60,6 +74,33 @@ function first_order_var(f,p_range,N,y0,v,p1,p2,p3)
     ys
 end
 
+function first_order_quad(f,p_range,y0,v,p1,quadalg)
+    ys = Array{typeof(y0.u)}(undef,length(p_range))
+    for i in 1:length(p_range)
+        y = zero(y0)
+        indices = [k for k in 1:length(p_range) if k != i]
+         
+        function f_q(x,p)
```
ChrisRackauckas: set this up with a batched function?

src/sobol_sensitivity.jl
```diff
@@ -60,6 +74,33 @@ function first_order_var(f,p_range,N,y0,v,p1,p2,p3)
...
+            p1[indices] = x
+            f(p1,p)
+        end
+        prob1 = QuadratureProblem(f_q,[p_range[i][1] for i in indices],[p_range[i][2] for i in indices])
+        
+        function f__(x,p)
```
ChrisRackauckas: `__f` is the standard

src/sobol_sensitivity.jl
```diff
@@ -60,6 +74,33 @@ function first_order_var(f,p_range,N,y0,v,p1,p2,p3)
...
+            p1[i] = x
```
ChrisRackauckas: indices?

Vaibhavdixit02: https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/105/files/9ee50445fe6f1dd97d0e771b935531f24013bf5f#diff-a5b8480a843d991444fd63346ec73c55R81

ChrisRackauckas: yes, this line should use that

Vaibhavdixit02: No, if you look at the last equation in page 281 http://www.andreasaltelli.eu/file/repository/Making_best_use.pdf, this is the outer integral, so this will just be the `i`

PR #114, opened 2019-11-17
Have `extract_local_sensitivities` return a 2D subarray

jlumpe:
The current return type for the sensitivity matrix (Array of Arrays) isn't too easy to work with and also copies/allocates memory unnecessarily. This isn't backwards compatible, so I added an additional set of methods taking an `ismatrix::Bool` argument to opt in to the behavior.

### DIFF
--- a/src/local_sensitivity.jl
+++ b/src/local_sensitivity.jl
@@ -125,24 +125,40 @@ function ODELocalSensitivityProblem(f,args...;kwargs...)
   ODELocalSensitivityProblem(ODEFunction(f),args...;kwargs...)
 end
 
+
+# Get ODE u vector and sensitivity values from all time points
 function extract_local_sensitivities(sol)
-  x = sol[1:sol.prob.f.numindvar,:]
-  x,[sol[sol.prob.f.numindvar*j+1:sol.prob.f.numindvar*(j+1),:] for j in 1:(length(sol.prob.p))]
+  ni = sol.prob.f.numindvar
+  u = sol[1:ni, :]
+  du = [sol[ni*j+1:ni*(j+1),:] for j in 1:sol.prob.f.numparams]
+  return u, du
 end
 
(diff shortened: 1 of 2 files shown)


src/local_sensitivity.jl
```diff
@@ -125,24 +125,40 @@ function ODELocalSensitivityProblem(f,args...;kwargs...)
...
+extract_local_sensitivities(sol, i::Integer) = _extract(sol, sol[i], Val{false}())
+extract_local_sensitivities(sol, i::Integer, asmatrix::Bool) = _extract(sol, sol[i], Val{asmatrix}())
+extract_local_sensitivities(sol, t) = _extract(sol, sol(t), Val{false}())
+extract_local_sensitivities(sol, t, asmatrix::Bool) = _extract(sol, sol(t), Val{asmatrix}())
+extract_local_sensitivities(tmp, sol, t) = _extract(sol, sol(tmp, t), Val{false}())
+extract_local_sensitivities(tmp, sol, t, asmatrix::Bool) = _extract(sol, sol(tmp, t), Val{asmatrix}())
+
+
+# Get ODE u vector and sensitivity values from sensitivity problem u vector
+function _extract(sol, su::Vector, asmatrix::Val)
```
ChrisRackauckas:
```suggestion
function _extract(sol, su::Vector, asmatrix::Val = Val(false))
```

jlumpe:
I could use some help resolving some type inference issues here. It seems to me that if I use `asmatrix=true` within a wrapper function that constant propagation should ensure the correct method is called, but it doesn't seem to be working.

ChrisRackauckas:
Yeah, I'm surprised constant prop doesn't kick in here, so I'd just take a Val

jlumpe:
Done, added methods for both.

jlumpe:
Should this be ready to merge?

ChrisRackauckas:
Yes indeed. Thanks!

PR #127, opened 2019-11-25
User defined callbacks will be by default merged with those defined i…

PiotrSokol:
…n the adjoint problem.

As per the discussion in the PR under (DiffEqBase)[https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/388].

Still requires documentation to be added to 'DifferentialEquations.jl'


ChrisRackauckas:
Oh yes, it would be really really weird if this was the only place where it merged by default... maybe the default has to be true...

ChrisRackauckas:
Changing the default. Playing with things locally, indeed it was really really weird. Thank you @PiotrSokol for really pushing us in the right direction here!

PR #129, opened 2019-11-26
batch sobol test

ChrisRackauckas:


ChrisRackauckas:
This now uses quasi Monte Carlo sampling from user given design matrices. Tests require QuasiMonteCarlo.jl to get registered.

Vaibhavdixit02:
Looks good to go to me once QuasiMonteCarlo is registered.

Issue #138, opened 2019-12-22
Morris Method is taking absolute value of Elementary Effects


platawiec:
In the implementation for the Morris method, it looks like the absolute value of the elementary effects are being taken:

https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/blob/dc9232277291b855c6fb9a9ff1c2be6d8ae264d1/src/morris_sensitivity.jl#L105

Then, the mean and variance of this distribution are taken.

According to Global Sensitivity Analysis by Saltelli, and the R package reference for the sensitivity analysis, this is the correct calculation for $\mu^*$ (the mean of the distribution of absolute values of the elementary effects) but it is not the correct calculation for the variance (which should be the variance of the distribution of the elementary effects).

I propose that the elementary effects are stored without having their absolute value taken, and the `MorrisResult` is restructured to:
```
struct MorrisResult{T1,T2}
    means::T1
    means_star::T1
    variances::T1
    elementary_effects::T2
end
```
where `means_star` is the current `means`, `means` is calculated without taking the absolute value, and `elementary_effects` are stored without taking the absolute value.

ChrisRackauckas:
That is a great idea. It has no extra function evaluations and gets both results.

platawiec:
Ok, PR en route 👍

PR #139, opened 2019-12-22
Store and compute on true value of elementary effects

platawiec:
Ref #138 

This stores and computes the `means` and `variances` of the elementary effects in the Morris method. The means of the absolute values of the Elementary Effects is now stored under `means_star`.

### DIFF
--- a/test/morris_method.jl
+++ b/test/morris_method.jl
@@ -11,7 +11,6 @@ function linear_batch(X)
     @. A*X[1,:]+B*X[2,:]
 end
 
-# #138
 function neg_linear_batch(X)
     A= -7
     B= 0.1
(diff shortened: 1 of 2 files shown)


test/morris_method.jl
```diff
@@ -11,6 +11,13 @@ function linear_batch(X)
     @. A*X[1,:]+B*X[2,:]
 end
 
+# #138
```
Vaibhavdixit02: This isn't necessary 🙂

platawiec:
Removed issue references in comments.

The build failed w/ DiffEqFlux but passed on Core. The failure seems unrelated to this PR.

ChrisRackauckas:
Yeah the DiffEqFlux stuff is unrelated so don't worry about it.

PR #140, opened 2019-12-22
Restructure sensitivity algorithms

ChrisRackauckas:
This PR rips apart some of the non-overlapping pieces of the sensitivity algorithms, adds AD as a possible choice for sensitivity algorithms. Some decisions made:

- Backsolve will not use quadrature. It makes no sense to pair the two together since then you don't have the `u` values.
- QuadratureAdjoint is pulled out as its own thing. This is because, if you want to quadrature, you wouldn't checkpoint because it's only good on low-memory problems anyways, so we can simplify the other codes a lot by isolating this.
- InterpolatingAdjoint got the checkpointing implementation, since the plan is to turn this into an interpolating checkpoint
- BacksolveAdjoint could get checkpointing in the future, but it would be different enough that for now I'm not worrying about it.
- Defaults on adjoints is to Quadrature, but not fixed on that. I may want to change that to just InterpolatingAdjoint since that's slightly more general.

Still TODO:

- [ ] extract_sensitivities should handle both forward mode and ForwardDiff mode
- [x] More simplification on the 3 adjoints
- [ ] Ability to use forward mode for Zygote's adjoints
- [x] Update tests and all of that


ChrisRackauckas:
>Since this would lose features of combining multiple configurations and use them together. E.g. checkpointed adjoint with quadrature and v'J via AD.

There is no practical algorithm lost by this from what I could tell. You pointed out the one "feature" that is lost, checkpointed adjoint with quadrature, but let's unpack that for a second. Quadrature requires that you get a continuous solution on the backpass. Adjoints require that you have a larger number of parameters than state variables. Checkpoints only make sense if your state is large enough so that you cannot fit the interpolation of the forward pass into memory. But since we're in adjoint mode, we want the backpass to take the same amount of memory, but it needs to be fully continuous? Also, we'd need to regenerate forward passes of `u` during the quadrature (since du/dp is needed), which assumed interpolation held before, so our previous implementation during checkpointing was using an inaccurate linear interpolation at this step! So, technically this combination existed before but wasn't useful or correct.

Note that v'J via AD and all of that is still there. Your suggestions on how to make the implementations simpler should still be done.

What this high-level split does is: 

- Fixes some correctness issues (like mentioned above)
- Makes it easier to document (the previous approach had a tons of weird switching going on in SensitivityAlg, and it wasn't actually fully correct for forward-mode)
- Makes it easier to handling dispatching with Zygote. For example, this allows us to have a version that allows for computing gradients via the forward-mode code, which will be better for small systems, and won't be a fancy work around but "just another sensitivity algorithm".
- Simplifies the code a ton (the amount of code re-use between these groups was quite minimal, and what was reused could get refactored like you mentioned).
- Makes it easy to add new methods, like shadow adjoints and compressed adjoints.
- Makes it easy to improve existing methods, like checkpointed backsolve. Since it's now separate from the standard checkpointing, the two parts won't clash (the implementations will be different).

But yes, we still need a better interface over `J*v` and `v'J` which would further simplify the code. In fact, once that's done, each of these implementations will look fairly simple.

src/quadrature_adjoint.jl
```diff
@@ -0,0 +1,234 @@
...
+  else
+    paramjac_config = build_param_jac_config(alg,pf,y,p)
+  end
+  AdjointSensitivityIntegrand(sol,adj_sol,p,y,λ,pf,f_cache,pJ,paramjac_config,alg)
+end
+
+function (S::AdjointSensitivityIntegrand)(out,t)
+  @unpack y, λ, pJ, pf, p, f_cache, paramjac_config, alg, sol, adj_sol = S
+  f = sol.prob.f
+  sol(y,t)
```
ChrisRackauckas: @YingboMa here is where checkpointed quadrature was broken before.

ChrisRackauckas:
I have a good feeling that I can significantly chop down the compile times too, since the size of the functions and the amount of stuff that needs to be removed at compile time is significantly decreased.

ChrisRackauckas:
`quad` being a boolean meant that previously it was always compiling that entire code pathway, so an entirely second integration wrapper and all of that quadrature code! With checkpointing, that meant compiling an integrator into all of the code as well. So, that's a big concrete reason for splitting as well.

ChrisRackauckas:
Fixing the checkpointing and making it use a solution instead of an integrator will greatly reduce the compilation as well.

ChrisRackauckas:
Since this is a fairly large change, I'm going to be merging it and we can clean up these other little details on the side. I wanted to open it up for you to start doing your parts.

PR #158, opened 2019-12-28
Allow Zygote.Params

dhairyagandhi96:
This allows Zygote.Params to be used instead of the `destructure`d Vector from DiffEqFlux, making it more straightforward to get Zygote up and running as the AD.

The current way breaks down the gradient calculation between the explicit and implicit parameters and is intended as a stopgap solution for the very short term.

Also needs to be validated against the recent refactor of the repo, and need to validate against more test cases. Critiques welcome, since I am not very used to the DiffEq* codebase

### DIFF
--- a/src/adjoint_sensitivity.jl
+++ b/src/adjoint_sensitivity.jl
@@ -20,7 +20,8 @@ struct ODEAdjointSensitivityFunction{rateType,uType,uType2,UF,PF,G,JC,GC,A,DG,TJ
 end
 
 @noinline function ODEAdjointSensitivityFunction(g,u0,p,alg,discrete,sol,dg,checkpoints,tspan,colorvec)
-  numparams = length(p)
+  p = p isa Params ? vcat(vec.(p)...) : p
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
ahh, I see what you're doing. This requires autojacvec, but I think that's an okay assumption if you're using `Params`. I'll carry this forward to master.

src/adjoint_sensitivity.jl
```diff
@@ -161,7 +162,29 @@ function (S::ODEAdjointSensitivityFunction)(du,u,p,t)
         vec(f(u, p, t))
       end
     end
-    dλ[:], dgrad[:] = back(λ)
+
+    _idy, iback = Zygote.pullback(S.sol.prob.p) do
```
ChrisRackauckas: why is this call separate from line 156? Won't this be computed simultaneously to the other?

ChrisRackauckas:
Superseded by https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/159

PR #163, opened 2019-12-31
Refactor adjoint sensitivity

YingboMa:


src/derivative_wrappers.jl
```diff
@@ -82,6 +82,17 @@ function vecjacobian!(dλ, λ, p, t, S; dgrad=nothing, dy=nothing)
...
     mul!(dλ',λ',J)
+
+    if dgrad !== nothing
+      @unpack pJ, pf, paramjac_config = S
+      if DiffEqBase.has_paramjac(f)
+        f.paramjac(pJ,y,sol.prob.p,t) # Calculate the parameter Jacobian into pJ
+      else
+        jacobian!(pJ, pf, sol.prob.p, f_cache, sensealg, paramjac_config)
+      end
+      mul!(dgrad',λ',pJ)
```
ChrisRackauckas: would make more sense to just do multiple vector Jacobian products instead of forming the Jacobian? With autodiff it's equivalent (if we want to not redo the primal), but with numerical it can save memory.

YingboMa: Ah right, doing a numerical `v'J`, like getting the gradient of `v'f(u)`. Yeah, that definitely helps with memory.

YingboMa:
I think this is a good stopping point, I will add numerical `v'J` in a separate PR.

ChrisRackauckas:
what about handling the integrand? https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/163/files#diff-7e3bd86f3441a3c1fbd3963712a42b6cR141

YingboMa:
There is only one occurrence, does that need a refactor, too?

ChrisRackauckas:
I guess not. I was just thinking that's the same as how it shows up in other spots, but the other spots have it embedded with the other vector Jacobian products so it's not a nice refactor.

PR #164, opened 2019-12-31
start testing concrete solve derivatives

ChrisRackauckas:


ChrisRackauckas:
@YingboMa can you take this on then?

YingboMa:
No problem.

YingboMa:
@ChrisRackauckas `TrackerAdjoint` is broken. But everything else is ready. I have changed `extract_local_sensitivities(sol_fs, Val(true))` to return a vector of matrices.

```julia
julia> extract_local_sensitivities(sol_fs, Val(true))[2] |> eltype
Array{Float64,2}

julia> extract_local_sensitivities(sol_fs, Val(true))[2][2]
2×4 Array{Float64,2}:
 0.000562536  -0.00056222   1.5806e-7    -1.58075e-7
 1.57986e-7   -1.57927e-7  -0.000561746   0.000561825
```

Maybe the fix of `TrackerAdjoint` can go into another PR.

src/local_sensitivity/concrete_solve.jl
```diff
@@ -66,23 +70,36 @@ function _concrete_solve_adjoint(prob,alg,sensealg::AbstractAdjointSensitivityAl
     end
 
     ts = sol.t[sol_idxs]
-    du0, dp = adjoint_sensitivities_u0(sol,args...,df,ts;
-                    kwargs_adj...)
+    if sensealg isa QuadratureAdjoint
+      du0, dp = nothing, adjoint_sensitivities(sol,alg,args...,df,ts; sensealg=sensealg,
```
ChrisRackauckas: This does the u0 on master

PR #166, opened 2019-12-31
Unify interface for GSA with kwarg N

Vaibhavdixit02:


ChrisRackauckas:
Instead of `N`, can we make it `samples`?

PR #174, opened 2020-01-09
fix concrete_solve handling saveat

ChrisRackauckas:
See this is the reason why I think checkpoint should default if using saveat

### DIFF
--- a/src/local_sensitivity/concrete_solve.jl
+++ b/src/local_sensitivity/concrete_solve.jl
@@ -57,16 +57,29 @@ function _concrete_solve_adjoint(prob,alg,
     sol = solve(_prob,alg,args...;save_start=true,save_end=true,kwargs...)
   end
 
-  no_start = !save_start
-  no_end = !save_end
-  sol_idxs = 1:length(sol)
-  no_start && (sol_idxs = sol_idxs[2:end])
-  no_end && (sol_idxs = sol_idxs[1:end-1])
-  # If didn't save start, take off first. If only wanted the end, return vector
-  only_end = length(sol_idxs) <= 1
-  u = sol[sol_idxs]
-  only_end && (sol_idxs = length(sol))
-  out = only_end ? sol[end] : reduce((x,y)->cat(x,y,dims=ndims(u)),u.u)
+  if saveat isa Number
+    if _prob.tspan[2] > _prob.tspan[1]
+      ts = _prob.tspan[1]:abs(saveat):_prob.tspan[2]
+    else
+      ts = _prob.tspan[2]:abs(saveat):_prob.tspan[1]
+    end
+    out = sol(ts)
+  elseif !isempty(saveat)
+    no_start = !save_start
+    no_end = !save_end
+    sol_idxs = 1:length(sol)
+    no_start && (sol_idxs = sol_idxs[2:end])
+    no_end && (sol_idxs = sol_idxs[1:end-1])
+    # If didn't save start, take off first. If only wanted the end, return vector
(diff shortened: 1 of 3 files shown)


src/local_sensitivity/sensitivity_algorithms.jl
```diff
@@ -58,5 +58,6 @@ struct ZygoteAdjoint <: AbstractAdjointSensitivityAlgorithm{nothing,true,nothing
 @inline get_chunksize(alg::DiffEqBase.AbstractSensitivityAlgorithm{CS,AD,FDT}) where {CS,AD,FDT} = CS
 @inline diff_type(alg::DiffEqBase.AbstractSensitivityAlgorithm{CS,AD,FDT}) where {CS,AD,FDT} = FDT
 @inline get_jacvec(alg::DiffEqBase.AbstractSensitivityAlgorithm) = alg.autojacvec
+@inline ischeckpointing(alg::DiffEqBase.AbstractSensitivityAlgorithm) = isdefined(alg, :checkpointing) ? alg.checkpointing : false
```
YingboMa:
```suggestion
@inline ischeckpointing(alg::DiffEqBase.AbstractSensitivityAlgorithm, ::Vararg) = isdefined(alg, :checkpointing) ? alg.checkpointing : false
```

src/local_sensitivity/sensitivity_algorithms.jl
```diff
@@ -58,5 +58,6 @@ struct ZygoteAdjoint <: AbstractAdjointSensitivityAlgorithm{nothing,true,nothing
 @inline get_chunksize(alg::DiffEqBase.AbstractSensitivityAlgorithm{CS,AD,FDT}) where {CS,AD,FDT} = CS
 @inline diff_type(alg::DiffEqBase.AbstractSensitivityAlgorithm{CS,AD,FDT}) where {CS,AD,FDT} = FDT
 @inline get_jacvec(alg::DiffEqBase.AbstractSensitivityAlgorithm) = alg.autojacvec
+@inline ischeckpointing(alg::DiffEqBase.AbstractSensitivityAlgorithm) = isdefined(alg, :checkpointing) ? alg.checkpointing : false
 @inline ischeckpointing(alg::DiffEqBase.AbstractSensitivityAlgorithm, ::Any) = isdefined(alg, :checkpointing) ? alg.checkpointing : false
```
YingboMa:
```suggestion
```

src/local_sensitivity/concrete_solve.jl
```diff
@@ -47,7 +50,12 @@ function _concrete_solve_adjoint(prob,alg,sensealg::AbstractAdjointSensitivityAl
   if haskey(kwargs, :callback_adj)
     kwargs_adj = merge(kwargs_adj, NamedTuple{(:callback,)}( [get(kwargs, :callback_adj, nothing)] ))
   end
-  sol = solve(_prob,alg,args...;save_start=true,save_end=true,kwargs...)
+
+  if ischeckpointing(sensealg)
+    sol = solve(_prob,alg,args...;save_start=true,save_end=true,saveat=saveat,kwargs...)
+  else
```
YingboMa:
```suggestion
  else
    isempty(saveat) || @warn "The algorithm has checkpointing disabled, but `saveat` is non-empty. Please either enable checkpointing to save memory usage or remove `saveat`"
```

ChrisRackauckas: I don't think we want to warn, since small problems with saveat is probably the most common use for this function, and so this would be warning on the 99% use case. But we really really need to document this well.

PR #189, opened 2020-02-06
Adjoint sensitivity with mass matrices

YingboMa:


src/local_sensitivity/adjoint_common.jl
```diff
@@ -83,7 +83,7 @@ function generate_callbacks(sensefun, g, λ, t, callback, init_cb)
       λ  = isq ? λ : @view(λ[1:idx])
       g(λ,y,p,t[cur_time[]],cur_time[])
       if isq
-        u .+= λ
+        u .+= integrator.f.mass_matrix \ λ
```
ChrisRackauckas:
```julia
MM = [1 0 0
      0 1 0
      0 0 0]
MM \ rand(3)
```

This won't handle DAEs? Do we need to run the reinitialization algorithm right after this?

PR #196, opened 2020-02-14
Handle singular mass matrices (QuadratureAdjoint only)

YingboMa:


src/local_sensitivity/adjoint_common.jl
```diff
@@ -23,7 +27,25 @@ function adjointdiffcache(g,sensealg,discrete,sol,dg;quad=false)
...
   numindvar = length(u0)
   isautojacvec = get_jacvec(sensealg)
-  J = isautojacvec ? nothing : similar(u0, numindvar, numindvar)
+
+  issemiexplicitdae = false
+  mass_matrix = sol.prob.f.mass_matrix'
+  if mass_matrix isa UniformScaling
+    factorized_mass_matrix = mass_matrix
+  else
+    diffvar_idxs = findall(x->any(!iszero, @view(mass_matrix[:, x])), axes(mass_matrix, 2))
```
ChrisRackauckas: why axes instead of eachcol?

src/local_sensitivity/adjoint_common.jl
```diff
@@ -23,7 +27,25 @@ function adjointdiffcache(g,sensealg,discrete,sol,dg;quad=false)
...
+    algevar_idxs = setdiff(eachindex(u0), diffvar_idxs)
+    M̃ = @view mass_matrix[diffvar_idxs, diffvar_idxs]
```
ChrisRackauckas: This won't handle `M` operators, but we should make an issue for that.

src/local_sensitivity/adjoint_common.jl
```diff
@@ -92,21 +117,35 @@ function generate_callbacks(sensefun, g, λ, t, callback, init_cb)
...
   time_choice = let cur_time=cur_time, t=t
     integrator -> cur_time[] > 0 ? t[cur_time[]] : nothing
   end
-  affect! = let isq = (sensealg isa QuadratureAdjoint), λ=λ, t=t, y=y, cur_time=cur_time, idx=length(prob.u0)
+  affect! = let isq = (sensealg isa QuadratureAdjoint), λ=λ, t=t, y=y, cur_time=cur_time, idx=length(prob.u0), F=factorized_mass_matrix
     function (integrator)
       p, u = integrator.p, integrator.u
       λ  = isq ? λ : @view(λ[1:idx])
-      g(λ,y,p,t[cur_time[]],cur_time[])
+      gᵤ = similar(λ)
```
ChrisRackauckas: do we always need this?

src/local_sensitivity/adjoint_common.jl
```diff
@@ -92,21 +117,35 @@ function generate_callbacks(sensefun, g, λ, t, callback, init_cb)
...
+      g(gᵤ,y,p,t[cur_time[]],cur_time[])
       if isq
-        u .+= integrator.f.mass_matrix \ λ
+        if issemiexplicitdae
+          jacobian!(J, uf, y, f_cache, sensealg, jac_config)
+          dhdd = J[algevar_idxs, diffvar_idxs]
+          dhda = J[algevar_idxs, algevar_idxs]
+          # TODO: maybe need a `conj`
+          Δλa = -dhda'\gᵤ[algevar_idxs]
```
ChrisRackauckas: we should make sure to save those notes PDFs in the notes repo.

test/adjoint.jl
```diff
@@ -589,4 +589,31 @@ using Test
...
+    du[3] =  y₁ + y₂ + y₃ - 1
+    nothing
+  end
+  M = [1. 0  0
+       0  1. 0
+       0  0  0]
+  f = ODEFunction(rober,mass_matrix=M)
+  p = [0.04,3e7,1e4]
+
+  Alg = Rosenbrock23 # Rodas5 fails because of interpolation weirdness. Maybe because this problem is very stiff? But that is very much unexpected.
```
ChrisRackauckas: Rodas5 doesn't use the stiff-aware interpolation, it just uses 3rd order Hermite. So yeah, this could cause an issue.

ChrisRackauckas: It does have a not implemented stiff-aware interpolation though, which might be good to add.

src/local_sensitivity/adjoint_common.jl
YingboMa: Because I need indices. `findall(x->any(!iszero, x), eachcol(M))` won't work.

PR #200, opened 2020-02-23
[WIP] handle non-uniqueness in checkpoints

ChrisRackauckas:
@yingboma can you take a good luck and possibly fix up small remaining bugs here? Essentially the idea is that, if the forward pass used callbacks, then you can have non-unique timepoints for the checkpoints. If that's the case, then we want to make sure that we pull one of the values out. The last value? That would be after the callbacks applied. If that makes sense, then we'd want to make sure that the callbacks aren't re-applied in the checkpointing interval calculation. So maybe the first is better? Do you know if unique always gets the first or second? Big questions.


YingboMa:
Yeah, I have to think more about this. I will take this PR.

PR #201, opened 2020-02-24
adapt for Tracked{CuArray}

dhairyagandhi96:
Part 1 of the compilation error for NeuralDSDE on GPU

Part2: need to get the gradients to be reported correctly


ChrisRackauckas:
This causes test failures.

dhairyagandhi96:
Yeah, the compilation issue is gone though, afaict, looking into the gradients

PR #216, opened 2020-03-26
Handle DESolution in GSA

Vaibhavdixit02:


ChrisRackauckas:
Why not have the user give it to you as an Array?

Vaibhavdixit02:
I think it is expected to support the ecosystem out of the box?

ChrisRackauckas:
then do this on `RecursiveArrayTools.AbstractVectorOfArray`

PR #217, opened 2020-03-26
Make InterpolatingAdjoint and BacksolveAdjoint mass matrix compatible

YingboMa:


src/local_sensitivity/adjoint_common.jl
```diff
@@ -127,27 +127,23 @@ function generate_callbacks(sensefun, g, λ, t, callback, init_cb)
   affect! = let isq = (sensealg isa QuadratureAdjoint), λ=λ, t=t, y=y, cur_time=cur_time, idx=length(prob.u0), F=factorized_mass_matrix
     function (integrator)
       p, u = integrator.p, integrator.u
-      # Warning: alias here! Be care with λ
+      # Warning: alias here! Be careful with λ
       gᵤ = isq ? λ : @view(λ[1:idx])
       g(gᵤ,y,p,t[cur_time[]],cur_time[])
-      if isq
```
ChrisRackauckas: you don't need isq?

YingboMa: `isq` is checked in line 131.

YingboMa:
CI failure is unrelated.

ChrisRackauckas:
One of the CI's just messed up.

Vaibhavdixit02:
Looks like this broke DiffEqFlux DAEs? 

https://travis-ci.org/github/SciML/DiffEqFlux.jl/builds/667581115#L872

What needs to change there? 
cc: @abhigupta768

ChrisRackauckas:
Looks like it needs a fix and a test for out of place functions.

abhigupta768:
>Looks like it needs a fix 
I think in https://github.com/SciML/DiffEqFlux.jl/blob/e620ea99053cd6a818cd9a0b8deeca8af26c5842/src/neural_de.jl#L298 `autojacvec=DiffEqSensitivity.ReverseDiffVJP(true)` is the reason for the failure? Should it be `InterpolatingAdjoint()` only now?

ChrisRackauckas:
No, it should be using ReverseDiffVJPs here, but we should figure out why it's erroring. It's not really known. @YingboMa can you look into this really quickly by the end of the day?

Vaibhavdixit02:
@YingboMa did you get a chance to look at this?

PR #220, opened 2020-04-02
Handle VecOfArray

dhairyagandhi96:
With sister PR https://github.com/SciML/RecursiveArrayTools.jl/pull/86

Checking to see what CI says


ChrisRackauckas:
I am not sure when that branch is hit? Do we have an example?

dhairyagandhi96:
Happened for https://github.com/SciML/DiffEqFlux.jl/issues/160

dhairyagandhi96:
Not happy about the branching here

ChrisRackauckas:
It doesn't look like this computes correct values. Added a test.

dhairyagandhi96:
Looking into the failure, hopefully it's just in the adjoint

ChrisRackauckas:
I think using `sum` is incorrect: it should just be the identity.

src/local_sensitivity/concrete_solve.jl
```diff
@@ -65,7 +65,11 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
       if only_end
         _out[:] .= -vec(Δ)
       else
-        _out[:] .= -adapt(typeof(u0),reshape(Δ, prod(size(Δ)[1:end-1]), size(Δ)[end])[:, i])
+        if typeof(Δ) <: AbstractArray{<:AbstractArray}
+          _out[:] .= sum.(Δ)
```
ChrisRackauckas, line 69: Maybe `_out[:] .= Δ[i]`

dhairyagandhi96: I'll try that out

dhairyagandhi96:
CI is green now

Issue #182, opened 2020-01-20
Adjoints of SteadyStateProblem


arnavs:
Perhaps because it's a multidimensional problem?

```
using DiffEqFlux, Flux, DifferentialEquations

z_grid = -5:0.1:5.
v̄(z) = 2*exp(z)
N = length(z_grid)

function f(du, u, p, t)
    v = u[1, :]
    dv = u[2, :]
    du[1, :] .= dv
    du[2, :] .= 2 * (v .- exp.(z_grid)) 
end 

u₀ = zeros(2, N)
prob = ODEProblem(f, u₀, (0., 0.))


function predict_adjoint() # Our 1-layer neural network
    Array(concrete_solve(prob, Tsit5(), u₀, (0., 0.)))[:, :, 1]
end

function loss_adjoint()
    v_sol = predict_adjoint()[1, :]
    return sum(abs2.(v_sol - v̄.(z_grid)))
end

data = Iterators.repeated((), 200)
opt = ADAM(0.1)
Flux.train!(loss_adjoint, Flux.params(u₀), data, opt)
```

With stack trace 

```
BoundsError: attempt to access 1-element Array{Float64,1} at index [2]

Stacktrace:
 [1] getindex(::Array{Float64,1}, ::Int64) at ./array.jl:744
 [2] ode_interpolation!(::Array{Float64,2}, ::Float64, ::Function, ::Nothing, ::Type, ::Tuple{Float64,Float64}, ::Symbol) at /Users/arnavsood/.julia/packages/OrdinaryDiffEq/6TJck/src/dense/generic_dense.jl:319
 [3] InterpolationData at /Users/arnavsood/.julia/packages/OrdinaryDiffEq/6TJck/src/interp_func.jl:75 [inlined]
 [4] #_#14 at /Users/arnavsood/.julia/packages/DiffEqBase/5h7rF/src/solutions/ode_solutions.jl:19 [inlined]
 [5] ODESolution at /Users/arnavsood/.julia/packages/DiffEqBase/5h7rF/src/solutions/ode_solutions.jl:19 [inlined] (repeats 2 times)
[...]

PR #235, opened 2020-04-24
[WIP] adjoint for steady state solution

frankschae:
I tried to implement the adjoint sensitivity function for the steady state solution #182 . I didn't manage to make the linsolve work, yet .. but I assumed we'd want that a user can choose the linear solver, too.. 

Also, I was not sure if del g/ del p in (3) of https://math.mit.edu/~stevenj/18.336/adjoint.pdf should be added automatically by AD/FD or by a user input. (For the tests in adjoint.jl I chose a loss function that doesn't explicitly depend on p so far. )


test/adjoint.jl
```diff
@@ -660,3 +660,136 @@ using LinearAlgebra, DiffEqSensitivity, OrdinaryDiffEq, ForwardDiff, QuadGK
     end
   end
 end
+
+
+using Test, LinearAlgebra
```
ChrisRackauckas, line 665: lets make this a different file.

src/local_sensitivity/sensitivity_interface.jl
```diff
@@ -22,6 +22,19 @@ function _adjoint_sensitivities(sol,sensealg,alg,g,t=nothing,dg=nothing;
...
 
+function _adjoint_sensitivities(sol,sensealg::SteadyStateAdjoint,alg,g,dg=nothing;
+                                   abstol=1e-6,reltol=1e-3,
+                                   kwargs...)
+  cache = SteadyStateAdjointSensitivityFunction(g,sensealg,sol,dg)
+
+  linsolve_tmp = cache.J'\vec(cache.gu')
+  #cache.linsolve(vec(cache.gu), cache.J', vec(cache.linsolve_tmp))
+
+  -linsolve_tmp'*cache.pJ #dg/dp?
```
ChrisRackauckas, line 33: yeah only the discrete form really works here so we should be taking dg/dp. Concrete solve might even be easier since that's the dy value from the backpass.

src/local_sensitivity/sensitivity_interface.jl
```diff
@@ -22,6 +22,19 @@ function _adjoint_sensitivities(sol,sensealg,alg,g,t=nothing,dg=nothing;
     adj_sol[end][(1:l) .+ length(sol.prob.u0)]'
 end
 
+function _adjoint_sensitivities(sol,sensealg::SteadyStateAdjoint,alg,g,dg=nothing;
+                                   abstol=1e-6,reltol=1e-3,
+                                   kwargs...)
+  cache = SteadyStateAdjointSensitivityFunction(g,sensealg,sol,dg)
+
+  linsolve_tmp = cache.J'\vec(cache.gu')
```
ChrisRackauckas, line 30: all that should we need to do is compute the vector'*Jacobian, which is the vecjacobian! stuff of the other adjoints. Copy that general workflow, since we want to use backprop on f for that. You might want to read some of the 18.337 notes, specifically https://mitmath.github.io/18337/lecture10/estimation_identification on vjps

frankschae: Thanks for the remarks and the link to the notes! I'll read them and try to update the files accordingly.

ChrisRackauckas:
I think this is probably good. Follow up PRs can handle the extra case and linsolve, but I think it would be good to the basics cleaned up and done first.

The other thing to handle will be `concrete_solve`, but that shouldn't be too difficult given what I see here.

frankschae:
Is there something missing before we can merge it? I am planning to do the `concrete_solve` next .

PR #239, opened 2020-05-12
Add CIs for Sobol method

Vaibhavdixit02:


src/global_sensitivity/sobol_sensitivity.jl
```diff
@@ -107,8 +107,38 @@ function gsa(f, method::Sobol, A::AbstractMatrix, B::AbstractMatrix;
     SobolResult(Sᵢ, nothing, 2 in method.order ? Sᵢⱼ : nothing, nothing, Tᵢ, nothing)
 end
 
+function gsa(f,method::Sobol,A::AbstractVector{<:AbstractMatrix},B::AbstractVector{<:AbstractMatrix};kwargs...)
+    d,n = size(A[1])
+    res = map(A,B) do A_,B_
+        Threads.@spawn gsa(f, method, A_, B_; kwargs...)
```
ChrisRackauckas, line 113: I think we should just make bigger design matrices and do the call to the user. If they want parallelism, they will already be using the batch form and will already have parallelism more optimally implemented.

src/global_sensitivity/sobol_sensitivity.jl
```diff
@@ -44,58 +45,90 @@ function gsa(f, method::Sobol, A::AbstractMatrix, B::AbstractMatrix;
         all_y = multioutput ? reduce(hcat, _y) : _y
     end
 
+    Eys = []
```
ChrisRackauckas, line 48: instability

ChrisRackauckas:
It looks really really ugly, but I guess it works.

src/global_sensitivity/sobol_sensitivity.jl
Vaibhavdixit02: It gets weird because of the multioutput and non-multioutput branches, I'll try to come up with something that works, I don't think it's that big an issue though.

Vaibhavdixit02:
Yeah..it was very tricky to get it set up this way. 

Do you have any specific things you think can be made less ugly?

ChrisRackauckas:
Not sure... I guess this is good enough. Any reasonable test for this?

Vaibhavdixit02:
It's hard to match the exact values, I think it is okay to assume that as long as the indices are correct the CIs are reasonable?

PR #242, opened 2020-05-16
[WIP] Backsolve SDE

frankschae:
I got stuck with the backsolve adjoint for SDEs.. 
The main problem is that  `NoiseGrid` didn't work for the solution of SDEs in reverse time. So I switched to the forward version (-t_end, -t_start).  Apparently I am missing some minus signs somewhere..


src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -37,6 +87,58 @@ function (S::ODEBacksolveSensitivityFunction)(du,u,p,t)
   return nothing
 end
 
+
+function (S::SDEBacksolveDriftSensitivityFunction)(du,u,p,t)
```
ChrisRackauckas, line 91: isn't this the same as the ODE one? Why not just reuse the code?

mschauer: Isnt that the place to deal with on Ito vs Stratonovich issues.

frankschae: Currently I think it would be best to add the correction already here:
https://github.com/SciML/DiffEqSensitivity.jl/pull/242/files#diff-99f3ec143121f5ddfb78d5cc3e62ca35R101-R105

frankschae: (I left them out for the moment, as not all the `tdir` related issues are solved)

src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -81,6 +82,76 @@ end
...
+  end
+
+  sdefun = SDEFunction(sense_drift,sense_diffusion,mass_matrix=mm)
+
+  # replicated noise
+  _sol = deepcopy(sol)
+  if typeof(_sol.W.dW) <: Number
+    noisearray =  _sol.W.W
+  elseif StochasticDiffEq.is_diagonal_noise(sol.prob) && length(sol.W.dW)==1
+    noisearray = [W[1] for W in _sol.W.W]
```
ChrisRackauckas, line 140: you shouldn't have to do that?

ChrisRackauckas: You should be able to just reverse the array of array IIUC. If not, we need to just create a new dispatch to reverse.

src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -81,6 +82,76 @@ end
...
+  discrete = t != nothing
+
+  p === DiffEqBase.NullParameters() && error("Your model does not have parameters, and thus it is impossible to calculate the derivative of the solution with respect to the parameters. Your model must have parameters to use parameter sensitivity calculations!")
+  numstates = length(u0)
+  numparams = length(p)
+
+  len = length(u0)+numparams
+  λ = similar(u0, len)
+
+  drift_function = ODEFunction(sol.prob.f)
```
ChrisRackauckas, line 102: that's a little awkward. How come this is required?

src/local_sensitivity/derivative_wrappers.jl
```diff
@@ -201,7 +201,11 @@ end
 function _vecjacobian!(dλ, λ, p, t, S::SensitivityFunction, isautojacvec::ZygoteVJP, dgrad, dy)
   @unpack y, sensealg = S
   prob = getprob(S)
-  f = prob.f
+  if prob isa Union{SDEProblem}
+    @unpack f = S
```
ChrisRackauckas, line 205: the goal should probably be to make the other SensitivityFunctions all have the `f` in there, but this works for now.

test/local_sensitivity/sde.jl
```diff
@@ -0,0 +1,139 @@
...
+					#,abstol=abstol, reltol=reltol
+					)
+  sol = solve(tmp_prob,RKMil(interpretation=:Stratonovich),dt=tend/1000,adaptive=false,saveat=t)
+  A = convert(Array,sol)
+  res = g(A,p,nothing)
+  @show res
+  res
+end
+res_sde_forward2 = ForwardDiff.gradient(GSDE,p2)
+res_sde_reverse2 = ReverseDiff.gradient(GSDE,p2)
```
ChrisRackauckas, line 124: I'd just remove ReverseDiff from here

frankschae: Although the tests are passing now for reverse diff?

src/local_sensitivity/backsolve_adjoint.jl
frankschae: I needed it for the diffusion.. to get it through the Jacobian checks and thought it might be better to have it symmetric with respect to the drift  (though there it is indeed not required)

src/local_sensitivity/backsolve_adjoint.jl
frankschae: the reverse was actually not the issue. I could construct the NoiseGrid and the SDEProblem but when calling solve on the inplace function I got a broadcasting error in perform_step!

ChrisRackauckas: This will only work on length 1 SDEs though, which is somewhat odd. This is the next thing to fix.

src/local_sensitivity/backsolve_adjoint.jl
ChrisRackauckas: `sol.prob.f` is already an ODEFunction though?

test/local_sensitivity/sde.jl
ChrisRackauckas: oh cool, nevermind then

src/local_sensitivity/derivative_wrappers.jl
ChrisRackauckas: We could merge without this, but open an issue on it.

src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -53,8 +67,8 @@ end
   numparams = length(p)
 
   len = length(u0)+numparams
-  λ = similar(p, len)
-  sense = ODEBacksolveSensitivityFunction(g,sensealg,discrete,sol,dg,f.colorvec)
+  λ = similar(u0, len)
```
ChrisRackauckas, line 70: why change from `p`?

src/local_sensitivity/derivative_wrappers.jl
```diff
@@ -222,15 +222,24 @@ function _vecjacobian!(dλ, λ, p, t, S::SensitivityFunction, isautojacvec::Zygo
...
     if typeof(y) <: ArrayPartition
       dλ .= ArrayPartition(tmp1.x)
     else
       dλ[:] .= vec(tmp1)
     end
-
-    dgrad !== nothing && (dgrad[:] .= vec(tmp2))
     dy !== nothing && (dy[:] .= vec(_dy))
+    if length(dgrad) == length(tmp2)
+      dgrad !== nothing && (dgrad[:] .= vec(tmp2))
```
ChrisRackauckas, line 233: you'll need to check for nothing before you check for its length

src/local_sensitivity/derivative_wrappers.jl
```diff
@@ -222,15 +222,24 @@ function _vecjacobian!(dλ, λ, p, t, S::SensitivityFunction, isautojacvec::Zygo
...
+    else
+      for (i, λi) in enumerate(λ)
+        _, back = Zygote.pullback(y, prob.p) do u, p
+          f(u, p, t)[i]
+        end
+        _,tmp2 = back(λi)
```
ChrisRackauckas, line 239: you should be able to just output the whole thing and then pullback lambda transpose. In fact, do you need a separate computation for this? I don't think you do.

src/local_sensitivity/backsolve_adjoint.jl
frankschae: oh this had no reason..

src/local_sensitivity/derivative_wrappers.jl
frankschae: Do you mean like:
```
_, back = Zygote.pullback(y, plorenz) do u, p
		vec(f(u,p,t))
end
_,tmp2 = back(λ')
```
then I only get one vector, which doesn't seem to agree with the lines above.

src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -1,31 +1,46 @@
-struct ODEBacksolveSensitivityFunction{C<:AdjointDiffCache,Alg<:BacksolveAdjoint,uType,pType,CV} <: SensitivityFunction
+struct ODEBacksolveSensitivityFunction{C<:AdjointDiffCache,Alg<:BacksolveAdjoint,uType,pType,fType<:DiffEqFunction,CV} <: SensitivityFunction
```
ChrisRackauckas, line 1: don't add bounds when they aren't necessary, and don't make them too tight. This was changed to ODEFunction,SDEFunction in this PR, which is what broke SecondOrderODEFunction.

src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -82,6 +97,70 @@ end
...
+    copyto!(@view(mm[idx, idx]), I)
+    idx = len+1:len2
+    copyto!(@view(mm[idx, idx]), sol.prob.f.mass_matrix)
+  end
+
+  sdefun = SDEFunction(sense_drift,sense_diffusion,mass_matrix=mm)
+
+  # replicated noise
+  _sol = deepcopy(sol)
+  backwardnoise = DiffEqNoiseProcess.NoiseGrid(reverse!(_sol.t),reverse!( _sol.W.W))
```
ChrisRackauckas, line 153: we probably need to add `reverse!(::AbstractNoiseProcess)` to generalize this. Open an issue on DiffEqNoiseProcess to track that.

src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -82,6 +97,70 @@ end
...
+
+  return SDEProblem(sdefun,sense_diffusion,z0,tspan,p,
+    callback=cb,
+    noise=backwardnoise,
+    noise_rate_prototype = zeros(length(z0),numstates)
```
ChrisRackauckas, line 158: this use of zeros here isn't GPU-safe. `similar(z0,length(z0),numstates)` followed by `x.=false` would be better.

src/local_sensitivity/derivative_wrappers.jl
```diff
@@ -224,9 +224,20 @@ function _vecjacobian!(dλ, λ, p, t, S::SensitivityFunction, isautojacvec::Zygo
     else
       dλ[:] .= vec(tmp1)
     end
-
-    dgrad !== nothing && (dgrad[:] .= vec(tmp2))
     dy !== nothing && (dy[:] .= vec(_dy))
+    if dgrad !== nothing
```
ChrisRackauckas, line 228: let's do this for now, but I think what we'll need to do is essentially revert it and in the SDE pass `dgrad = nothing`, and do this part separately. It can be much more efficient to calculate the Jacobian via forward mode in many cases, so we probably want options for how to build this Jacobian rather than hardcoding it. Then, that would keep the focus of `vecjacobian!` just on vector-Jacobian products. Since this loop doesn't share code with what's above it, it just doesn't make all that much sense in this spot.

ChrisRackauckas: Let's open an issue to track this.

src/local_sensitivity/backsolve_adjoint.jl
frankschae: what does `x.=false` mean?

ChrisRackauckas: overwrites all of the values to 0

PR #243, opened 2020-05-20
concrete_solve adjoints for second order ODEs

ChrisRackauckas:


src/local_sensitivity/derivative_wrappers.jl
```diff
@@ -218,7 +218,13 @@ function _vecjacobian!(dλ, λ, p, t, S::SensitivityFunction, isautojacvec::Zygo
       vec(f(u, p, t))
     end
     tmp1,tmp2 = back(λ)
-    dλ[:] .= vec(tmp1)
+
+    if typeof(y) <: ArrayPartition
+      dλ .= ArrayPartition(tmp1.x)
```
ChrisRackauckas, line 223: @dhairyagandhi96 is there a way to make Zygote return the ArrayPartition instead of (x = ...)?

dhairyagandhi96: No I don't think so, since we can't rely on having the appropriate constructors available for the `NamedTuple`

dhairyagandhi96: Oh wait this is in the forward pass! That should happen if the function returns the right type

ChrisRackauckas: The issue is a bit deeper. It's an ArrayPartition forward, and a NamedTuple back. However, the size of the structure isn't really known. It's constructed from an tuple of arrays, but trying to define a literal getindex doesn't seem to work because then you don't get the same size ArrayPartition? The examples show putting `nothing` into the constructor but that won't work here.

dhairyagandhi96: How do you constructors the partition back from the nt?

ChrisRackauckas: It constructs from tuples and `.x` is the correct tuple.

dhairyagandhi96: Yeah, its difficult to return the type in the back.

dhairyagandhi96: But seems like you have the right structure to construct them from. Maybe a function like 
```julia
maybeap(x) = vec(x)
maybeap(x::NamedTuple) = ArrayPartition(x.x)
```
could be inlined? I am not sure of a nicer way to send the right value across while being generic. If you have an adjoint defined already, then returning the right type is trivial of course.

```julia
julia> struct M{T}
         a::T
       end

julia> myfn(a::M, b::M) = a.a + b.a
myfn (generic function with 1 method)

julia> gradient((x,y) -> myfn(x,y), M(1), M(3))
((a = 1,), (a = 1,))

julia> @adjoint function myfn(a::M, b::M)
           myfn(a,b), Δ -> begin
             (M(Δ[1].a), M(Δ[2].a))
           end
       end

julia> gradient((x,y) -> myfn(x,y).a, M(1), M(3))
(M{Int64}(1), M{Int64}(1))
```

PR #248, opened 2020-05-23
Fix save_idxs support in adjoints

ChrisRackauckas:
@vaibhavdixit02 this should be what you need.


ChrisRackauckas:
@YingboMa we need to think a bit about how to do forward modes with save_idxs

Issue #198, opened 2020-02-19
Zygote fails for all Rosenbrock methods


scheidan:
I've observed that Zygote currently fails to take the gradient of a solution calculated by any Rosenbrock method with the following error:
```
TypeError(:typeassert, "", Float64, Dual{Nothing}(3.0455870596161523,0.045931680132707244))
```
and the stack trace is _really_ long...

The code below tests all implicit methods of `OrdinaryDiffEq` I could find in the documentation. The good new is that all other methods seem to work fine - which is already great!
```Julia
using OrdinaryDiffEq, DiffEqSensitivity
import Zygote

# all implicit solvers
solvers = [
    # SDIRK Methods (ok)
    ImplicitEuler(),
    ImplicitMidpoint(),
    Trapezoid(),
    TRBDF2(),
    SDIRK2(),
    Kvaerno3(),
    KenCarp3(),
    Cash4(),
    Hairer4(),
    Hairer42(),
    Kvaerno4(),
    KenCarp4(),
    Kvaerno5(),
    KenCarp5(),
    # Fully-Implicit Runge-Kutta Methods (FIRK)
    RadauIIA5(),
    # Fully-Implicit Runge-Kutta Methods (FIRK)
    PDIRK44(),
    # Rosenbrock Methods (all fail)
    ROS3P(),
    Rodas3(),
    RosShamp4(),
    Veldd4(),
    Velds4(),
    GRK4T(),
    GRK4A(),
    Ros4LStab(),
    Rodas4(),
    Rodas42(),
    Rodas4P(),
    Rodas5(),
    # Rosenbrock-W Methods (all fail)
    Rosenbrock23(),
    Rosenbrock32(),
    RosenbrockW6S4OS(),
    ROS34PW1a(),
    ROS34PW1b(),
    ROS34PW2(),
    ROS34PW3(),
    # Stabilized Explicit Methods (ok)
    ROCK2(),
    ROCK4(),
    RKC(),
    # SERK2v2(), not defined?
    ESERK5()];


function test()
    errors_solution = Dict()
    errors_gradient = Dict()

    p = rand(3)

    function dudt(u,p,t)
        u .* p
    end

    for solver in solvers
        function loss(p)
            prob = ODEProblem(dudt, [3.0, 2.0, 1.0], (0.0, 1.0), p)
            sol = concrete_solve(prob, solver, dt=0.01)
            sum(abs2, Array(sol))
        end

        try
            loss(p)
        catch e
[...]

PR #263, opened 2020-05-30
Allow stiff solver AD in adjoints

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/198, fixes https://github.com/SciML/DiffEqFlux.jl/issues/31, fixes https://github.com/SciML/DiffEqFlux.jl/issues/271

### DIFF
--- a/test/local_sensitivity/stiff_adjoints.jl
+++ b/test/local_sensitivity/stiff_adjoints.jl
@@ -165,7 +165,7 @@ solvers = [
     RadauIIA5(),
     # Fully-Implicit Runge-Kutta Methods (FIRK)
     PDIRK44(),
-    # Rosenbrock Methods (all fail)
+    # Rosenbrock Methods
     ROS3P(),
     Rodas3(),
     RosShamp4(),
--- a/test/local_sensitivity/stiff_adjoints.jl
+++ b/test/local_sensitivity/stiff_adjoints.jl
@@ -1,0 +1,212 @@
+using Zygote, DiffEqSensitivity, OrdinaryDiffEq, ForwardDiff, Test
+
+function lotka_volterra(u, p, t)
+  x, y = u
+  α, β, δ, γ = p
+  [α * x - β * x * y,-δ * y + γ * x * y]
+end
+
+function lotka_volterra(du, u, p, t)
+  x, y = u
+  α, β, δ, γ = p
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
We do. This just adds fallbacks where, if the caches don't match types, whatever we make a new tape and do it. I'll open an issue about adding caches for those fallback parts, but getting something working is higher priority and, given the current state of AD, probably doesn't make all that much of a difference to the final time.

YingboMa:
Yes, for sure.

test/local_sensitivity/stiff_adjoints.jl
```diff
@@ -0,0 +1,212 @@
...
+    Hairer42(),
+    Kvaerno4(),
+    KenCarp4(),
+    Kvaerno5(),
+    KenCarp5(),
+    # Fully-Implicit Runge-Kutta Methods (FIRK)
+    RadauIIA5(),
+    # Fully-Implicit Runge-Kutta Methods (FIRK)
+    PDIRK44(),
+    # Rosenbrock Methods (all fail)
```
YingboMa, line 168: Do Rosenbrock methods still fail after the fix?

ChrisRackauckas: nope, they all work now.

ChrisRackauckas: that is just copy-pasted from an issue.

YingboMa, line 168:
```suggestion
```

test/local_sensitivity/stiff_adjoints.jl
```diff
@@ -0,0 +1,212 @@
...
+    Veldd4(),
+    Velds4(),
+    GRK4T(),
+    GRK4A(),
+    Ros4LStab(),
+    Rodas4(),
+    Rodas42(),
+    Rodas4P(),
+    Rodas5(),
+    # Rosenbrock-W Methods (all fail)
```
YingboMa, line 181:
```suggestion
    # Rosenbrock-W Methods
```

ChrisRackauckas:
It passed but just went over time. I'll merge and stuff so that everything downstream is fixed, but add a new test group for the really long stiff solver tests that cover like every method.

Issue #257, opened 2020-05-24
inplace support for SDE Adjoint


frankschae:
These lines
 ```
 for (i, λi) in enumerate(λ)
      _, back = Zygote.pullback(y, prob.p) do u, p
        f(u, p, t)[i]
      end
      _,tmp2 = back(λi)
      dgrad[:,i] .= vec(tmp2)
    end
```
in backsolve_adjoint.jl require an oop function `f ` . Inplace version should be added similar to vecjacobian! code.

ChrisRackauckas:
In fact, we should make this have a ReverseDiffNoise, ZygoteNoise, etc. kind of setup like the other part. It's a little tedious though, and Tracker is kind of dead so we might as well just have those two and forward mode.

PR #265, opened 2020-05-31
iip tests

frankschae:
These are the inplace tests for the SDE adjoints (see issue https://github.com/SciML/DiffEqSensitivity.jl/issues/257). Only slight modifications in the src files had to be made. 

There is a somewhat strange behavior (not directly related to the Adjoints) because of that behavior it took some time to see that the tests are actually ok.

```
Random.seed!(seed)
prob_sde = SDEProblem(f!,σ!,u₀,trange,p2)
sol_sde = solve(prob_sde,RKMil(interpretation=:Stratonovich),dt=tend/1e6,adaptive=false, save_noise=true)
# at first evaluation this part is correct, if evaluated a second time,  the second last noise value changes and the result is wrong
noise = [sol_sde.W(t)[1][1] for t in tarray]
resp1 = sum(tarray.*u₀.^2 .*exp.(2*p2[1].*tarray+2*p2[2].*noise))
resp2 = sum(noise.*u₀.^2 .*exp.(2*p2[1].*tarray+2*p2[2].*noise))
resp = [resp1, resp2]
resu0 = sum(u₀.* exp.(2*p2[1].*tarray+2*p2[2].*noise))
```

If running the lines starting from 'noise = '  a second time, the second last noise value gets the same value as the last one, such that the final result is off.

### DIFF
--- a/src/local_sensitivity/adjoint_common.jl
+++ b/src/local_sensitivity/adjoint_common.jl
@@ -96,7 +96,7 @@ function adjointdiffcache(g,sensealg,discrete,sol,dg,f;quad=false,noiseterm=fals
 
     if prob isa DiffEqBase.SteadyStateProblem
        if DiffEqBase.isinplace(prob)
-         tape = ReverseDiff.GradientTape((y, prob.p)) do u,p
+        tape = ReverseDiff.GradientTape((y, prob.p)) do u,p
           du1 = similar(p, size(u))
           f(du1,u,p,nothing)
           return vec(du1)
@@ -109,13 +109,13 @@ function adjointdiffcache(g,sensealg,discrete,sol,dg,f;quad=false,noiseterm=fals
     else
       if DiffEqBase.isinplace(prob)
         tape = ReverseDiff.GradientTape((y, prob.p, [tspan[2]])) do u,p,t
-         du1 = similar(p, size(u))
-         f(du1,u,p,first(t))
-         return vec(du1)
+          du1 = similar(p, size(u))
+          f(du1,u,p,first(t))
+          return vec(du1)
        end
      else
        tape = ReverseDiff.GradientTape((y, prob.p, [tspan[2]])) do u,p,t
-         vec(f(u,p,first(t)))
+          vec(f(u,p,first(t)))
        end
      end
     end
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
>If running the lines starting from 'noise = ' a second time, the second last noise value gets the same value as the last one, such that the final result is off.

That's odd, it could be due to one of the end result fixes like https://github.com/SciML/StochasticDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L198-L214 . Not a super high priority fix, but it should probably get at least an issue, and if it's not hard it would be nice to have it fixed.

frankschae:
The issue is linked to `DiffEqNoiseProcess`. The sensible check is:
```Julia
u₀ = [0.5]
tstart = 0.0
tend = 1.0
dt = 0.1
trange = (tstart, tend)
t = tstart:dt:tend
tarray = collect(t)

p2 = [1.01,0.87]
f!(du,u,p,t) = du.=p[1]*u
σ!(du,u,p,t) = du.=p[2]*u
Random.seed!(seed)
prob_sde = SDEProblem(f!,σ!,u₀,trange,p2)
sol_sde = solve(prob_sde,RKMil(interpretation=:Stratonovich),dt=tend/1e1,adaptive=false, save_noise=true)
length(sol_sde.W.W) #this will the same length as tarray
for i in 1:1:length(tarray)
  t = tarray[i]
  sol_sde.W(t)
  println(" ")
  #@show t, sol_sde.W(t)[1][1]
end
length(sol_sde.W.W) #this will be larger
```

The arrays are getting larger because searchsortedfirst() from the interpolate! function doesn't hit the right index (https://github.com/SciML/DiffEqNoiseProcess.jl/blob/78f9fe2149c88cb326c625a86e9a75856a7bc1fc/src/noise_interfaces/noise_process_interface.jl#L353):

```Julia
a = [0.0, 0.1, 0.2, 0.30000000000000004, 0.4, 0.5, 0.6, 0.7, 0.7999999999999999, 0.8999999999999999, 0.9999999999999999]
searchsortedfirst(a,  0.7) # 8
searchsortedfirst(a,  0.8) # 10
```

for inplace definitions this overwrites some noise values.

Issue #254, opened 2020-05-24
SDE Adjoint Scalar Noise


ChrisRackauckas:
We should make sure that that SDE adjoint handles the scalar noise case

PR #268, opened 2020-06-01
SDE Adjoint for scalar noise

frankschae:
This adds a switch for scalar noise (associated issue: https://github.com/SciML/DiffEqSensitivity.jl/issues/254).

For scalar noise, we don't have to consider the full noise Jacobian broadcasted against lambda. Since all dWs are equal for scalar noise, the computation simplifies to another vjp, and can be calculated using `vecjacobian! `with `dgrad=dgrad`. Thus in case of scalar noise, also the choice of the solver is less restrictive.

### DIFF
--- a/test/local_sensitivity/sde.jl
+++ b/test/local_sensitivity/sde.jl
@@ -30,8 +30,6 @@ end
 
 p2 = [1.01,0.87]
 
-
-
 @testset "SDE oop Tests" begin
   f_oop_linear(u,p,t) = p[1]*u
   σ_oop_linear(u,p,t) = p[2]*u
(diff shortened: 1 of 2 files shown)


test/local_sensitivity/sde.jl
```diff
@@ -298,29 +296,31 @@ end
...
-  @test isapprox(res_sde_u0 ,res_sde_u02, rtol = 1e-5)
-
-  res_sde_u02, res_sde_p2 = adjoint_sensitivities(sol_sde,EulerHeun(),dg!,Array(t)
-      ,dt=tend/1e6,adaptive=false,sensealg=BacksolveAdjoint(noise=DiffEqSensitivity.ReverseDiffNoise()))
-
-  @info res_sde_p2
-
-  @test isapprox(res_sde_p, res_sde_p2, rtol = 1e-10)
-  @test isapprox(res_sde_u0 ,res_sde_u02, rtol = 1e-10)
+  ### The following tests are not needed anymore here, as for scalar SDEs it is sufficient to compute the vjps for the dsigma/dp terms
```
ChrisRackauckas, line 299: then delete them?

ChrisRackauckas: Looks to me like it's still a useful test though. It's stressing a different part of the adjoint method. Technically it no user should be using this here because it is a scalar SDE, but it's hitting a code path that would be used on larger examples.

PR #270, opened 2020-06-08
fix extract_local_sensitivities

YingboMa:


src/local_sensitivity/forward_sensitivity.jl
```diff
@@ -205,7 +205,15 @@ function extract_local_sensitivities(sol,::ForwardDiffSensitivity, ::Val{false})
   _sol = adapt(eltype(sol.u),sol)
   u = ForwardDiff.value.(_sol)
   du_full = ForwardDiff.partials.(_sol)
-  return u, [[du_full[i,j][k] for i in 1:size(du_full,1), j in 1:size(du_full,2)] for k in 1:length(du_full[1])]
+  @show size(du_full)
```
ChrisRackauckas, line 208:
```suggestion
```

src/local_sensitivity/forward_sensitivity.jl
```diff
@@ -205,7 +205,15 @@ function extract_local_sensitivities(sol,::ForwardDiffSensitivity, ::Val{false})
...
+  firststate = first(du_full)
```
ChrisRackauckas, line 209:
```suggestion
```

src/local_sensitivity/forward_sensitivity.jl
YingboMa: Oh haha

src/local_sensitivity/forward_sensitivity.jl
YingboMa: `firststate` is actually used.

ChrisRackauckas: oh I see

PR #275, opened 2020-06-09
SDE Adjoint for non-diagonal noise and diagonal noise with mixing terms

frankschae:
I have added one branch for non-diagonal noise processes and one for diagonal noise processes where the noise is more general than du[I] = f(u[I]).

The second one is controlled by 
`isnoisemixing(S.sensealg)`.
By default it currently assumes that the noise is not mixing the terms.

For the ForwardDiff version, I have some problems when using the JacobianConfig option. Therefore, in this part of  `_jacNoise!()` I am currently using directly:
`ForwardDiff.jacobian!(J,uf,dy,y)`
(instead of `jacobian!(J,uf, dy, f_cache, sensealg, jac_config)` )

### DIFF
--- a/src/local_sensitivity/backsolve_adjoint.jl
+++ b/src/local_sensitivity/backsolve_adjoint.jl
@@ -22,25 +22,38 @@ function (S::ODEBacksolveSensitivityFunction)(du,u,p,t)
   @unpack y, prob, f, discrete = S
   idx = length(y)
 
+  λ     = @view u[1:idx]
+  grad  = @view u[idx+1:end-idx]
+  _y    = @view u[end-idx+1:end]
+
   if length(u) == length(du)
     # ODE/Drift term and scalar noise
-    λ     = @view u[1:idx]
-    grad  = @view u[idx+1:end-idx]
-    _y    = @view u[end-idx+1:end]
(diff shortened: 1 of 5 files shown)


src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -112,15 +127,14 @@ end
 @noinline function SDEAdjointProblem(sol,sensealg::BacksolveAdjoint,
                                      g,t=nothing,dg=nothing;
                                      checkpoints=sol.t,
-                                     callback=CallbackSet(),kwargs...)
+                                     callback=CallbackSet(),
+                                     diffusion_jac=nothing, diffusion_paramjac=nothing,kwargs...)
```
ChrisRackauckas, line 131: assume these will never be called directly. These would need to be added to the `SDEFunction` to be used.

frankschae: to the `SDEFunction` or `ODEFunction`? 
They are indeed not yet tested (or used and could be left out for the moment) -- I thought that something like this might be a nice way to encode if one of the two Jacobians dsigma/dx or dsigma/dp is zero or constant (to speed up the computation).

PR #276, opened 2020-06-09
add tests for adjoints with sparse jacobians

omalled:
This is an attempt to create tests for the adjoints on differential equations with sparse Jacobians. Currently, the sparsity is not exploited on the backward pass, making these adjoint calculations not performant.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -47,6 +47,7 @@ ZygoteRules = "0.2"
 julia = "1.2"
 
 [extras]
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
This is a pretty awesome set of tests. It's hard to imagine a set of tests that will fail if the adjoint is not using sparsity, but this does it. I'll open an issue to track our progress on it, but it can be merged to that way we can more easily keep it up to date.

PR #284, opened 2020-06-15
Add jac_prototype for adjoint

### LINKED ISSUE #283: Sparsity in the reverse pass
https://github.com/SciML/DiffEqSensitivity.jl/pull/276 demonstrates that we do not exploit sparsity in the implicit methods in the reverse pass. We need to derive what the correct sparsity patterns for the adjoints based on the sparsity pattern of `f` and use that to generate a `jac_prototype` for the reverse. 

@omalled @YingboMa

YingboMa:
Fix #283

`jac_prototype` for quadrature adjoint is straightforward, it's just `jac_prototype'`. `jac_prototype` for interpolating adjoint is
```
[J' 0]
[0  I]
```
since the argument integral is pointwise, the Jacobian is naturally `I`.

Backsolve's `jac_prototype` is
```
[J' 0  0]
[0  I  0]
[0  0  J]
```

### DIFF
--- a/src/local_sensitivity/interpolating_adjoint.jl
+++ b/src/local_sensitivity/interpolating_adjoint.jl
@@ -122,10 +122,23 @@ end
   if original_mm === I || original_mm === (I,I)
     mm = I
   else
-    mm = zeros(len, len)
-    copyto!(@view(mm[1:numstates, 1:numstates]), sol.prob.f.mass_matrix')
-    copyto!(@view(mm[numstates+1:end, numstates+1:end]), I)
+    adjmm = sol.prob.f.mass_matrix'
+    zzz = zero(adjmm)
+    mm = [adjmm zzz
+          zzz   adjmm]
   end
-  odefun = ODEFunction(sense, mass_matrix=mm)
+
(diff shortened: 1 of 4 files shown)


src/local_sensitivity/interpolating_adjoint.jl
```diff
@@ -123,10 +122,23 @@ end
...
+
+  jac_prototype = sol.prob.f.jac_prototype
+  if !sense.discrete || jac_prototype === nothing
+    adjoint_jac_prototype = nothing
+  else
+    _adjoint_jac_prototype = jac_prototype'
+    zzz = zero(_adjoint_jac_prototype)
+    augmented_integral_jac_prototype = Diagonal(I, len-numstates)
+    adjoint_jac_prototype = [_adjoint_jac_prototype zzz
+                             zzz                    augmented_integral_jac_prototype]
```
YingboMa, line 139: Ah, this is not quite right. It's tough to construct this kind of matrix in an efficient way generically.

ChrisRackauckas:
why not backsolve adjoint?

YingboMa:
@ChrisRackauckas ready for another review

PR #285, opened 2020-06-18
Add the missing dgdp term in the sensitivity integral

### LINKED ISSUE #92: Sensitivities do not match when the running cost explicitly depends on the parameters
When the running-cost function, g, depends explicitly on the system parameters, p. I get a mismatch between the gradients reported by *adjoint_sensitivities* and *Calculus.gradient*.

I suspect this discrepancy is due to the fact that I calculate dg as the derivative of g w.r.t. only the state, and omit its derivatives w.r.t. the parameters, p.

Below, I provide a minimal example. The dynamics is that of a simple pendulum, controlled at the upward equilibrium point.

An observation: if I make the function g independent of the parameters, then the derivatives do match.

```
using DifferentialEquations.OrdinaryDiffEq
using DiffEqSensitivity
using DiffEqBase
using Calculus
using ForwardDiff
using QuadGK

function pendulum_eom(dx, x, p, t)
    dx[1] = x[2]
    dx[2] = -sin(x[1]) + (-p[1]*sin(x[1]) + p[2]*x[2])  # Second term is a simple controller that stabilizes π
end

x0 = [0.1, 0.0]
tspan = (0.0, 10.0)
p = [-24.05, -19.137]
​
prob = ODEProblem(pendulum_eom, x0, tspan, p)
sol = solve(prob, Vern9(), abstol=1e-8, reltol=1e-8)

g(x, p, t) = 1.0*(x[1] - π)^2 + 1.0*x[2]^2 + 5.0*(-p[1]*sin(x[1]) + p[2]*x[2])^2
dg(out, y, p, t) = ForwardDiff.gradient!(out, x -> g(x, p, t), y)

res = adjoint_sensitivities(sol,Vern9(),g,nothing,dg,abstol=1e-8,
                                reltol=1e-8,iabstol=1e-8,ireltol=1e-8)

function G(p)
    tmp_prob = remake(prob,p=p)
    sol = solve(tmp_prob,Vern9(),abstol=1e-8,reltol=1e-8)
    res,err = quadgk((t)-> g(sol(t), p, t), 0.0,10.0,atol=1e-8,rtol=1e-8)
    res
end
res2 = Calculus.gradient(G,p)

res' ≈ res2

```

YingboMa:
Fix #92

I did not set up the AD code for computing `dgdp` and the `dgdp` term for the quadrature adjoint. Do you want to do that part? @ChrisRackauckas

### DIFF
--- a/src/local_sensitivity/backsolve_adjoint.jl
+++ b/src/local_sensitivity/backsolve_adjoint.jl
@@ -72,7 +72,7 @@ function (S::ODEBacksolveSensitivityFunction)(du,u,p,t)
 
   dλ .*= -one(eltype(λ))
 
-  discrete || accumulate_dgdu!(dλ, y, p, t, S)
+  discrete || accumulate_dgdu!(dλ, y, p, t, S, dgrad)
   return nothing
 end
 
--- a/src/local_sensitivity/backsolve_adjoint.jl
+++ b/src/local_sensitivity/backsolve_adjoint.jl
@@ -72,7 +72,7 @@ function (S::ODEBacksolveSensitivityFunction)(du,u,p,t)
 
   dλ .*= -one(eltype(λ))
 
-  discrete || accumulate_dgdu!(dλ, y, p, t, S, dgrad)
+  discrete || accumulate_cost!(dλ, y, p, t, S, dgrad)
   return nothing
 end
 
--- a/src/local_sensitivity/derivative_wrappers.jl
+++ b/src/local_sensitivity/derivative_wrappers.jl
@@ -468,7 +468,7 @@ function _jacNoise!(λ, y, p, t, S::SensitivityFunction, isnoise::ZygoteNoise, d
 end
 
 
-function accumulate_dgdu!(dλ, y, p, t, S::SensitivityFunction, dgrad=nothing)
+function accumulate_cost!(dλ, y, p, t, S::SensitivityFunction, dgrad=nothing)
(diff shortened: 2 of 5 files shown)


src/local_sensitivity/backsolve_adjoint.jl
```diff
@@ -72,7 +72,7 @@ function (S::ODEBacksolveSensitivityFunction)(du,u,p,t)
 
   dλ .*= -one(eltype(λ))
 
-  discrete || accumulate_dgdu!(dλ, y, p, t, S)
+  discrete || accumulate_dgdu!(dλ, y, p, t, S, dgrad)
```
ChrisRackauckas, line 75: we need this in all of them though?

YingboMa: Yeah, I only fixed the continuous case. Since you are more familiar with the AD code, do you want to take it from here?

YingboMa:
The missing dg/dp term is in eq (18) https://github.com/SciML/DiffEqDevMaterials/blob/master/dae_adjoint/output/main.pdf

src/local_sensitivity/backsolve_adjoint.jl
ChrisRackauckas: I meant, the same change is all that's needed for interpolatingadjoint, right?

ChrisRackauckas: and something very similar quadratureadjoint, so let's just do all of the continuous cases here.

Issue #288, opened 2020-06-19
Inplace dgdu and dgdp output length issues with continuous adjoint sensitivities


mjohnson541:
It seems like the inplace dgdp might be being fed a vector the length of y instead of p resulting in an error in the dgdp calculation. The array length for all of these out of bounds errors has consistently been the lenght of y each time I encountered it. 
```
using DifferentialEquations
using DiffEqSensitivity
using Sundials
using ForwardDiff

p = [2.0,3.0]
u0 = [2.0]
function f(du,u,p,t)
    du[1] = -u[1]*p[1]-p[2]
end

prob = ODEProblem(f,u0,(0.0,1.0),p)
sol = solve(prob,CVODE_BDF(),abstol=1e-10,reltol=1e-10);

g(u,p,t) = -u[1]*p[1]-p[2]

dgdu(out, y, p, t) = ForwardDiff.gradient!(out, y -> g(y, p, t), y)
dgdp(out, y, p, t) = ForwardDiff.gradient!(out, p -> g(y, p, t), p)
du0,dp = adjoint_sensitivities(sol,Vern9(),g,nothing,(dgdu,dgdp);abstol=1e-10,reltol=1e-10)
```

Gives traceback: 
```
BoundsError: attempt to access 1-element Array{Float64,1} at index [Base.OneTo(2)]

Stacktrace:
 [1] throw_boundserror(::Array{Float64,1}, ::Tuple{Base.OneTo{Int64}}) at ./abstractarray.jl:538
 [2] checkbounds at ./abstractarray.jl:503 [inlined]
 [3] copyto! at ./multidimensional.jl:902 [inlined]
 [4] extract_gradient! at /Users/mattjohnson/.julia/packages/ForwardDiff/cXTw0/src/gradient.jl:77 [inlined]
 [5] vector_mode_gradient!(::Array{Float64,1}, ::var"#18#19"{Array{Float64,1},Float64}, ::Array{Float64,1}, ::ForwardDiff.GradientConfig{ForwardDiff.Tag{var"#18#19"{Array{Float64,1},Float64},Float64},Float64,2,Array{ForwardDiff.Dual{ForwardDiff.Tag{var"#18#19"{Array{Float64,1},Float64},Float64},Float64,2},1}}) at /Users/mattjohnson/.julia/packages/ForwardDiff/cXTw0/src/gradient.jl:104
[...]

PR #291, opened 2020-06-20
properly size the dg/dp accumulation

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/288


ChrisRackauckas:
Tested locally is fine. The stuff on CI is just Zygote versioning issues.

Issue #251, opened 2020-05-24
InterpolatingAdjoint for SDEs


ChrisRackauckas:
The same basic concept as BacksolveAdjoint can be done for SDEs

@frankschae

PR #295, opened 2020-07-08
Interpolating adjoint

frankschae:
This PR adds the `InterpolatingAdjoint` method for SDEs from issue #251. It is (like BacksolveAdjoint) based on  `NoiseGrid` at the moment. 

The tests cover scalar, diagonal, diagonal mixing, and non-diagonal noise processes for all possible options of the noise process. I didn't include all possibilities for the vjps so far . For `noise=false ` I had to put some tests_broken/ relax the testing case quite a bit in a few cases.  

It currently handles floating point issues by checking if the checkpoints are unique:
```Julia
if length(unique(round.(checkpoints, digits=13))) != length(checkpoints)
    @warn "The given checkpoints are not unique. To avoid issues in the interpolation the checkpoints were redefined. You may want to check sol.t if default checkpoints were used."
    checkpoints = unique(round.(checkpoints, digits=13))
  end
```

### DIFF
--- a/src/local_sensitivity/interpolating_adjoint.jl
+++ b/src/local_sensitivity/interpolating_adjoint.jl
@@ -33,10 +33,13 @@ function ODEInterpolatingAdjointSensitivityFunction(g,sensealg,discrete,sol,dg,f
     if typeof(sol.prob) <: SDEProblem
       # replicated noise
       _sol = deepcopy(sol)
-      idx1 = searchsortedfirst(_sol.t, interval[1])
+      idx1 = searchsortedfirst(_sol.t, interval[1]-1000eps(interval[1]))
       idx2 = searchsortedfirst(_sol.t, interval[2])
       forwardnoise = DiffEqNoiseProcess.NoiseGrid(_sol.t[idx1:idx2], _sol.W.W[idx1:idx2])
-      dt = maximum([abs(_sol.t[end]-_sol.t[end-1]), interval[2] - interval[1]])
+      dt = abs(_sol.W.dt)
+      if dt < 10000eps(_sol.t[end])
+        dt = interval[2] - interval[1]
+      end
       cpsol = solve(remake(sol.prob, tspan=interval, u0=sol(interval[1]), noise=forwardnoise), sol.alg, save_noise=false; dt=dt, tols...)
     else
       cpsol = solve(remake(sol.prob, tspan=interval, u0=sol(interval[1])), sol.alg; tols...)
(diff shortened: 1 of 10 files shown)


src/local_sensitivity/interpolating_adjoint.jl
```diff
@@ -27,17 +29,31 @@ function ODEInterpolatingAdjointSensitivityFunction(g,sensealg,discrete,sol,dg,c
...
-    cpsol = solve(remake(sol.prob, tspan=interval, u0=sol(interval[1])), sol.alg; tols...)
+
+    if typeof(sol.prob) <: SDEProblem
+      # replicated noise
+      _sol = deepcopy(sol)
+      idx1 = searchsortedfirst(_sol.t, interval[1]-1000eps(interval[1]))
+      idx2 = searchsortedfirst(_sol.t, interval[2])
+      forwardnoise = DiffEqNoiseProcess.NoiseGrid(_sol.t[idx1:idx2], _sol.W.W[idx1:idx2])
+      dt = abs(_sol.W.dt)
+      if dt < 10000eps(_sol.t[end])
```
ChrisRackauckas, line 40: 10000 is a bit high. Is 100 not sufficient?

frankschae: 100 doesn't change dt in the test case. 1000 seems to work..  It is not yet optimal.. the unique check is also sensitive to the discretization.. for 1e-9 for example, the current version would need a different number of digits when checking the uniqueness ..

```Julia
using DiffEqSensitivity, StochasticDiffEq
using Random
seed = 100
Random.seed!(seed)
u₀ = [0.5]
tstart = 0.0
tend = 0.1
dt = 0.005
trange = (tstart, tend)
t = tstart:dt:tend
tarray = collect(t)

function g(u,p,t)
  sum(u.^2.0/2.0)
end

function dg!(out,u,p,t,i)
  (out.=-u)
end

p2 = [1.01,0.87]


f_oop_linear(u,p,t) = p[1]*u
σ_oop_linear(u,p,t) = p[2]*u

dt1 = tend/1e4

Random.seed!(seed)
prob_oop = SDEProblem(f_oop_linear,σ_oop_linear,u₀,trange,p2)
sol_oop = solve(prob_oop,EulerHeun(),dt=dt1,adaptive=false,save_noise=true)

@show length(sol_oop)

checkpoints=sol_oop.t

if length(unique(round.(checkpoints, digits=13))) != length(checkpoints)
  @warn "The given checkpoints are not unique. To avoid issues in the interpolation the checkpoints were redefined. You may want to check sol.t if default checkpoints were used."
  checkpoints = unique(round.(checkpoints, digits=13))
end
@show length(checkpoints)

intervals = map(tuple, @view(checkpoints[1:end-1]), @view(checkpoints[2:end]))
interval_end = intervals[end][end]
prob_oop.tspan[1] > interval_end && push!(intervals, (interval_end, prob_oop.tspan[1]))
cursor = lastindex(intervals)
interval = intervals[cursor]


_sol = deepcopy(sol_oop)
idx1 = searchsortedfirst(_sol.t, interval[1]-1000eps(interval[1]))
idx2 = searchsortedfirst(_sol.t, interval[2])
dt = abs(_sol.W.dt)
if dt < 1000eps(_sol.t[end])
  dt = interval[2] - interval[1]
end
@show dt
```

ChrisRackauckas:
might just need another test group

src/local_sensitivity/interpolating_adjoint.jl
```diff
@@ -27,17 +29,31 @@ function ODEInterpolatingAdjointSensitivityFunction(g,sensealg,discrete,sol,dg,c
...
     cursor = lastindex(intervals)
     interval = intervals[cursor]
-    cpsol = solve(remake(sol.prob, tspan=interval, u0=sol(interval[1])), sol.alg; tols...)
+
+    if typeof(sol.prob) <: SDEProblem
+      # replicated noise
+      _sol = deepcopy(sol)
+      idx1 = searchsortedfirst(_sol.t, interval[1]-1000eps(interval[1]))
+      idx2 = searchsortedfirst(_sol.t, interval[2])
+      forwardnoise = DiffEqNoiseProcess.NoiseGrid(_sol.t[idx1:idx2], _sol.W.W[idx1:idx2])
```
ChrisRackauckas, line 38: let's make sure to generalize this soon

PR #298, opened 2020-07-10
added citations for backsolve, steadystate and interpolation adjoint

frankschae:
Is this: 
Lev Semenovich Pontryagin, EF Mishchenko, VG Boltyanskii, and RV Gamkrelidze. The mathematical theory of optimal processes. 1962.

the old reference you mentioned earlier, @ChrisRackauckas  ? This was cited in the Neural ODE paper http://papers.nips.cc/paper/7892-neural-ordinary-differential-equations.pdf


ChrisRackauckas:
yeah that was it.

Issue #253, opened 2020-05-24
More general noise types for the reverse in SDE adjoints


ChrisRackauckas:
Currently it hardcodes to reverse via a NoiseGrid. In some cases this can lose accuracy since it's only using the samples points from the forward pass. One https://github.com/SciML/DiffEqNoiseProcess.jl/issues/50 is complete, it should by default just reverse the current noise pass and keep the type the user chose.

ChrisRackauckas:
Now it's only seen in the InterpolatingAdjoint checkpointing

PR #299, opened 2020-07-13
replace NoiseGrid by NoiseWrapper

frankschae:
Replaces all calls to `NoiseGrid` by `NoiseWrapper`, see https://github.com/SciML/DiffEqSensitivity.jl/issues/253 + https://github.com/SciML/DiffEqSensitivity.jl/issues/253

### DIFF
--- a/src/local_sensitivity/backsolve_adjoint.jl
+++ b/src/local_sensitivity/backsolve_adjoint.jl
@@ -185,7 +185,7 @@ end
 
   # replicated noise
   _sol = deepcopy(sol)
-  backwardnoise = DiffEqNoiseProcess.NoiseGrid(reverse!(_sol.t),reverse!( _sol.W.W))
+  backwardnoise = DiffEqNoiseProcess.NoiseWrapper(_sol.W, reverse=true)
 
   if StochasticDiffEq.is_diagonal_noise(sol.prob) && typeof(sol.W[end])<:Number
     # scalar noise case
--- a/src/local_sensitivity/interpolating_adjoint.jl
+++ b/src/local_sensitivity/interpolating_adjoint.jl
@@ -88,7 +88,7 @@ function (S::ODEInterpolatingAdjointSensitivityFunction)(du,u,p,t)
       end
       if typeof(sol.prob) <: SDEProblem
         idx1 = searchsortedfirst(sol.t, interval[1])
-        forwardnoise = DiffEqNoiseProcess.NoiseWrapper(_sol.W, indx=idx1)
+        forwardnoise = DiffEqNoiseProcess.NoiseWrapper(sol.W, indx=idx1)
         prob′ = remake(prob, tspan=intervals[cursor′], u0=y, noise=forwardnoise)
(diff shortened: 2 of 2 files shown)


frankschae:
InterpolatingAdjoint became much slower -- but it still executes locally without any errors (fourth line is the one which failed on Travis). I'll check why it takes much longer. Maybe it is because while `NoiseGrid` only stored the time points with their associated noise values in each interval, NoiseWrapper stores the entire process in `source`.  
```
[ Info: SDE Non-Diagonal Noise Adjoints

[ Info: [0.736545474290214 0.6809015165765869 0.5373209107227818 0.4991686900379727]
[ Info: [0.736545474290214 0.6809015165765869 0.5373209107227818 0.4991686900379727]
[ Info: [0.736545474290214 0.6809015165765869 0.5373209107227818 0.4991686900379727]
[ Info: [0.736545384426385 0.6809014328620033 0.5373207511401008 0.499168541442926]
```

ChrisRackauckas:
The way that new values have to get added to the Brownian process is probably really slow on memory. Can you profile it and see what you find? There should be like an `insert at` which essentially ends up being an O(n) data movement, which is really bad. `NoiseGrid` might effectively be better because of that, though with less accuracy because of how it's interpolating linearly instead of respecting the variance of the Brownian bridge (but linear is the mean, so it's good). Having an option here would probably be good, with a default to maybe NoiseGrid if it's accurate enough.


[What we should probably do is abstract the interface a little bit more, so that any noise process can be used. I.e. `reversenoise!(noise)` returns a NoiseWrapper if it's a NoiseProcess, returns a reversed NoiseGrid if the input was a NoiseGrid, if it's a NoiseFunction then it just returns the original, etc. ]

frankschae:
Is there a way to improve the speed at the first compilation?

I find that `InterpolatingAdjoint` (e.g., 
 `res_sde_u0a, res_sde_pa = @time adjoint_sensitivities(soloop,EulerHeun(),dg!,Array(t)
    ,dt=dtnd,adaptive=false,sensealg=InterpolatingAdjoint(noise=DiffEqSensitivity.ZygoteNoise()))`
) takes 
99.663339 seconds (288.24 M allocations: 100.984 GiB, 15.38% gc time)
when it is run the first time but afterwards it only takes ~0.05s ( 0.047974 seconds (921.68 k allocations: 39.726 MiB)). While the first compilation is almost 10 times slower than a similar `BacksolveAdjoint`
11.399672 seconds (33.75 M allocations: 7.513 GiB, 11.45% gc time)
the runs afterwards have almost the same speed ~ 0.04s (0.043630 seconds (855.80 k allocations: 35.445 MiB))

ChrisRackauckas:
A huge amount of it will be covered by https://github.com/JuliaLang/www.julialang.org/pull/794 , so I wouldn't worry about it for now and just carry forward thinking it'll get fixed to a large extent.

ChrisRackauckas:
Merged this, but Github is being weird.

PR #301, opened 2020-07-15
fix shaped adjoints

ChrisRackauckas:


ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqFlux.jl/issues/360

Issue #300, opened 2020-07-14
Checkpointing flag makes no difference on BacksolveAdjoint gradients


samuela:
I was surprised to find today that the `checkpoint` argument ([docs here](https://docs.sciml.ai/stable/analysis/sensitivity/#Sensitivity-Algorithms-1)) to the `BacksolveAdjoint` sensealg has no real effect on the gradient output.

Here's a reproduction:
```julia
import DifferentialEquations: Tsit5
import DiffEqFlux: FastDense, ODEProblem, solve
import DiffEqSensitivity: BacksolveAdjoint, ODEAdjointProblem
import LinearAlgebra: I
import ControlSystems

# Changing x_dim or T doesn't make any difference :(
x_dim = 2
T = 40.0

A, B, Q, R = -I, I, I, I
cost = (x, u) -> x' * Q * x + u' * R * u
K = ControlSystems.lqr(Matrix{Float64}(A, x_dim, x_dim), B, Q, R)
lqr_params = vcat(-K[:], zeros(x_dim))
policy = FastDense(x_dim, x_dim)

function dynamics!(dx, x, policy_params, t)
    dx .= A * x + B * policy(x, policy_params)
end

function backsolve_grad(x0, policy_params, checkpointing)
    fwd_sol = solve(
        ODEProblem(dynamics!, x0, (0, T), policy_params),
        Tsit5(),
        u0 = x0,
        p = policy_params,
        dense = false,
    )
    bwd_sol = solve(
        ODEAdjointProblem(
            fwd_sol,
            BacksolveAdjoint(checkpointing = checkpointing),
            (x, policy_params, t) -> cost(x, policy(x, policy_params)),
        ),
        Tsit5(),
        dense = false,
        save_everystep = false,
    )

    # We can see that the gradients are exactly equal whereas the recovered x(0)
    # shown in the last x_dim dimensions changes when checkpointing is
    # enabled/disabled.
    @show fwd_sol.u[end]
    @show bwd_sol.u[end]

    # In the backsolve adjoint, the last x_dim dimensions are for the
    # reconstructed x state.
    bwd_sol.u[end][1:end-x_dim]
end

x0 = ones(x_dim)
backsolve_results = backsolve_grad(x0, lqr_params, false)
backsolve_checkpointing_results = backsolve_grad(x0, lqr_params, true)
@assert backsolve_results != backsolve_checkpointing_results
```
[...]

PR #303, opened 2020-07-15
fix checkpointing for BacksolveAdjoint

frankschae:
Replaces `DiscreteCallback` by  `PresetTimeCallback`. 

Tests on (a simplified version of) the example from issue #300 . With checkpointing enabled the gradients from `BacksolveAdjoint` and `InterpolatingAdjoint` agree.

### DIFF
--- a/src/local_sensitivity/backsolve_adjoint.jl
+++ b/src/local_sensitivity/backsolve_adjoint.jl
@@ -206,10 +206,6 @@ end
 function backsolve_checkpoint_callbacks(sensefun, sol, checkpoints, callback)
   prob = sol.prob
   cur_time = Ref(length(checkpoints))
-  condition = let checkpoints=checkpoints
-    (u,t,integrator) ->
-      checkpoints !== nothing && ((idx = searchsortedfirst(checkpoints, t)) <= length(checkpoints)) && checkpoints[idx] == t
-  end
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Unrelated test failure. We should make that more robust, but let's not have that block this.

frankschae:
I'll check the non diagonal tests later locally.  It might be that checkpointing does something bad on runtime there (since the first one is the `BacksolveAdjoint` and by default checkpointing is enabled).  When writing those tests, I had to use 1e6 time steps (`dtnd = tend/1e6`) to get accurate gradients. So while before the callback never triggered we now have 1e6 additional actions..  (which I think should not harm a lot). 

Anyways, it might be something good -- if checkpointing now helps here, we could maybe decrease the number of time steps. In the worst case, setting `checkpointing = false` should bring back the original runtimes.

ChrisRackauckas:
Can you do a follow up PR to start tweaking the tests to make Travis happy? I already tagged, so no one is blocked in usage, but we should make sure Travis doesn't fall behind here.

frankschae:
Yep, I'll play around later.

samuela:
Has this made it into a release yet, or should I depend on the github version of the package?

ChrisRackauckas:
it's in 6.24

samuela:
Awesome, I'll try upgrading!

PR #311, opened 2020-07-25
add DistributionsAD (again)

ChrisRackauckas:


devmotion:
Why is DistributionsAD required by DiffEqSensitivity? Is it due to some of the ReverseDiff implementation/piracy that has not been migrated to ReverseDiff yet?

ChrisRackauckas:
It's part of https://github.com/SciML/DiffEqSensitivity.jl/issues/316 . We need the broadcast overload, though it looks like it may be causing issues.

mohamed82008:
This should not be necessary after https://github.com/JuliaDiff/ReverseDiff.jl/pull/141.

ChrisRackauckas:
Awesome! I'll test it out.

PR #313, opened 2020-07-26
More flexible Δ

YingboMa:


ChrisRackauckas:
Merging since it looks to be correct, but we should get more tests on this. Notably, there's a few concrete_solve tests that need to be fixed, which we can follow up with in https://github.com/SciML/DiffEqSensitivity.jl/issues/315

PR #317, opened 2020-07-29
Ito sense for SDE adjoints

frankschae:
@ChrisRackauckas  Could you check if the way to add the Ito sense goes in the right direction? 

I thought an additional type for the transformed function with its own vips are the best way to do it. However, I couldn't resolve up to now all errors in the vjps from the SensitivityFunction. (ReverseDiff is the only one which I could make running through the code but also there the gradients look slightly off.. )


frankschae:
All tests in  `sde_transformation_test.jl` are passing. So the vips are doing the proper transformation on the drift term.

frankschae:
The plot below shows the evolution of the state (top left), derivative wrt initial state (top right), derivative wrt p[1] and p[2] (bottom left and right, respectively).
plot

PR #324, opened 2020-08-15
Fix compiled mode via pull_value!

ChrisRackauckas:
See the discussion in https://github.com/JuliaDiff/ReverseDiff.jl/issues/143

### DIFF
--- a/test/local_sensitivity/sde.jl
+++ b/test/local_sensitivity/sde.jl
@@ -37,13 +37,11 @@ p2 = [1.01,0.87]
   σ_oop_linear(u,p,t) = p[2]*u
 
   function f_oop_linear(u::Tracker.TrackedArray,p,t)
-    dx = p[1]*u[1]
-    Tracker.collect([dx])
+    p[1].*u
   end
 
   function σ_oop_linear(u::Tracker.TrackedArray,p,t)
-    dx = p[2]*u[1]
-    Tracker.collect([dx])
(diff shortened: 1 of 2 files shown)


test/local_sensitivity/sde.jl
```diff
@@ -37,13 +37,11 @@ p2 = [1.01,0.87]
   σ_oop_linear(u,p,t) = p[2]*u
 
   function f_oop_linear(u::Tracker.TrackedArray,p,t)
-    dx = p[1]*u[1]
-    Tracker.collect([dx])
+    p[1].*u
```
ChrisRackauckas, line 40: @frankschae this decreases it a ton and now the SO goes away. Not really the best long term solution, but it's at least an improvement we should do.

ChrisRackauckas:
Why does the first `@info` in the SDE tests give a different value? @frankschae

frankschae:
The first one is a comparison to the ODE case, i.e., the parameter in front of the diffusion term is set to zero -- thought this might be useful to check if something happens just because of the SDE part.

test/local_sensitivity/sde.jl
frankschae: Oh, I didn't know that this works. I always used Tracker with the collect so far.

ChrisRackauckas:
Interesting okay, we should probably note that somewhere.

PR #346, opened 2020-09-30
Fix CI calculation for second order Sobol indices

Vaibhavdixit02:
Looks like it broke in #338


test/global_sensitivity/sobol_method.jl
```diff
@@ -46,6 +46,8 @@ res2 = gsa(linear_batch,Sobol(),A,B,batch=true)
 @test res1.ST ≈ [0.9997953478183109, 0.0002040399766938839, 0.0, 0.0] atol=1e-4
 @test res2.ST ≈ [0.9997953478183109, 0.0002040399766938839, 0.0, 0.0] atol=1e-4
 
+@test_nowarn gsa(linear,Sobol(order=[0,1,2], nboot = 100),A,B)
```
ChrisRackauckas, line 49: is the ci not tested?

Vaibhavdixit02: Wasn't here. I don't recall why, it came up in Pumas where I had a test for them. I think the issue with getting second order CI in R was that it messed up the first order and total order values. I'll check at some point but since we needed a release for Pumas (where it will get tested) I have added the `test_nowarn` so it gets run.

Issue #4, opened 2017-09-21
Integration with callbacks?


ikirill:
In Sundials.jl there's a way to get parameter sensitivities for ODEs with discontinuities, but when I tried the same approach with ODELocalSensitivityProblem, there either doesn't seem to be a way to do it, or it's undocumented and I didn't find it. The obvious thing didn't work:

```
using DifferentialEquations

f = @ode_def FFF begin
    dx = -a
    dy = b
end a=>2.0 b=>1.0

cb = ContinuousCallback((t,u,i) -> u[1], (args...)->(println("Stopped.");f.b=0.0))
prob = ODELocalSensitivityProblem(f, [1.0, 0.0], (0.0, 1.0), callback=cb)
sol = solve(prob, DP8())
display(sol)
```

The solution is
```
 [-1.0, 0.5, -1.0, 0.0, 0.0, 1.0]                            
```
and the derivative `dy/da` is reported as zero, but should be -0.25. This is an issue whenever the callback changes some state (here b) and the trigger function depends on the state.

ChrisRackauckas:
Hmm, I didn't think about this. How exactly is the discontinuity supposed to be handled? It should apply the change to the derivatives, but how exactly?

felixlen:
Sensitivities have to be updated at every discontinuity, see for example equations 4.48a and 4.48b in [this thesis](https://mathopt.de/PUBLICATIONS/Kirches2006.pdf).

ChrisRackauckas:
Updated version:

```julia
using DiffEqSensitivity, ParameterizedFunctions
using OrdinaryDiffEq, Calculus

f = @ode_def AAA begin
    dx = -a
    dy = b
end a b

cb = ContinuousCallback((u,t,i) -> u[1], (integrator)->(println("Stopped.");integrator.p[2]=0.0))
p = [2.0, 1.0]
prob = ODELocalSensitivityProblem(f, [1.0, 0.0], (0.0, 1.0), p)
sol = solve(prob, Tsit5(), callback=cb)

DiffEqBase.has_syms(::DiffEqSensitivity.ODELocalSensitvityFunction) = false
using Plots; gr(); plot(sol,vars=[1,2])

ls = extract_local_sensitivities(sol, length(sol))[2]
ls

#> ls
# 2-element Array{Array{Float64,1},1}:
# [-1.0, 0.0]
# [0.0, 1.0]

function test_f(p)
[...]

Issue #197, opened 2020-02-17
Neural ODEs with callbacks cannot be trained with most sensitivity algorithms


sdobber:
Training Neural ODEs with callbacks currently fails for most of the sensitivity algorithms available to `concrete_solve`: Only `ForwardDiffSensitivity` works for the following MWE. However, this method scales badly with the number of parameters. Other sensitivity methods either do not update the parameters, or throw errors. (This issue is based on [this discussion on Slack](https://julialang.slack.com/archives/C7LFJTXV5/p1581812397477100?thread_ts=1581454514.389600&cid=C7LFJTXV5).)

MWE:
```
using DiffEqSensitivity, DifferentialEquations, DiffEqFlux, Flux, Tracker
using Plots

sensealg = ForwardDiffSensitivity()

datalength = 100
tspan = (0.0,100.0)
t = range(tspan[1],tspan[2],length=datalength)
target = 3.0*(1:datalength)./datalength  # some dummy data to fit to
cbinput = rand(1, datalength) #some external ODE contribution

pmodel = Chain(
    Dense(2, 10, initW=zeros),
    Dense(10, 2, initW=zeros))
p, re = Flux.destructure(pmodel)
dudt(u,p,t) = re(p)(u)

# callback changes the first component of the solution every time
# t is an integer
function affect!(integrator, cbinput)
    event_index = round(Int,integrator.t)
    integrator.u[1] += 0.2*cbinput[event_index]
end
callback = PresetTimeCallback(collect(1:datalength),(int)->affect!(int, cbinput))

# ODE with Callback
prob = ODEProblem(dudt,[0.0, 1.0],tspan, p, saveat=2, callback = callback)

function predict_n_ode(p)
    arr = Array(concrete_solve(prob, Tsit5(), [0.0, 1.0],
            p, sensealg=sensealg ))[1,2:2:end]
    return arr[1:datalength]
end

function loss_n_ode()
    pred = predict_n_ode(p)
    loss = sum(abs2,target .- pred)./datalength
end

cb = function () #callback function to observe training
  pred = predict_n_ode(p)
  display(loss_n_ode())
  pl = plot(1:datalength,target,label="data")
  plot!(pl,t,pred,label="prediction")
  display(pl)
end

Flux.train!(loss_n_ode, Flux.params(p), Iterators.repeated((), 10), ADAM(0.01), cb = cb)
```

Changing `sensealg` to other algorithms breaks the training.

[...]

PR #350, opened 2020-10-14
Handle callbacks in adjoint methods

ChrisRackauckas:
This is an implementation of https://arxiv.org/pdf/1905.10403.pdf to handle callback methods in the adjoint methods. There's a lot of tricky details as to why it works out well. First is safety. This implementation is safe for two reasons. One, the way we are generating the wrapped callbacks via broadcast reverses them in the adjoint pass. Since if discrete callbacks have events at the same time are evaluated in order, this will do the reverse order in the reverse direction and thus handle that appropriately. 

Next, we use `FakeIntegrator` in the internal `affect!` call of the wrapped callback that only supports `(u,p,t)`, so if the user tries to modify anything that's not `(u,p,t)` they will simply get an error. While that seems restrictive, reverse callback points only occur if the change happens with `u_modified!`: if `u` is not modified in the forward pass then it won't be triggered in the reverse pass. This means that things like AutoAbstol will work simply because they only modify the absolute tolerance, but not the system, so as a solver level detail it's just untouched. This is something we could change the behavior of, and let callback_adj handle in more detail, but this should make things fairly robust for most users.

The paper was missing a detail on what happens if a user changes `p`. You can easily see from the adjoint equation though that you simply have to update the integration variable, so this can be handled in the BacksolveAdjoint and the InterpolatingAdjoint, though in QuadratureAdjoint it would need to error on callbacks or have an accumulator built up the boundary terms and then append this to the integral. 

Right now the affect! function is hardcoded to be for BacksolveAdjoint. I think we might even go all the way to merging with that, since I think we will want to do a larger refactor to make a function `λ,grad,y = decompose(u,S)` in order to better handle the decomposition uniformly, and then just use `decompose` in the `affect!`. But that's a tedious refactoring that would muddy this PR so I'll hold off on that.

This is missing test cases, hence the WIP, but I know there are a lot of people who want to see this and help out so I thought placing a skeleton for people to help hit the remaining issues on would accelerate this to being completed. 

Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/4 . We should also check https://github.com/SciML/DiffEqSensitivity.jl/issues/197


src/local_sensitivity/concrete_solve.jl
```diff
@@ -34,27 +34,33 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
...
                                  save_idxs = nothing,
                                  kwargs...)
-  _prob = remake(prob,u0=u0,p=p)
 
-  # Force `save_start` and `save_end` in the forward pass This forces the
-  # solver to do the backsolve all the way back to `u0` Since the start aliases
-  # `_prob.u0`, this doesn't actually use more memory But it cleans up the
-  # implementation and makes `save_start` and `save_end` arg safe.
+  if haskey(kwargs, :callback)
+    cb = track_callbacks(CallbackSet(prob.cb,kwargs[:callback]),t)
```
timdkim, line 39: I get `type ODEProblem has no field cb` error.

src/local_sensitivity/callback_tracking.jl
```diff
@@ -0,0 +1,123 @@
...
+
+        du = get_tmp_cache(integrator)
+        dλ    = @view du[1:idx]
+        dgrad = @view du[idx+1:end-idx]
+        dy    = @view du[end-idx+1:end]
+
+        vecjacobian!(dλ, y, λ, p, t, S::SensitivityFunction;
+                              dgrad=dgrad, dy=dy)
+        integrator.u .+= du
+        _p != integrator.p && integrator.p = _p
```
timdkim, line 117: This line also threw an error for me: changing to (_p != integrator.p) && (integrator.p = _p) seemed to work.

src/local_sensitivity/callback_tracking.jl
```diff
@@ -0,0 +1,123 @@
...
+"""
+Sets up callbacks for the adjoint pass. This is a version that has an effect
+at each event of the forward pass and defines the reverse pass values via the
+vjps as described in https://arxiv.org/pdf/1905.10403.pdf Equation 13.
+
+For more information, see https://github.com/SciML/DiffEqSensitivity.jl/issues/4
+"""
+setup_reverse_callbacks(cb,sensealg) = setup_reverse_callbacks(CallbackSet(cb),sensealg)
+function setup_reverse_callbacks(cb::CallbackSet,sensealg)
+    cb = CallbackSet(_setup_reverse_callbacks.(cb.continuous_callbacks,sensealg)
```
timdkim, line 79: Missing comma at the end of this line.

timdkim (commented):
I tried running a few lines of code with the `callbacks` branch of DiffEqSensitivity.jl and left comments below.

ChrisRackauckas:
Alright, the infrastructure is all there. Now the only issue is that it seems to... compute the wrong value. But there's a test and the whole algorithm is seemingly implemented, so it's just bug hunting now. It would probably be good if @YingboMa combed through it to see if he can see anything. The meat of the algorithm is really just:

https://github.com/SciML/DiffEqSensitivity.jl/pull/350/files#diff-37fbf6ccf4a89df4f73a77046547985f3f2221fa8eaaa2b017e822160c50ff04R85

At that point the algorithm already knows where it should do the event, and it's all set up to do a vjp through the event handling. But it's getting something wrong, and my guess is that it's not exactly doing Equation 13 from https://arxiv.org/pdf/1905.10403.pdf

test/local_sensitivity/callbacks.jl
```diff
@@ -1,5 +1,178 @@
...
+
+@test du02 ≈ dstuff[1:2]
+@test dp2 ≈ dstuff[3:6]
+
+
+# callback with no effect except saving the state
+
+condition(u,t,integrator) = t == 5
+affect!(integrator) = integrator.u[1] += 0.0
+cb = DiscreteCallback(condition,affect!)
```
frankschae, line 63: just found that this test is already broken at the moment, i.e., `save_positions=(true,true)` instead of `save_positions=(false,false)`.

src/local_sensitivity/sensitivity_interface.jl
```diff
@@ -27,7 +27,7 @@ function _adjoint_sensitivities(sol,sensealg,alg,g,t=nothing,dg=nothing;
   tstops = ischeckpointing(sensealg, sol) ? checkpoints : similar(sol.t, 0)
   adj_sol = solve(adj_prob,alg;
                   save_everystep=false,save_start=false,saveat=eltype(sol[1])[],
-                  abstol=abstol,reltol=reltol,kwargs...)
+                  tstops=tstops,abstol=abstol,reltol=reltol,kwargs...)
```
frankschae: The `tstops` here seems to be necessary to make checkpointing work as expected. Otherwise, these tests:
https://github.com/SciML/DiffEqSensitivity.jl/blob/e0ceda43ea1184ec0a6c3eab202f7f5099401854/test/local_sensitivity/adjoint.jl#L610
are failing.

src/local_sensitivity/concrete_solve.jl
```diff
@@ -154,10 +155,10 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
       end
     end
 
-    if haskey(kwargs, :callback_adj)
+    if haskey(kwargs_adj, :callback_adj)
       cb2 = CallbackSet(cb,kwargs[:callback_adj])
     else
-      cb2 = nothing
+      cb2 = cb
```
frankschae, line 161: This enabled the passing tests in the previous test runs.. The reverse callback was actually not triggered.

test/local_sensitivity/callbacks.jl
```diff
@@ -117,8 +165,31 @@ dstuff = ForwardDiff.gradient(
...
+						 abstol=1e-14,reltol=1e-14)
+adj_sol1 = solve(adj_prob, Tsit5(), tstops=sol_track.t, abstol=1e-14,reltol=1e-14)
+
+
+@test du01 ≈ -adj_sol1[1:2,end]
+@test dp1 ≈ adj_sol1[3:6,end]
+
+
+adj_prob = ODEAdjointProblem(sol_track,BacksolveAdjoint(),dg!,sol_track.t,nothing,
+						 callback = nothing,
```
frankschae, line 183: This is at the moment the odd thing.. When the reverse callback is not used, the gradients are correctly computed (because of the checkpointing..)

test/local_sensitivity/callbacks.jl
```diff
@@ -56,7 +58,24 @@ dstuff = ForwardDiff.gradient(
...
+
+adj_prob = ODEAdjointProblem(sol_track,BacksolveAdjoint(),dg!,sol_track.t,nothing,
+						 callback = cb2,
+						 abstol=1e-14,reltol=1e-14)
+adj_sol = solve(adj_prob, Tsit5(), abstol=1e-14,reltol=1e-14)
+@test adj_sol(sol1.t)[end-1:end,:] ≈ sol1[:,:]
+@test du01 ≈ -adj_sol[1:2,end]
+@test dp1 ≈ adj_sol[3:6,end]
+
+### callback with no effect except saving the state
```
frankschae, line 78: Making this one work is also a bit tricky.. Probably because of the non-uniqueness of the checkpoints?

ChrisRackauckas: oh interesting. Maybe that's it. I didn't look too deep into why it failed.

test/local_sensitivity/callbacks.jl
ChrisRackauckas: oh... hahahaha what?

test/local_sensitivity/callbacks.jl
ChrisRackauckas: Should we internally change save_positions or something? That might give odd behavior.

test/local_sensitivity/callbacks.jl
frankschae: This one fails because  `t` is not unique in
https://github.com/SciML/DiffEqSensitivity.jl/blob/beb0366d1992758f131cc7c95210fd20371f672a/src/adjoint_common.jl#L277

Therefore, the callbacks in the adjoint solution (for `dg`) stop to be triggered after the reverse callback is applied. (See the difference in the two plots. First one is "no effect except saving the state"; second one is "no effect")
with_save
without_save


So a possible fix could be to use
```_t = unique(t)
cur_time = Ref(length(_t))```
and to add another callback for the duplicates.

**Comment by ChrisRackauckas** [comment]
Hmm, what happened to the CI actions? Did an old master get merged in?

**Comment by frankschae** [comment]
I am not sure... I thought that might be resolved once I rebase this PR onto the current master and solve the merge conflicts. 
The issue with the saving option is now solved with the last refactorization. However, the tests with a jump applied still fail.

**Comment by ChrisRackauckas** [comment]
Callbacks should be working now?

**Comment by frankschae** [comment]
Tests are working for all examples, except:
```
condition(u,t,integrator) = t == 5
affect!(integrator) = (integrator.u[1] = 2.0; @show "triggered!")
cb = DiscreteCallback(condition,affect!)
```
with the small handicap that the checkpointing is not yet fixed analoguesly to the callbacks and "checkpointing=false" has to be used.

ChrisRackauckas:
Looks like you un-rebased it?

ChrisRackauckas:
Let's get this merged with what it has. Set that one failing. This is already a big jump and we can tackle the little issues around it next.

ChrisRackauckas:
This seemed to find an odd issue in PresetCallbacks which the previous handling somehow ignored. But the upstream change https://github.com/SciML/DiffEqCallbacks.jl/pull/94 should fix it. Testing locally.

ChrisRackauckas:
Local tests seem good. That seems to have done the trick.

PR #356, opened 2020-11-10
make reversediffadjoint return a sensitivity solution

ChrisRackauckas:


ChrisRackauckas:
Supersceded by https://github.com/SciML/DiffEqSensitivity.jl/pull/358

PR #360, opened 2020-11-25
fix SDEs + ReverseDiff

mohamed82008:
This PR fixes the issue in https://github.com/JuliaDiff/ReverseDiff.jl/issues/159. Let's see what CI says.

CC: @Libbum


src/local_sensitivity/concrete_solve.jl
```diff
@@ -332,7 +332,13 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,sensealg::ReverseDiffAdjoin
   typeof(p) <: DiffEqBase.NullParameters || ReverseDiff.value!(tp, p)
   ReverseDiff.forward_pass!(tape)
   function reversediff_adjoint_backpass(ybar)
-    ReverseDiff.increment_deriv!(output, ybar)
+    if prob isa SDEProblem
```
devmotion, line 335: Is this a problem of SDEs only, i.e. does it work for ODEs? The type of `sol.u`, `sol.t` etc should usually be the same, I would assume.

mohamed82008: The problematic case was an SDE. I am assuming it worked for ODEs, since tests pass and all.

Libbum:
Think this is working as expected!

src/local_sensitivity/concrete_solve.jl
devmotion: It might be good to add the SDE example as a test.

mohamed82008: Yes

test/local_sensitivity/sde_neural.jl
```diff
@@ -0,0 +1,102 @@
...
+p_ = Float32[1.1, 1.0, 0.0, 2.0, 1.0, 1.0, 1e-6, 1.0]
+
+prob = SDEProblem(sys!, noise!, u0, tspan, p_)
+ensembleprob = EnsembleProblem(prob)
+
+solution = solve(
+    ensembleprob,
+    SOSRI(),
+    EnsembleThreads();  
+    trajectories = 1000,
```
Libbum, line 39: Probably best to drop the trajectories down to like 2 for the tests. The real values aren't important here.

test/local_sensitivity/sde_neural.jl
```diff
@@ -0,0 +1,102 @@
...
+    false
+end
+
+res1 = DiffEqFlux.sciml_train(
+    loss,
+    α,
+    ADAM(0.1),
+    cb = callback,
+    maxiters = 200,
+)
```
devmotion, line 102: I assume the callback can be removed in the tests:
```suggestion
res1 = DiffEqFlux.sciml_train(
    loss,
    α,
    ADAM(0.1),
    maxiters = 200,
)
```

src/local_sensitivity/concrete_solve.jl
ChrisRackauckas: But why does this happen? I I don't get it.

Project.toml
```diff
@@ -50,6 +50,8 @@ julia = "1.5"
 [extras]
 AlgebraicMultigrid = "2169fc97-5a83-5252-b627-83903c6c433c"
 Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
+DiffEqFlux = "aae7a2af-3d4f-5e19-a356-7da93b79d9d0"
```
ChrisRackauckas, line 53: This should get a separate test section with a separate Project.toml for downstream. See https://github.com/SciML/DiffEqBase.jl/blob/master/test/runtests.jl#L38-L40 as an example.

Libbum:
Thanks for taking another stab at this Chris. When I said it was 'working' before, I meant in the sense that it didn't spit errors at the start of the training. Convergence was a different story&mdash;that was mostly unsuccessful. I'll see if these changes help out in that regard.

ChrisRackauckas:
Thanks @mohamed82008 !

@Libbum well at least these tests show it computes the correct gradients. If it's not converging, there's something else in the story that's causing that.

Issue #376, opened 2021-01-09
Support callbacks with certain kinds of SDEs: refactoring


ChrisRackauckas:
Callbacks want:

```julia
        # Hardcoding to match BacksolveAdjoint
        λ     = @view integrator.u[1:idx]
        grad  = @view integrator.u[idx+1:end-idx]
        y     = @view integrator.u[end-idx+1:end]

        du = first(get_tmp_cache(integrator))
        dλ    = @view du[1:idx]
        dgrad = @view du[idx+1:end-idx]
        dy    = @view du[end-idx+1:end]
```

what is that for InterpolatingAdjoint?

```julia
  @unpack sol, y, checkpoint_sol, discrete, prob, f = S
  idx = length(y)

  if checkpoint_sol === nothing
    if typeof(t) <: ForwardDiff.Dual && eltype(S.y) <: AbstractFloat
      y = sol(t)
    else
      sol(y,t)
    end
  else
    intervals = checkpoint_sol.intervals
    interval = intervals[checkpoint_sol.cursor]
    if !(interval[1] <= t <= interval[2])
      cursor′ = findcursor(intervals, t)
      interval = intervals[cursor′]
      cpsol_t = checkpoint_sol.cpsol.t
      if typeof(t) <: ForwardDiff.Dual && eltype(S.y) <: AbstractFloat
        y = sol(interval[1])
      else
        sol(y, interval[1])
      end
      if typeof(sol.prob) <: SDEProblem
        #idx1 = searchsortedfirst(sol.t, interval[1])
        _sol = deepcopy(sol)
        _sol.W.save_everystep = false
        idx1 = searchsortedfirst(_sol.t, interval[1]-100eps(interval[1]))
        idx2 = searchsortedfirst(_sol.t, interval[2])
        #forwardnoise = DiffEqNoiseProcess.NoiseGrid(_sol.t[idx1:idx2], _sol.W.W[idx1:idx2])
        forwardnoise = DiffEqNoiseProcess.NoiseWrapper(_sol.W, indx=idx1)
        prob′ = remake(prob, tspan=intervals[cursor′], u0=y, noise=forwardnoise)
        dt = abs(cpsol_t[end]-cpsol_t[end-1])
        if dt < 10000eps(cpsol_t[end])
          dt = interval[2] - interval[1]
        end
[...]

Issue #375, opened 2021-01-09
Support callbacks with InterpolatingAdjoint and QuadratureAdjoint


ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/pull/350 hardcoded the callback support for BacksolveAdjoint. The other adjoint choices should be handled soon.

PR #377, opened 2021-01-09
refactor state splitting and allow interpolating and quadrature cbs

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/376

Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/375

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -327,11 +327,10 @@ end
 function generate_callbacks(sensefun, g, λ, t, callback, init_cb)
 
   reverse_cbs = setup_reverse_callbacks(callback,sensefun.sensealg)
-  sensefun.discrete || return reverse_cbs
+  sensefun.discrete || return reverse_cbs, nothing
 
   # callbacks can lead to non-unique time points
   _t, duplicate_iterator_times = separate_nonunique(t)
-
   rlcb = ReverseLossCallback(sensefun, λ, _t, g)
 
   cb = PresetTimeCallback(_t,rlcb)
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -340,9 +340,9 @@ function generate_callbacks(sensefun, g, λ, t, callback, init_cb)
   if duplicate_iterator_times!==nothing
     cbrev_dupl_affect = ReverseLossCallback(sensefun, λ, duplicate_iterator_times[1], g)
     cb_dupl = PresetTimeCallback(duplicate_iterator_times[1],cbrev_dupl_affect)
-    return CallbackSet(cb,reverse_cbs,cb_dupl)
+    return CallbackSet(cb,reverse_cbs,cb_dupl), duplicate_iterator_times
   else
-    return CallbackSet(cb,reverse_cbs)
+    return CallbackSet(cb,reverse_cbs), duplicate_iterator_times
   end
 end
 
--- a/src/backsolve_adjoint.jl
+++ b/src/backsolve_adjoint.jl
@@ -18,7 +18,32 @@ end
 
 # u = λ'
 function (S::ODEBacksolveSensitivityFunction)(du,u,p,t)
-  @unpack y, prob, f, discrete = S
+  @unpack y, prob, discrete = S
+
+  λ,grad,_y,dλ,dgrad,dy = split_states(du,u,t,S)
+  copyto!(vec(y), _y)
+
+  if S.noiseterm
+    if length(u) == length(du)
+      vecjacobian!(dλ, y, λ, p, t, S, dgrad=dgrad,dy=dy)
+    elseif length(u) != length(du) &&  StochasticDiffEq.is_diagonal_noise(prob) && !isnoisemixing(S.sensealg)
+      vecjacobian!(dλ, y, λ, p, t, S, dy=dy)
+      jacNoise!(λ, y, p, t, S, dgrad=dgrad)
+    else
+      jacNoise!(λ, y, p, t, S, dgrad=dgrad, dλ=dλ, dy=dy)
+    end
+  else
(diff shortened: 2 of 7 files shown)


frankschae:
Replacing:
```julia
      if _prob.tspan[2] > _prob.tspan[1]
        ts = _prob.tspan[1]:convert(typeof(_prob.tspan[2]),abs(saveat)):_prob.tspan[2]
      else
        ts = _prob.tspan[2]:convert(typeof(_prob.tspan[2]),abs(saveat)):_prob.tspan[1]
      end
```
by:
```julia
if cb === nothing
      if _prob.tspan[2] > _prob.tspan[1]
        ts = _prob.tspan[1]:convert(typeof(_prob.tspan[2]),abs(saveat)):_prob.tspan[2]
      else
        ts = _prob.tspan[2]:convert(typeof(_prob.tspan[2]),abs(saveat)):_prob.tspan[1]
      end
else
      ts = sol.t
end
```
in https://github.com/SciML/DiffEqSensitivity.jl/blob/cb2e252a4e45bcaa3a238c9720f7a0fecb7a513e/src/concrete_solve.jl#L82 would fix the first failing test  (callback does nothing but saving one additional step at 0.5)  for `InterpolatingAdjoint` if `checkpointing=true` is set.

This is because at the moment we obtain `ts = 0.0:0.1:10.0` while we actually would like to have `[0.0, ... 0.4, 0.5, 0.5, 0.6, ..10.0]`. The first failing test therefore at the moment returns gradient values which are equal to the gradient values in the case of the callback without the step saving option enabled.

src/backsolve_adjoint.jl
```diff
@@ -215,9 +215,14 @@ end
 
 
 
-function backsolve_checkpoint_callbacks(sensefun, sol, checkpoints, callback)
+function backsolve_checkpoint_callbacks(sensefun, sol, checkpoints, callback, duplicate_iterator_times=nothing)
   prob = sol.prob
-  cur_time = Ref(length(checkpoints))
+  if duplicate_iterator_times !== nothing
+    _checkpoints = filter(x->x ∉ duplicate_iterator_times[1], checkpoints)
```
frankschae, line 221: The callbacks in Backsolve didn't work because ultimately the wrong state u entered into the reverse callback.  This filters the duplicates fully out of the checkpointing. If we want to keep a checkpoint at the affect times we should somehow make sure that the checkpoint always refers to the proper limit.

ChrisRackauckas: I think removing the event times is a good idea.

src/concrete_solve.jl
```diff
@@ -61,10 +61,10 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
 
   # Capture the callback_adj for the reverse pass and remove both callbacks
   kwargs_adj = NamedTuple{Base.diff_names(Base._nt_names(values(kwargs)), (:callback_adj,:callback))}(values(kwargs))
-
+  isq = sensealg isa QuadratureAdjoint
   if typeof(sensealg) <: BacksolveAdjoint
     sol = solve(_prob,alg,args...;save_noise=true,save_start=save_start,save_end=save_end,saveat=saveat,kwargs_fwd...)
-  elseif ischeckpointing(sensealg)
+  elseif ischeckpointing(sensealg) || (cb!==nothing && !isq)
```
frankschae, line 67: This was needed to make the first two InterpolatingAdjoint tests work. Without this, sol.ts won't be given by saveat + affect times (but by the adaptively chosen time steps).

frankschae:
All tests for the `QuadratureAdjoint` should now pass. I marked the `InterpolatingAdjoint` tests as broken to see if  indeed  all quadrature tests pass.  I think for `InterpolatingAdjoint`  the remaining thing is to handle the time spans with (t*, t*) correctly.

src/adjoint_common.jl
```diff
@@ -327,11 +327,10 @@ end
 function generate_callbacks(sensefun, g, λ, t, callback, init_cb)
 
   reverse_cbs = setup_reverse_callbacks(callback,sensefun.sensealg)
-  sensefun.discrete || return reverse_cbs
+  sensefun.discrete || return reverse_cbs, nothing
```
ChrisRackauckas, line 330: oh...

frankschae: This was added because of the error from the earlier CI run https://github.com/SciML/DiffEqSensitivity.jl/runs/1696580565

ChrisRackauckas: I'm saying, yeah oops, I should've noticed this one immediately haha.

ChrisRackauckas:
🎉 got one.

src/concrete_solve.jl
```diff
@@ -80,18 +80,14 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
...
-      end
+    if _prob.tspan[2] > _prob.tspan[1]
+      ts = _prob.tspan[1]:convert(typeof(_prob.tspan[2]),abs(saveat)):_prob.tspan[2]
     else
-        ts = sol.t
+      ts = _prob.tspan[2]:convert(typeof(_prob.tspan[2]),abs(saveat)):_prob.tspan[1]
+    end
+    if cb !== nothing
+      _, duplicate_iterator_times = separate_nonunique(sol.t)
+      duplicate_iterator_times!==nothing && (ts = sort(vcat(ts, vcat(duplicate_iterator_times...))))
```
frankschae, line 90: These changes make the InterpolatingAdjoint + callback method closer to the InterpolatingAdjoint without callback method, since now the checkpoints are computed from the adaptive solution in both cases. The rest basically just assigns proper tstops in all solve calls.

Issue #374, opened 2021-01-09
Support state-dependent continuous callbacks with adjoints


ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/pull/350 did the discrete callbacks. It shouldn't be much more to similarly wrap the continuous callbacks.

ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/issues/197 should get tested and closed when this is completed.

ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/pull/381 found that it's all good except when `u`-dependent, so it requires an extra term.

ChrisRackauckas:
@frankschae was this handled?

frankschae:
No, it's not done yet. Michael and Stefano pointed me to this paper: https://arxiv.org/abs/2011.03902 but I think the derivation lacks the details about which limit one has to take (I still try to figure that out but generally it looks like that will potentially solve the issue). Maybe @YingboMa could also have a look what term from the chain rule we should add to our code.

PR #381, opened 2021-01-21
Add ContinuousCallback tests 

frankschae:
Handling a `ContinuousCallback`  in the adjoints (#374) seems to work already equally well as in case of `DiscreteCallback`s.

For example, defining a `ContinuousCallback` with stopping time `tstop`:
```julia
function condition(u,t,integrator) 
   t - tstop
end
function affect!(integrator)
  integrator.u[2] += vafter-vbefore
end
cb = ContinuousCallback(condition,affect!
```
works properly.

Based on the test set, there are however also a few things which don't work yet.
1. When the condition is state dependent, the `ContinuousCallbacks` are not yet working
```julia
function condition(u,t,integrator) # Event when event_f(u,t) == 0
   u[1]
end
```
2.  The affect function must currently be written with a `+=`. This also holds for `DiscreteCallback`s. So, so tests with, e.g.,
```julia
function affect!(integrator)
  integrator.u[2] = -integrator.p[2]*integrator.u[2]
end
```
or
```julia
function affect!(integrator)
  integrator.u[2] = 2.0
end
```
are broken.
3. If the position is saved, i.e., `save_positions=(true,true)`. Then, gradients are dropped when slices on the solution are used
```julia
du01,dp1 = Zygote.gradient(
  (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,
   callback=cb,
   sensealg=BacksolveAdjoint())[end]),u0,p)
```
This also holds for both types of callbacks.

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -362,7 +362,7 @@ function separate_nonunique(t)
     itrs = nothing
   else
     maxoc = maximum(occurances)
-    maxoc > 2 && warning("More than two occurances of the same time point. Please report this.")
+    maxoc > 2 && error("More than two occurances of the same time point. Please report this.")
     # handle also more than two occurances
     itrs = [ts[occurances .>= i] for i=2:maxoc]
   end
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Oh, wild. Let's get an issue open to start working through these last 3 issues. (1) sounds like a missing term in the update. (2) is something tricky I don't understand right now. (3) seems like another one of those off-by-one callback time things.

PR #386, opened 2021-01-24
Fix ReverseDiffAdjoint

ChrisRackauckas:
https://www.reddit.com/r/Julia/comments/l3243t/odd_behavior_neural_network_hybrid_differential/ found that this change needed to be rolled back. While it doesn't exactly conform to the interface, it seems ReverseDiff just cannot support the sol object directly, so this needs to be done for correctness. Because our other callback adjoint tests didn't seem to catch this, I added a downstream test on that case specifically to catch gradient issues here.

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -391,7 +391,6 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,sensealg::ReverseDiffAdjoin
     else
       u = map(ReverseDiff.value,sol.u)
     end
-
     Array(sol)
   end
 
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
@mohamed82008 this brings a regression in another spot. Is there a way to support an array like this well?

mohamed82008:
A correctness or performance regression? I think your current fix makes sense. If you describe the problem in more details, I may be able to help.

ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/pull/386/checks?check_run_id=1757395253#step:6:404

ChrisRackauckas:
I got it, the pullback vector was a vector of vectors and needed to be converted.

But the `Array(sol)` always returns an array of TrackedReal, even if `u0` was a TrackedArray (and `sol[i]` are TrackedArray`). Even `reduce(hcat,sol.u)` does that, so I just went with the generic solution.

PR #396, opened 2021-03-13
better forward mode tags

ChrisRackauckas:


ChrisRackauckas:
@mohamed82008 is this a recent change? https://github.com/SciML/DiffEqSensitivity.jl/pull/396/checks?check_run_id=2102525656#step:6:210

mohamed82008:
yes

mohamed82008:
it was to fix https://github.com/JuliaDiff/ReverseDiff.jl/issues/76

ChrisRackauckas:
That change actually revealed a bug which has now been fixed in the Ito handling @frankschae

Issue #398, opened 2021-03-15
Type assert errors when using discrete sensitivity methods with sparse Jacobian


bgroenks96:
Specifying sparse Jacobians appears to be incompatible with discrete sensitivity methods. As mentioned in #393, if one specifies the Jacobian as a `Tridiagonal` (without loss of generality) with type `Float64` it will fail when differentiating because everything is of type `ForwardDiff.Dual`. However, if one defines it with type `ForwardDiff.Dual` it will fail on the forward pass of sensitivity analysis (or with numerical diff) when things are type `Float64`. Thus, there needs to be a way to make `DiffEqSensitivity` automatically infer/convert the Jacobian type agnostic to whether or not autodiff is being used.

Simple MWE with the heat equation:

```julia
using DifferentialEquations, DiffEqSensitivity, DiffEqFlux, Flux
using ComponentArrays, LinearAlgebra
using Plots

const nknots = 10
const h = 1.0/(nknots+1)
x = range(0, step=h, length=nknots)
u0 = sin.(π*x)

@inline function f(du,u,p,t)
  du .= zero(eltype(u))
  u₃ = @view u[3:end]
  u₂ = @view u[2:end-1]
  u₁ = @view u[1:end-2]
  @. du[2:end-1] = p.k*((u₃ - 2*u₂ + u₁)/(h^2.0))
  nothing
end

p_true = ComponentArray(k=0.42)
jac_proto = Tridiagonal(similar(u0,nknots-1), similar(u0), similar(u0, nknots-1))
prob = ODEProblem(ODEFunction(f,jac_prototype=jac_proto), u0, (0.0,1.0), p_true)
@time sol_true = solve(prob, Rodas4P(), saveat=0.1)
plot(sol_true, legend=nothing)

function loss(prob0, p)
  prob = remake(prob0, p=p)
  sol = solve(prob, Rodas4P(autodiff=false), saveat=0.1, sensealg=ForwardDiffSensitivity())
  sum((sol .- sol_true).^2)
end

p0 = ComponentArray(k=1.0)
@time res = DiffEqFlux.sciml_train(p -> loss(prob,p), p0, ADAM(0.01), maxiters=100)
```

`TypeError: in typeassert, expected Float64, got a value of type ForwardDiff.Dual{Nothing,Float64,1}`

bgroenks96:
[...]

PR #399, opened 2021-03-23
Handle sparsity in ForwardDiffSensitivity + test ComponentArrays

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/398 . I like this test: no holds barred, it really digs deep.


bgroenks96:
Happy to accidentally contribute a test case :)

ChrisRackauckas:
The test is only failing because ComponentArrays.jl needs to fix a few compats: https://github.com/jonniedie/ComponentArrays.jl/pull/75

PR #408, opened 2021-04-22
Fix array save_idxs

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqFlux.jl/issues/534


ChrisRackauckas:
Fixing ReverseDiff.jl to v1.7 for now because it causes a test failure:

https://github.com/SciML/DiffEqSensitivity.jl/pull/408/checks?check_run_id=2407872706#step:6:304

PR #416, opened 2021-05-28
DiscreteCallback fixes

frankschae:
This PR fixes several issues related to continuous adjoints and (discrete) callbacks.

- The computation of the vjp for the event needs the Jacobian with respect to the state *before* the event. Therefore, I added another field `uleft` to `TrackedAffect`.
- There shouldn't be a `+=` as written in Eq.(13) in https://arxiv.org/abs/1905.10403. The operation must be: `λ .= dλ` (I'll need to add the update for callbacks that change the parameter as well in a follow up.) 
- In `InterpolatingAdjoint`, we need to take the right limit (`continuity=:right`) when computing `y`. Otherwise, we "jump" over the discontinuity and compute a false loss function value.
- Similarly the  `out_and_ts()` makes sure that we compute the loss function with respect to the correct `u` values (regarding e.g. the left and right limit and the order of the time values).  Another problem with the discrete callbacks was that it only worked if the callback time is contained in the saving times. 
- Before, we had two `cur_time[]` Ref values. This is also problematic for selecting the appropriate u values within the direct `_concrete_solve_adjoint` interface. We are now back at a single Ref shared between the loss callbacks for the "standard" saving times and the "callback" times, respectively. The difference between the two `ReverseLossCallback` is given by the times at which the `PresetTimeCallback` ultimately is triggered.  
- I extended the test suite quite a bit.

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -316,9 +316,7 @@ struct ReverseLossCallback{λType,timeType,yType,RefType,FMType,AlgType,gType,ca
   diffcache::cacheType
 end
 
-function ReverseLossCallback(sensefun, λ, t, g)
-  cur_time = Ref(length(t))
-
+function ReverseLossCallback(sensefun, λ, t, g, cur_time)
   @unpack sensealg, y = sensefun
   isq = (sensealg isa QuadratureAdjoint)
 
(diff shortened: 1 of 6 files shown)


src/adjoint_common.jl
```diff
@@ -403,3 +409,31 @@ function separate_nonunique(t)
...
+    # value for them. This information is given by sol.t/checkpoints.
+    # Additionally we need to store the left and the right limit, respectively.
+    duplicate_times = duplicate_iterator_times[1] # just treat two occurances at the moment (see separate_nonunique above) 
+    _ts = Array(_ts)
+    for d in duplicate_times
+      (d ∉ _ts) && push!(_ts, d)
+    end
+
+    u1 = sol(_ts)[:]
+    u2 = sol(duplicate_times,continuity=:right)[:]
```
ChrisRackauckas, line 431: if ts is right to left, right, else left?

frankschae: ts always left, duplicate_times always right, but duplicate_times left and right if not contained in ts.

PR #417, opened 2021-06-04
allow for changes of `p` in callbacks

frankschae:
To handle changes in `p`, one needs to add an additional vjp. That consists of `grad` (gradient of the loss function with respect to the parameter) as the vector and the Jacobian of the callback function with respect to the parameters before the callback. Slightly abusing the `vecjacobian!` interface, we may use:
```julia
function wp(dp,p,u,t)
    fakeinteg = FakeIntegrator([x for x in u],[x for x in p],t)
    cb.affect!.affect!(fakeinteg)
    dp .= fakeinteg.p
end

fakeSp = CallbackSensitivityFunction(wp,sensealg,S.diffcache,integrator.sol.prob)
#vjp with Jacobin given by dw/dp before event and vector given by grad
vecjacobian!(dgrad, integrator.p, grad, y, integrator.t, fakeSp;
                                  dgrad=nothing, dy=nothing)
grad .= dgrad       
```

For `QuadratureAdjoint`, we need to accumulate/update the grads during the integration for `p` when duplicate times are reached. (There is still an error in the current version of the code as `dp` doesn't come out correctly..)
For `InterpolatingAdjoint` with  checkpointing,  one needs to do a check if a callback might have occured in a given time interval. If that is the case we need to search for the left-most callback time and use the associated stored left-limit of the callback for the checkpointed solution. 

I have not yet figured out how `ReverseDiffAdjoint` could be made compatible with 
```julia
affect!(integrator) = (integrator.p .= 2*integrator.p .- 0.5)
```
.. it fails with "TrackedArrays do not support `setindex`". It works with 
```julia
affect!(integrator) = (integrator.p = 2*integrator.p .- 0.5) # no inplace
```
though. For `integrator.u` it works both ways. Could someone point me to the trick?

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -4,42 +4,45 @@ the reverse pass. The rationale is explain in:
 
 https://github.com/SciML/DiffEqSensitivity.jl/issues/4
 """
-track_callbacks(cb,t,u) = track_callbacks(CallbackSet(cb),t,u)
-track_callbacks(cb::CallbackSet,t,u) = CallbackSet(
-   map(cb->_track_callback(cb,t,u), cb.continuous_callbacks),
-   map(cb->_track_callback(cb,t,u), cb.discrete_callbacks))
+track_callbacks(cb,t,u) = track_callbacks(CallbackSet(cb),t,u,p)
+track_callbacks(cb::CallbackSet,t,u,p) = CallbackSet(
+   map(cb->_track_callback(cb,t,u,p), cb.continuous_callbacks),
+   map(cb->_track_callback(cb,t,u,p), cb.discrete_callbacks))
 
-struct TrackedAffect{T,T2,T3}
+struct TrackedAffect{T,T2,T3,T4}
     event_times::Vector{T}
     uleft::Vector{T2}
-    affect!::T3
+    pleft::Vector{T3}
+    affect!::T4
 end
 
-TrackedAffect(t::Number,u,affect!::Nothing) = nothing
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
>.. it fails with "TrackedArrays do not support setindex". It works with

You'd have to transform `p` from a TrackedArray to an `Array{TrackedReal}`, which is done via `convert.(eltype(p),p)` before solving with it. This of course has performance implications since then it would be tracking each scalar interaction instead of tracking the array as a whole, which would increase the tape size if tons of array operations are used. If it's common to do scalar getindex on the `p` though, then the compute graph is already scalarizing and so it would be a performance hit. So... it somehow needs to switch in a smart way. For inplace vs out of place, it just only switches when inplace since it would just fail otherwise. Here, 🤷 open an issue. Maybe it can be an option in `ReverseDiffAdjoint` to flip. I think with one of the common use cases being neural networks, scalarizing `p` by default would be a bad idea.

ChrisRackauckas:
Mark as `@test_broken` for now? Or implement the option.

frankschae:
> For QuadratureAdjoint, we need to accumulate/update the grads during the integration for p when duplicate times are reached. (There is still an error in the current version of the code as dp doesn't come out correctly..)

The integration between the intervals of successive `ts` had to be reversed:
```julia
for i in 1:length(t)-1
```
```julia
for i in length(t)-1:-1:1
```
to compute the vjp with respect to the accumulated gradient of larger times.

PR #419, opened 2021-06-07
fix broadcasting error on steady state adjoint

frankschae:
Fixes the broadcasting error on CPU: https://github.com/SciML/DiffEqFlux.jl/issues/567#issuecomment-855819416

I haven't tested yet if that also fixes the GPU error. @QiyaoWei

### DIFF
--- a/src/steadystate_adjoint.jl
+++ b/src/steadystate_adjoint.jl
@@ -62,7 +62,7 @@ end
     end
   end
 
-  copyto!(λ[:], diffcache.J'\vec(diffcache.dg_val')) # use linsolve here
+  copyto!(vec(λ), diffcache.J'\vec(diffcache.dg_val')) # use linsolve here
   vecjacobian!(vec(diffcache.dg_val), y, λ, p, nothing, sense, dgrad=vjp, dy=nothing)
 
   if g != nothing
--- a/src/steadystate_adjoint.jl
+++ b/src/steadystate_adjoint.jl
@@ -62,7 +62,7 @@ end
     end
   end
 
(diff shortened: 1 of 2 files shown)


src/steadystate_adjoint.jl
```diff
@@ -62,7 +62,7 @@ end
     end
   end
 
-  λ .= diffcache.J'\vec(diffcache.dg_val') # use linsolve here
+  copyto!(λ[:], diffcache.J'\vec(diffcache.dg_val')) # use linsolve here
```
ChrisRackauckas, line 65:
```suggestion
  copyto!vec((λ), diffcache.J'\vec(diffcache.dg_val')) # use linsolve here
```

ChrisRackauckas (approved):

frankschae:
It failed in the first run at the line:
```julia
 res4c = adjoint_sensitivities(soloop,DynamicSS(Rodas5()),sensealg=SteadyStateAdjoint(autodiff=false),g,nothing)
 for p: Test Failed at /home/runner/work/DiffEqSensitivity.jl/DiffEqSensitivity.jl/test/steady_state.jl:163
  Expression: norm(res_analytical' .- res4c) < 1.0e-7
   Evaluated: NaN < 1.0e-7
```
it looks like a rare random error related to the computation of the Jacobian by using `FiniteDifferences`. I tested that part locally >10 times and got only once a NaN..

ChrisRackauckas:
>FiniteDifferences

Instead of FiniteDiff?

frankschae:
oops no sorry.  I meant FiniteDiff.

ChrisRackauckas:
Look for un-zero'd memory, i.e. what's using `similar` that never gets changed and why?

Issue #102, opened 2019-10-04
Shadow adjoints for chaotic systems


ChrisRackauckas:
These are all methods necessary for getting gradients to fit chaotic systems.

https://www.sciencedirect.com/science/article/pii/S0021999117304783
https://arxiv.org/abs/1807.05568
https://arxiv.org/abs/1902.11112

ChrisRackauckas:
https://arxiv.org/abs/2103.08816

PR #420, opened 2021-06-10
Forward Least Squares Shadowing (LSS)

frankschae:
Implements the (original) LSS method in forward mode due to Qiqi Wang et al. https://arxiv.org/abs/1204.0159.
I followed their [python repository](https://github.com/qiqi/lssode ) for different solving/windowing options.

It's quite a memory intense method. Partly, this could probably be solved by wrapping `sparse` around some of the matrices used in the shadowing problem. It's inherently a block diagonal structure, so we are sure that the matrices will always be sparse even if, e.g., the Jacobian is not sparse. On the other side there is `NILSS` which avoids the computations of the Jacobian entirely.. 

I didn't implemented the  `concrete_solve` dispatch yet. Also the `adjoint` mode needs to be added. 

Related issue: #102

### DIFF
--- a/src/lss.jl
+++ b/src/lss.jl
@@ -176,7 +176,7 @@ function ForwardLSSProblem(sol, sensealg::ForwardLSS, g, dg = nothing;
   ForwardLSSProblem{typeof(sensealg),typeof(sense),typeof(sol),typeof(dt),
     typeof(umid),typeof(dudt),
     typeof(S),typeof(b),typeof(η),typeof(w),typeof(v),typeof(window),typeof(Δt),
-    typeof(g),typeof(g0),typeof(res)}(sensealg,sense,sol,dt,umid,dudt,S,b,η,w,v,window,Δt,Nt,g,g0,
+    typeof(g0),typeof(g),typeof(res)}(sensealg,sense,sol,dt,umid,dudt,S,b,η,w,v,window,Δt,Nt,g0,g,
     res)
 end
 
@@ -297,7 +297,7 @@ function solve(prob::ForwardLSSProblem; t0skip=zero(prob.Δt), t1skip=zero(prob.
 end
 
 function _solve(prob::ForwardLSSProblem,sensealg::ForwardLSS,alpha::Number,t0skip,t1skip)
-  @unpack sol, S, window, Δt, diffcache, b, w, v, η, res, g, g0 = prob
+  @unpack sol, S, window, Δt, diffcache, b, w, v, η, res, g, g0, umid = prob
   @unpack wBinv, wEinv, B, E, Smat = S
   @unpack dg_val, pgpu, pgpu_config, pgpp, pgpp_config, numparams, numindvar, uf = diffcache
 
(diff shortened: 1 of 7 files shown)


src/lss.jl
```diff
@@ -0,0 +1,422 @@
...
+  @unpack wBinv, wEinv, B, E, Smat = S
+  @unpack dg_val, pgpu, pgpu_config, pgpp, pgpp_config, numparams, numindvar, uf = diffcache
+
+
+  n0 = searchsortedfirst(sol.t, sol.t[1]+t0skip)
+  n1 = searchsortedfirst(sol.t, sol.t[end]-t1skip)
+
+  b!(b,prob)
+
+  ures = @view sol.u[n0:n1]
```
ChrisRackauckas, line 310: This code is making many assumptions that `u` is an `AbstractVector`. I think that's fine, but we should probably get an issue open about this limitation as something to address in the future.

src/lss.jl
```diff
@@ -0,0 +1,422 @@
...
+  umidres = @view umid[:,n0:n1-1]
+
+  # reset
+  res .*=false
+
+  for i=1:numparams
+    #running average
+    g0 *= false
+    bpar = @view b[:,i]
+    w .= Smat\bpar
```
ChrisRackauckas, line 320: Smat is constant in the loop right? Then you will want to hoist out `_Smat = lu(Smat)` so it's only factorized once.

ChrisRackauckas:
>It's quite a memory intense method.

Yeah, that's a bit expected given the work it does in normalizing the derivative.

> Partly, this could probably be solved by wrapping sparse around some of the matrices used in the shadowing problem. It's inherently a block diagonal structure, so we are sure that the matrices will always be sparse even if, e.g., the Jacobian is not sparse. 

You could use https://github.com/JuliaArrays/BlockArrays.jl to keep structure.

src/lss.jl
frankschae: yep, Smat is constant. Sounds good!

src/lss.jl
frankschae: we could probably in concrete_solve add a `! u isa AbstractVector && error`  to avoid unforeseen issues.

src/lss.jl
```diff
@@ -0,0 +1,428 @@
...
+      pf.u = u
+      jacobian!(pJ, pf, uf.p, f_cache, sensealg, paramjac_config)
+    end
+    tmp = @view b[(i-1)*numindvar+1:i*numindvar,:]
+    copyto!(tmp, pJ)
+  end
+  return nothing
+end
+
+function solve(prob::ForwardLSSProblem; t0skip=zero(prob.Δt), t1skip=zero(prob.Δt))
```
frankschae, line 295: is the name`solve` problematic? 
https://github.com/SciML/DiffEqSensitivity.jl/pull/420/checks?check_run_id=2797420128

src/lss.jl
ChrisRackauckas: Yeah, throw a nice error message before the user gets a worse one. It's always a good idea in these scenarios.

frankschae:
> You could use https://github.com/JuliaArrays/BlockArrays.jl to keep structure.

I didn't find an example in their docs for large sparse arrays. https://github.com/KristofferC/BlockSparseMatrices.jl looks related as well.

frankschae:
The [CI error in the GSA methods](https://github.com/SciML/DiffEqSensitivity.jl/pull/420/checks?check_run_id=2803576942) looks unrelated:

`type eFASTResult has no field first_order`

PR #422, opened 2021-06-14
Adjoint-mode for the LSS method

frankschae:
Implements the adjoint mode wrt #420 . Parameter-dependence in the loss function is now explicitly tested as well. I am not sure if windowing would work the same way, so I implemented only the time-dilation option. 

(see related issue: #102)

### DIFF
--- a/src/lss.jl
+++ b/src/lss.jl
@@ -1,9 +1,8 @@
-struct LSSSchur{wBType,wEType,BType,EType,SType}
+struct LSSSchur{wBType,wEType,BType,EType}
   wBinv::wBType
   wEinv::wEType
   B::BType
   E::EType
-  Smat::SType
 end
 
 struct LSSSensitivityFunction{iip,F,A,J,JP,S,PJ,UF,PF,JC,PJC,Alg,fc,JM,pJM,MM,CV,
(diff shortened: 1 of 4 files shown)


src/lss.jl
```diff
@@ -427,3 +427,158 @@ function __solve(prob::ForwardLSSProblem,sensealg::ForwardLSS,alpha::Cos2Windowi
...
+  __solve(prob,prob.sensealg,prob.sensealg.alpha,t0skip,t1skip)
+end
+
+function __solve(prob::AdjointLSSProblem,sensealg::AdjointLSS,alpha::Number,t0skip,t1skip)
+  @unpack sol, S, Δt, diffcache, h, b, wa, res, g, g0, umid = prob
+  @unpack wBinv, B, E, Smat = S
+  @unpack dg_val, pgpp, pgpp_config, numparams, numindvar, uf, f, f_cache, pJ, pf, paramjac_config = diffcache
+
+  b .= E*h + B*wBinv
+  wa .= Smat\b
```
ChrisRackauckas, line 552: pre-caching the Smat factorization here?

frankschae: Should this go to `AdjointLSSProblem` then? In this case `Smat\b` is only called once.

ChrisRackauckas: yeah, just `lu!(Smat)` before saving it.

ChrisRackauckas (approved):

src/lss.jl
ChrisRackauckas: is this the same as forward LSS? If it's not, we should probably get that updated. I forgot about `lu!` for that 😅

ChrisRackauckas:
Shadowing could be a good use case for DiffEqGPU?

src/lss.jl
frankschae: yup I updated it for both. `lu!(Smat)` just returned the `LU` decomposition though (i.e., `Smat` stayed the same), so I went for removing `Smat` from the `LSSSchur` struct and instead added a field storing the LU decomposition to the respective LSSProblem.

frankschae:
yeah I think it could benefit from GPUs - at least it's full of matrix-matrix or matrix-vector multiplications.

Issue #246, opened 2020-05-22
Proper chunk mode for `ODEForwardSensitivity`


YingboMa:
https://github.com/SciML/DiffEqSensitivity.jl/blob/c70e76252be6f85df190aa544a6faefbb67414a7/src/local_sensitivity/forward_sensitivity.jl#L155-L160

~It only works when there are <= 12 states because we are not re-evaluating the primal function to build the full Jacobian matrix.~

This works, but if the state is too large, the compile-time could blow up, and the performance may also be bad because it will start to heap allocate.

PR #426, opened 2021-06-21
Much more extensive default sensitivity choices

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/246


src/adjoint_common.jl
```diff
@@ -165,8 +165,20 @@ function adjointdiffcache(g,sensealg,discrete,sol,dg,f;quad=false,noiseterm=fals
         end
       end
     end
+
     if compile_tape(sensealg.autojacvec)
       paramjac_config = ReverseDiff.compile(tape)
+    elseif sensealg.autojacvec isa Bool && sensealg.autojacvec && DiffEqBase.isinplace(prob)
+      compile = try
+          !hasbranching(prob.f,copy(u0),u0,p,prob.tspan[2]))
```
ChrisRackauckas, line 173: @simeonschaub `prob.f` mutates the first argument: is the copy required here for safety?

src/concrete_solve.jl
```diff
@@ -2,24 +2,41 @@
...
-      else
-        InterpolatingAdjoint()
-      end
+      InterpolatingAdjoint(autojacvec=ZygoteVJP())
+    end
+  else
+    # Determine if we can compile ReverseDiff
+    compile = try
+        if DiffEqBase.isinplace(prob)
+          !hasbranching(prob.f,copy(u0),u0,p,prob.tspan[1])
```
ChrisRackauckas, line 26: note copy.

src/adjoint_common.jl
```diff
@@ -224,6 +236,17 @@ function adjointdiffcache(g,sensealg,discrete,sol,dg,f;quad=false,noiseterm=fals
           tapei = noisetape(i)
           if compile_tape(sensealg.noise)
             push!(paramjac_noise_config, ReverseDiff.compile(tapei))
+          elseif sensealg.noise isa Bool && sensealg.noise && DiffEqBase.isinplace(prob)
+              compile = try
+                  !hasbranching(prob.f,copy(u0),u0,p,prob.tspan[2]))
```
ChrisRackauckas, line 241: note copy

src/adjoint_common.jl
simeonschaub: Yes, the overdub will always run everything as well, so this is needed if you don't want the check to mutate `u0`. Might even need to be a `deepcopy` if elements of `u0` are allowed to be mutated as well.

PR #427, opened 2021-06-21
Add EnzymeVJP

ChrisRackauckas:
https://github.com/wsmoses/Enzyme.jl/blob/acc040083067b2a5c6eb9caeb81376a203940c83/src/Enzyme.jl#L96-L104

it only returns the pullback and not the value? @wsmoses


wsmoses:
Oh is the convention for pullback to also return the value? We should expose that as an option in the API. Shouldn't be too bad, but we probably shouldn't tell Enzyme that "returnValue=false" always (https://github.com/wsmoses/Enzyme.jl/blob/acc040083067b2a5c6eb9caeb81376a203940c83/src/compiler.jl) :P

cc @vchuravy

ChrisRackauckas:
yeah it's the standard because you're already computing it.

ChrisRackauckas:
@wsmoses it looks like I can't pullback a vector? MWE:

```julia
using Enzyme, Zygote
function f(du,u,p,t)
  du[1] = dx = p[1]*u[1] - p[2]*u[1]*u[2]*t
end

function foop(u,p,t)
  dx = p[1]*u[1] - p[2]*u[1]*u[2]*t
  dy = -p[3]*u[2] + p[4]*u[1]*u[2]
  [dx,dy]
end

p = [1.5,1.0,3.0,1.0]; y = [1.0,1.0]; t=0.0
back = Enzyme.pullback(y, p) do u, p
  vec(foop(u, p, t))
end
λ = [2.0;1.0]
tmp1,tmp2 = back(λ)

back = Enzyme.pullback(y, p) do u, p
  out_ = similar(u)
  f(du, u, p, t)
  vec(du)
end
λ = [2.0;1.0]
tmp1,tmp2 = back(λ)

dy, back = Zygote.pullback(y, p) do u, p
  vec(foop(u, p, t))
end
λ = [2.0,1.0]
tmp1,tmp2 = back(λ)
```

wsmoses:
I think here you want to use a duplicated argument via the more general autodiff function (see below):

```
using Enzyme

p = [1.5,1.0,3.0,1.0]; y = [1.0,1.0]; t=0.0

out = [0.0, 0.0]

tmp1 = similar(y); tmp1 .= 0
tmp2 = similar(p); tmp1 .= 0

λ = [2.0;1.0]

Enzyme.autodiff(Duplicated(y, tmp1),
                Duplicated(p, tmp2),
                Duplicated(out, λ)) do u,p,out
  out[1] = p[1]*u[1] - p[2]*u[1]*u[2]*1.0
  out[2] = -p[3]*u[2] + p[4]*u[1]*u[2]
  nothing
end
@show tmp1, tmp2, λ
# (tmp1, tmp2, λ) = ([2.0, -4.0], [2.0, -2.0, -1.0, 1.0], [0.0, 0.0])
([2.0, -4.0], [2.0, -2.0, -1.0, 1.0], [0.0, 0.0])
```

Note that here I explicitly use 1.0 instead of t as it appears that using `t` causes a `jl_apply_generic`, which @vchuravy are actually about to tackle, but nevertheless would error.

Also, having skimmed your use case you may want "split" AD (e.g. the forward pass runs, caches potentially overwritten values in an object for use in the reverse pass to be run later). This hasn't been fully exposed in a nice Julia API yet but most recently @Roger-luo was taking a look iirc (and if it's urgent can look at).

ChrisRackauckas:
>Also, having skimmed your use case you may want "split" AD (e.g. the forward pass runs, caches potentially overwritten values in an object for use in the reverse pass to be run later). This hasn't been fully exposed in a nice Julia API yet but most recently @Roger-luo was taking a look iirc (and if it's urgent can look at).

Nah we are doing the vjp immediately so we don't need to split it.

What does the `Duplicated` do and why is it `similar` to `y`? Is it mutating it and using the `tmp`s as some kind of cache?

wsmoses:
So basically Enzyme defines every active variable to have a shadow variable where derivatives are stored. If a pointer-like argument is passed, it must also pass the shadow pointer via the duplicated mechanism. Thus basically we are just taking advantage of the fact that Enzyme plays nicely with mutating values and even avoiding an extraneous allocation.

Since the adjoint does a `+=` to accumulate the gradient, we should pass an equivalent data structure to the primal, except with zeros where appropriate. I was lazy and copied some cuda code where we do this -- you could have instead said tmp1 = [0.0, 0.0].

Here, lambda is what the shadow of the output is (being the derivative we want to propagate), so we just pass that as the shadow for out.

Roger-luo:
>  but most recently @Roger-luo was taking a look iirc (and if it's urgent can look at).

Yeah I just had a rough look at the implementation last weekend, but I haven't get much time to actually do a PR yet, I mainly need it to hook things from Yao. (But I guess doesn't need me doing anything for this PR for now)

wsmoses:
Also, (again forgive my quick skim) since it looks like you might be iterating the value of lambda, using read-write memory might allow you to do something far more efficient:

```julia
using Enzyme

p = [1.5,1.0,3.0,1.0]; y = [1.0,1.0]; t=0.0

out = [0.0, 0.0]

tmp2 = similar(p); tmp2 .= 0

λ = [2.0;1.0]

Enzyme.autodiff(Duplicated(y, λ),
                Duplicated(p, tmp2)) do u,p
  dx = p[1]*u[1] - p[2]*u[1]*u[2]*1.0
  dy = -p[3]*u[2] + p[4]*u[1]*u[2]
  u[1] = dx
  u[2] = dy
  nothing
end
@show tmp2, λ
```

ChrisRackauckas:
```julia
using Enzyme

p = [1.5,1.0,3.0,1.0]; y = [1.0,1.0]; t=0.0

out = [0.0, 0.0]

tmp1 = similar(y); tmp1 .= 0
tmp2 = similar(p); tmp1 .= 0

λ = [2.0;1.0]

Enzyme.autodiff(Duplicated(y, tmp1),
                Duplicated(p, tmp2),
                Duplicated(out, λ)) do u,p,out
  out[1] = p[1]*u[1] - p[2]*u[1]*u[2]*1.0
  out[2] = -p[3]*u[2] + p[4]*u[1]*u[2]
  nothing
end
@show tmp1, tmp2, λ
# (tmp1, tmp2, λ) = ([2.0, -4.0], [1.9911074014132516e209, 1.9034751571459084e214, 7.172658975525257e252, 9.198806578341978e63], [0.0, 0.0])
```

is master of some package required for that example to work?

ChrisRackauckas:
Using the following to assess performance:

```julia
using DiffEqSensitivity, OrdinaryDiffEq, Zygote
using Test, ForwardDiff

function fiip(du,u,p,t)
end
function foop(u,p,t)
  dx = p[1]*u[1] - p[2]*u[1]*u[2]
  dy = -p[3]*u[2] + p[4]*u[1]*u[2]
  [dx,dy]
end

loss = (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,abstol=1e-14,reltol=1e-14,saveat=0.1,sensealg=InterpolatingAdjoint()))
@time du02,dp2 = Zygote.gradient(loss,u0,p);

loss = (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,abstol=1e-14,reltol=1e-14,saveat=0.1,sensealg=InterpolatingAdjoint(autojacvec=ReverseDiffVJP(true))))
@time du02,dp2 = Zygote.gradient(loss,u0,p);

loss = (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,abstol=1e-14,reltol=1e-14,saveat=0.1,sensealg=InterpolatingAdjoint(autojacvec=EnzymeVJP())))
@time du02,dp2 = Zygote.gradient(loss,u0,p);

using Profile
@profile Zygote.gradient(loss,u0,p);
Juno.profiler()
Profile.clear()
```

seadra:
Is it possible to make Enzyme.jl dependency optional and use Zygote instead? [This Enzyme issue](https://github.com/wsmoses/Enzyme.jl/issues/69) renders DiffEqFlux unusable on Arch Linux, and based on the response in that issue, it's not clear if they will eventually address it or not.

ChrisRackauckas:
It looks like the answer is to use the standard binary which is suggested for all Julia installations anyways? Is there any issue with the generic binary?

Enzyme.jl is optional as in it's a choice to be used for vjps. Zygote is still what's used in general. This fixes mutation, so we want it to show up in the automatic algorithm selection.

seadra:
It's a valid workaround for most users who are not running in restrictive computing nodes.

The problem with julia-bin is that it's from user-maintained repository and is not a part of Arch Linux. It's usually similar for other distros, in that they don't use third-party binaries (except for certain proprietary non-free packages like skype, nvidia drivers etc).

ChrisRackauckas:
Let's be correct with our words here: it's not a "workaround", it's simply incorrect to be using the wrong version of LLVM. Using a completely untested LLVM version could give you wrong numerical values at any point in the computation for any reason: that's not good or suggested behavior. From this I have opened an inquiry with the compiler team about how this has happened, and the suggestion is to put a static assert in Julia itself so that it will fail at build time asserting that it's using an incorrect version of the compiler. That would make the incorrect build fail at build time and make it clear to Arch that it didn't create a usable binary. 

I don't think it should fall into the domain of packages to support alternative LLVMs because there's absolutely no expectation that alternative LLVMs even pass Julia's basic `a+b+c` tests, let alone match the syntax of the correct LLVM version close enough to make compiler transformations like AD work.

seadra:
"A completely untested LLVM version" by whom? A large number of Gentoo and Arch Linux users have already been using it that way, for years actually. 

LLVM 12 is the current *stable* release of LLVM, and it *is* a workaround from the OS package management perspective if you're using Arch Linux.

A static assert that locks Julia into an old release of LLVM would certainly lead to some difficult decisions for Arch Linux and Gentoo package managers.

ChrisRackauckas:
>"A completely untested LLVM version" by whom? The package builds do run the available tests (at the the ones that don't check for hardcoded version numbers), and a large number of Gentoo and Arch Linux users have already been using it that way, for years actually.

And has been a common source of bugs for years.

https://github.com/SciML/DifferentialEquations.jl/issues/737
https://github.com/SciML/DifferentialEquations.jl/issues/527
https://github.com/SciML/DifferentialEquations.jl/issues/663

etc. So yes, I know people have been using it for years, because those Linux distros have been shipping slightly broken versions for years and have made it my problem to tell people that. It's tested in the sense that it's known to give incorrect results and errors. It's missing the patches applied to Base which are required for numerical accuracy. Simple things like `^` can be defined incorrectly from that kind of thing. It would be a good thing to throw the error sooner to people so that way they would know. LLVM 12 is the stable release, but it gives numerically incorrect values for things which Julia has already patched but not upstreamed. So the non-patched versions almost certainly do not pass all of Julia Base's tests.

>A static assert that locks Julia into an old release of LLVM would certainly lead to some difficult decisions for Arch Linux and Gentoo package managers.

That might be better than shipping a version that literally fails tests and silently lets users think it's an appropriate and correct Julia binary. Explicit errors are usually better than silent bugs. What has happened here is that Enzyme has just made the error more explicit, letting you know you won't necessarily get correct results with that build, and I'm inclined to believe that's not a bad thing. I mean, if the differential equation solver just gave out bad results instead of erroring, wouldn't that be a mess? I would think the underlying numerical tooling should have as much defensive erroring as possible.

If you care about this a lot, then talk to the people working on upstreaming the patches to LLVM so they are all in LLVM 13 and we're all happy. But I don't work on that area, I am just trying to make sure there's a big fast numerical stack that is correct, and part of that means I need to assume someone is using a correct and working version of Julia.

vchuravy:
> If you care about this a lot, then talk to the people working on upstreaming the patches to LLVM so they are all in LLVM 13 and we're all happy.

Even if we upstream all patches in time for LLVM 13, there will yet again be other patches. and other patches again. We are very actively upstreaming the bug fixes we find, but we find them *after* the LLVM release. If someone wants to pay for the effort that it would take for us to find them *before* the LLVM release I would gladly take their money and hire 1-2 FTE for that.

vchuravy:
>"A completely untested LLVM version" by whom? The package builds do run the available tests (at the the ones that don't check for hardcoded version numbers), and a large number of Gentoo and Arch Linux users have already been using it that way, for years actually.

The ArchLinux&Gentoo maintainers don't get the bug-reports. We do, so yes the configuration is untested and unsupported by the Julia developers who respond to the bug reports. It takes significant effort to isolate and minimize LLVM related bugs. 

ArchLinux actually used to ship the Julia LLVM Version https://bugs.archlinux.org/task/40231, but someone seems to have changed it back to the system LLVM since then. (Apparently in August 2020)

seadra:
I regularly get bitten by bugs like that, but I would instead argue that those are bugs that should be fixed anyway.

Such unbundling of interned dependencies by package managers is very common and normal, and from their perspective, it is weird to ask for exceptions. If a package (julia in this case) has dependencies (llvm or a fork of it), it should in general be a separate package.

If Julia requires a heavily patched version of LLVM, one way to work with Linux distributions (and not against) would be to rename that version of LLVM (including the library file names) to something else. Then package managers can create a package for that, which won't conflict with the already existing package name llvm11, and use that for Julia.

> If you care about this a lot, then talk to the people working on upstreaming the patches to LLVM 

Given the current situation despite the efforts by Julia team, I don't believe anything I would say would realistically make an iota of difference on their development process.

vchuravy:
> If Julia requires a heavily patched version of LLVM, one way to work with Linux distributions (and not against) would be to rename that version of LLVM (including the library file names) to something else.

That is what we effectively do. Our version of LLVM is called LLVM-11jl and we use symbol versioning.

Sorry that I sound so annoyed or bitter about this, but the fact that Linux maintainer insist on treating LLVM as a normal dependency has caused me quite a bit of work... 

Just to make the point, when we (the Julia maintainers) Update the LLVM version of Julia we run the regression tests of the entire Julia ecosystem and we don't upgrade until we at least understand the failures and we don't release a version of Julia until all the found bugs are resolved. This takes time and a lot of effort.

You mentioned that we should provide a renamed fork of LLVM, I have been thinking of releasing the patched source tree (since right now I maintain the patch list in two places, which is one more than I would want). If there is interest from some of the Linux packagers I can take time do that.

seadra:
On Arch Linux and Gentoo, there is a relatively easy process for submitting packages.

Anyone can actually create packages on the [AUR](https://wiki.archlinux.org/title/AUR_submission_guidelines), and if a patched source tree is readily available, a pair of packages llvm-11jl and julia-official (or some other names) that makes use of it can be prepared.  Then, based on this julia-official, a proposal can be made for the [julia package](https://github.com/archlinux/svntogit-community/blob/master/julia/repos/community-x86_64/PKGBUILD#L80) to use llvm-11jl, which would require the inclusion of llvm-11jl in the official repositories.

The [process](https://wiki.gentoo.org/wiki/Submitting_ebuilds) is slightly different on Gentoo, but definitely much easier than the heavily bureaucratic process in Debian.

vchuravy:
I do use Archlinux personally, but I do not have the bandwidth to take on the additional burden of maintaining the Julia package there... I personally just use the julia-bin AUR

nalimilan:
FWIW, in Fedora I bundle LLVM with the Julia package instead of using the system LLVM. Originally I applied Julia-specific patches to a separate LLVM package, but in the end this I realized it was painful without any clear benefit so I stopped. What's the point of creating a separate package if Julia is the only user?

seadra:
It is more about the policy of packaging and package maintenance rather than practical benefits, but at least in the context of Arch Linux, [multiple variants of julia in AUR](https://aur.archlinux.org/packages/?O=0&SeB=nd&K=julia&outdated=&SB=n&SO=a&PP=50&do_Search=Go) (mkl variants and git versions) can benefit from such a package.

Assuming the availability of the patched repo, would it actually require a lot changes to the current [LLVM 11 PKGBUILD?](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=llvm11)?

seadra:
Yes, unfortunately, there is no possibility for julia-bin to become a part of the official packages, because it is a binary distribution of a FOSS project.

I can volunteer for creating and maintaining the llvm-julia package and a push for its usage in the official julia package, provided that a patched LLVM repo exist (something akin to [wine-staging](https://github.com/wine-staging/wine-staging) for which [an official package](https://archlinux.org/packages/multilib/x86_64/wine-staging/) exists in Arch Linux as well).

nalimilan:
>  [multiple variants of julia in AUR](https://aur.archlinux.org/packages/?O=0&SeB=nd&K=julia&outdated=&SB=n&SO=a&PP=50&do_Search=Go) (mkl variants and git versions) can benefit from such a package.

But git versions of Julia will likely need a different LLVM. For each release cycle, there will be a point where Julia from git master needs LLVM N+1 while the current release uses version N.

seadra:
Sure, that can be done in a separate package like llvm-julia-git which tracks the master branch of julia's llvm repo.

nalimilan:
Yes but what's the point then? It just takes more packaging work for zero benefit...

seadra:
The point would be to support julia-git with properly patched LLVM, for those who prefer to use the git version. It would be a minimal, mostly effortless work on the packaging side, which I'd be happy to do, provided that a repo with proper branches exists.

Anyway, that's just tangential. The main issue is the julia (and julia-mkl) packages.

PR #428, opened 2021-06-21
ForwardDiffSensitivity for u0

ChrisRackauckas:
Everything works except the thunks. @yingboma @oxinabox am I doing the thunks wrong?

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -242,35 +242,8 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
                                  save_idxs = nothing,
                                  kwargs...)
   _save_idxs = save_idxs === nothing ? (1:length(u0)) : save_idxs
-  pdual = seed_duals([vec(p);u0],prob.f)
-  u0dual = convert.(eltype(pdual),u0)
-
-  if (convert_tspan(sensealg) === nothing && (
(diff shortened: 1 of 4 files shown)


src/concrete_solve.jl
```diff
@@ -211,7 +211,10 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,sensealg::AbstractForwardSe
        end
        J'v
      end
-     (nothing,nothing,nothing,adj,nothing,ntuple(_->nothing, length(args))...)
+     du0 = ChainRulesCore.@thunk begin
+         error("ForwardSensitivity does not differentiate with respect to u0. Change your sensealg.")
+     end
```
oxinabox, line 216:
```suggestion
     du0 = ChainRulesCore.@not_implemented(
         "ForwardSensitivity does not differentiate with respect to u0. Change your sensealg."
     )
```
https://juliadiff.org/ChainRulesCore.jl/dev/api.html#ChainRulesCore.@not_implemented-Tuple{Any}

@devmotion  added `@not_implemented` returning it's own differential type.
Because if an AD system like Zygote eagerly `unthunk`s everything, then even if this output is not used  it would still error.

ChrisRackauckas:
Tested locally, and needed for https://github.com/SciML/DiffEqBase.jl/pull/673 so I'm going to rush this through, tag, and see what's up.

PR #434, opened 2021-06-25
Use unperturbed norm in reverse mode forwards

ChrisRackauckas:
While the correctness of https://github.com/SciML/DiffEqBase.jl/pull/529 works for most users, the OrdinaryDiffEq AD events tests showed that norm is not necessarily a good idea for this forward reverse thing, and the reason is because it can slightly change the event locations. This can cause the number of steps to change in the forward pass because you can be off by 100eps(t), get a saved step, and then an event. To correct for this, probably the best thing to do is to re-solve with the same t going forward, which is somewhat required for the backpass to generally be well-defined. While this does lose some robustness guarantees, it does seem to be the only correct solution, other than filling all of the duals before the backpass, which could be a waste due to the thunking.

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -366,7 +366,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
               _saveat = saveat
             end
 
-            _sol = solve(_prob,alg,args...;saveat=sol.t,save_idxs = _save_idxs, kwargs...)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Got all the way back to https://github.com/SciML/DiffEqSensitivity.jl/pull/274 ... yeah it's impossible to make that one pass with this approach.

PR #436, opened 2021-06-26
Fix forward event times

ChrisRackauckas:
Yeah, this one is complicated. To understand it, you kind of need to understand why the other approaches failed:

https://github.com/SciML/DiffEqSensitivity.jl/pull/435
https://github.com/SciML/DiffEqSensitivity.jl/pull/434

Essentially, you need the dual numbers in the callbacks otherwise there are cases where derivatives are not bounded by error tolerance at all. Period, no change whatsoever given the definition of the errors. But this means that your time stepping in the pure forward pass is different, and every chunk is slightly different to. That is kind of an issue for chunked forward mode when it's used in reverse mode, but we proceed. 

So the first idea was to track events to pull out the event times, but it turns out the thread-safety copy behavior makes this really hard. So the most sane solution seemed to be to discard the event times by checking non-uniqueness. In reality this is only a partial solution because someone could define events with save_positions=(true,false) or (false,true), and want to use that in a cost function. In practice, defining cost functions w.r.t. times which are not defined in advance is really... ill-defined? 

So this literally only shows up if the user somehow defined a cost function which uses whatever random t's the integrator spits out (so, not some fit against data), and if they've modified the callback so that it saves at only a single extra random point at each event. The only case I can think of even remotely sane for that is somehow a loss function on a Poincure diagram, but that's ill-defined because it's chaotic, so... I'm punting on that. What will happen in this case is someone will get an error saying that `du` or `dp` is not sized correctly to match the solution. If that error shows up in a user code we can handle it (possibly by just having them try UNPERTURBED_NORM), so at least this PR is safe and will throw an appropriate error in the bad case.

That said, this does show that chunked forward mode in reverse is a little ill-defined, and so that motivates bringing back a fully non-chunked mode. The things you learn from weird user examples...

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -481,8 +481,8 @@ function out_and_ts(_ts, duplicate_iterator_times, sol)
       (d ∉ _ts) && push!(_ts, d)
     end
 
-    u1 = sol(_ts)[:]
-    u2 = sol(duplicate_times,continuity=:right)[:]
+    u1 = sol(_ts).u
+    u2 = sol(duplicate_times,continuity=:right).u
     saveat = vcat(_ts,  duplicate_times...)
     perm = sortperm(saveat)
     ts = saveat[perm]
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -299,9 +299,27 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
     _saveat = saveat
   end
 
-
-
   sol = solve(remake(prob,p=p,u0=u0),alg,args...;saveat=_saveat,save_idxs = _save_idxs, kwargs...)
+
+  # saveat values
+  # seems overcomplicated, but see the PR
+  if length(sol.t) == 1
+      ts = sol.t
+  else
+      ts = eltype(sol.t)[]
+      if sol.t[2] != sol.t[1]
+          push!(ts,sol.t[1])
+      end
+      for i in 2:length(sol.t)-1
+          if sol.t[i] != sol.t[i+1] && sol.t[i] != sol.t[i-1]
+              push!(ts,sol.t[i])
+          end
+      end
+      if sol.t[end] != sol.t[end-1]
+          push!(ts,sol.t[end])
+      end
+  end
+
   function forward_sensitivity_backpass(Δ)
     dp = @thunk begin
 
@@ -366,7 +384,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
               _saveat = saveat
             end
 
-            _sol = solve(_prob,alg,args...;saveat=sol.t,save_idxs = _save_idxs, kwargs...)
+            _sol = solve(_prob,alg,args...;saveat=ts,save_idxs = _save_idxs, kwargs...)
             _,du = extract_local_sensitivities(_sol, sensealg, Val(true))
 
             _dp = sum(eachindex(du)) do i
@@ -446,7 +464,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
               _saveat = saveat
             end
 
-            _sol = solve(_prob,alg,args...;saveat=sol.t,save_idxs = _save_idxs, kwargs...)
+            _sol = solve(_prob,alg,args...;saveat=ts,save_idxs = _save_idxs, kwargs...)
             _,du = extract_local_sensitivities(_sol, sensealg, Val(true))
 
             _du0 = sum(eachindex(du)) do i
--- a/test/autodiff_events.jl
+++ b/test/autodiff_events.jl
@@ -1,0 +1,62 @@
+using DiffEqSensitivity
+using OrdinaryDiffEq, Calculus, Test
+using Zygote
+
+function f(du,u,p,t)
+  du[1] = u[2]
+  du[2] = -p[1]
+end
+
+function condition(u,t,integrator) # Event when event_f(u,t) == 0
+  u[1]
+end
+
+function affect!(integrator)
+  @show integrator.t
+  println("bounced.")
+  integrator.u[2] = -integrator.p[2]*integrator.u[2]
+end
+
+cb = ContinuousCallback(condition, affect!)
+p = [9.8, 0.8]
+prob = ODEProblem(f,eltype(p).([1.0,0.0]),eltype(p).((0.0,1.0)),copy(p))
+
+function test_f(p)
+  _prob = remake(prob, p=p)
+  solve(_prob,Tsit5(),abstol=1e-14,reltol=1e-14,callback=cb,save_everystep=false)[end]
+end
+findiff = Calculus.finite_difference_jacobian(test_f,p)
+findiff
+
+using ForwardDiff
+ad = ForwardDiff.jacobian(test_f,p)
+ad
+
+@test ad ≈ findiff
+
+function test_f2(p, sensealg=ForwardDiffSensitivity(), controller=nothing, alg=Tsit5())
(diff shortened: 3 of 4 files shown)


ChrisRackauckas:
Chris Rackauckas  37 minutes ago
yeah, it's a much harder one to explain, but essentially ODE solver error estimates are (rightfully) insensitive to things which they can integrate exactly. So they say "I have zero error on this", and keep growing dt. But, the sensitivity equations themselves are ODEs, and they may not have that property even while the original ODEs do. So if you don't change the adaptive time stepping to account for the terms in the dual part, you can end up with O(1) error on these equations. And these show up in pharmacology, so :sweat_smile: .

Chris Rackauckas  36 minutes ago
But that means that your time stepping is now dependent on the derivatives you're calculating: there's no independent truth to what "the forward pass" was: it can adapt based on how many duals you ended up chunking by. Normally that's not an issue: the user usually says "save at every dt=1", you interpolate to those points, and you're good :heavy_check_mark: .

Chris Rackauckas  34 minutes ago
But now if you put chunked forward mode into the reverse pass, you need to make sure that chunked forward mode has the time steps at the same spot as the forward pass, otherwise the outputs don't align. So if the user happened to define a cost function where the adaptive time stepping chose the outputs (so this weird case without data?), then you have to be careful

Chris Rackauckas  34 minutes ago
So the obvious thing is "well for the derivative chunks just say 'save at the times for this specific forward pass, the one that I returned as the primal'", and that is almost good, but it has an issue :wink:

Chris Rackauckas  33 minutes ago
Someone can define events that say "add a saved point whenever there is an event"

Chris Rackauckas  33 minutes ago
So now since time stepping changes slightly in the chunked pass, those event time points can be off. Probably not by more than like 1e-15, but they are different time points.

Chris Rackauckas  32 minutes ago
So if you said "save at all points that other pass did", you can get a save at 40.1014141 (the event time of the forward pass), and `40.1014141+1e-15 (the new event time now that adaptivity has slightly changed). And tada: now you have too many points to reverse from!

Lyndon White  31 minutes ago
I think I see how that works.
Let me repeat it back to you
Basically computing sensititivity  reverse will be assuming that you have exact same time stepping as primal only computation.
(And so are allowed to refer to those time steps)
But when you do it via computing full jacobian with fowards AD with an extended ODE you need to do different time stepping to control for error in the derivatives.
And I guess the chunking comes in to it because different bases elements being progated might mean needing different time-stepping again?
I think I don't quite have it right

Chris Rackauckas  29 minutes ago
yeah: if you don't make the stepping dependent on dual numbers, you have cases with unbounded error. So you make the stepping dependent on the dual numbers. But that means each chunk can step differently. Normally you can interpolate to re-align them and no one will care. But, if someone also makes events where they say "save whenever u(t) == 1 , and change u(t)+=1", the time at which that will occur will also slightly change. So what do you do: interpolate back to the original times? Now you're missing the +=1 part. Keep the event at the unperturbed times?

Lyndon White  28 minutes ago
I see.

Lyndon White  27 minutes ago
But its reasonable in general to say "Stop referring to type steps that you didn't ask for explictly"?

Chris Rackauckas  27 minutes ago
So I drop the saves of the primal for the events, so  the chunks can save at their events which will be at a slightly different time, and that's a nice hack to solve most cases. But, there's a counter example to that.

Chris Rackauckas  26 minutes ago
Yeah, the reasonable thing is to say "your loss function should only be dependent on time points you've asked for". In general that's true: you have fixed data points at fixed points in time, so you want the integrator to give you the value at those points. For that case, there's no issue.

Lyndon White  26 minutes ago
This is pretty cool

Chris Rackauckas  25 minutes ago
It only shows up if you really care about the random points the ODE solver spits out, and if you really care about it spitting out the points at which specific events occurred. And if you really care about chunked forward mode. Those things together gives a slightly ill-defined behavior, if you want to make the stepping control the derivative error.

Chris Rackauckas  24 minutes ago
Yeah, and it's all because "forward mode AD should compute the same primal as when not doing AD" is necessarily violated in this case if you don't want unbounded error on the derivatives. But if your primal can now calculate in a different way only when differentiating, you need to be really careful.
:+1:
1


Lyndon White  24 minutes ago
You should really care about chunked forwards more though.
If you are doing reverse via computing jacobian in forwards, it is the thing you want.
Caring about the random points the ODE solver spits out is weird
I don't get why this only occurs in chunked forwards though

Chris Rackauckas  23 minutes ago
If you aren't chunking, then you can just solve the primal and the derivatives all in one go, and then reverse via that Jacobian

Chris Rackauckas  23 minutes ago
one solve, boom you're good.

Lyndon White  23 minutes ago
Oh, I thought you could do that for chucked also?

Chris Rackauckas  22 minutes ago
With chunking you'll need to keep re-running it with different dual number seeds

Chris Rackauckas  22 minutes ago
and if the norm in the diffeq error handling takes into account those dual numbers, every seed choice can give potentially different stepping

Chris Rackauckas  22 minutes ago
So the issue isn't necessarily that we aren't doing ForwardDiff on the primal call, it would even occur if we did one chunk in the primal and finished the other chunks in the adjoint.

Lyndon White  21 minutes ago
With chunking you'll need to keep re-running it with different dual number seeds
because you are unchunking it.
Can't we change the system of equations to be able to handle a whole chunk?
Or at least up to a given number of basis elements?

Chris Rackauckas  21 minutes ago
chunking == solving more than once

Lyndon White  20 minutes ago
oh that is the opposite of how I was using the term

Lyndon White  20 minutes ago
I mean propagating a whole basis at a time

Chris Rackauckas  20 minutes ago
I mean it in the ForwardDiff sense.

Lyndon White  20 minutes ago
so did I, i thought.

Chris Rackauckas  19 minutes ago
You want to differentiate w.r.t. 24 quantities, then it chunks it (12,12) and calls f twice.

Lyndon White  19 minutes ago
Oh, I see.
Because of cache sizes or something that has that being faster than doing all 24 basis elements at once?

Chris Rackauckas  18 minutes ago
It's mostly to get around compiler limitations

Lyndon White  18 minutes ago
because it stops specializing after 16?

Chris Rackauckas  18 minutes ago
a dual of 100 partials is a tuple of 101 elements for each number (:scream: )

Chris Rackauckas  18 minutes ago
yeah so it splits it up

Lyndon White  18 minutes ago
I see.

Chris Rackauckas  17 minutes ago
But here f's behavior is then dependent on the values of the dual numbers in there, so... :boom:

Lyndon White  17 minutes ago
So if doing chucked, it might use different time steps for each chunk and for what the primal would have done on its own
if doing unchuncked then will get the same timestep all chunks and the primal you run, but that still might not align with behavour of primal on its own.

Chris Rackauckas  15 minutes ago
and if you don't change f 's behavior based on the dual numbers, you can technically get a derivative, it's just not always dependent on the abstol or reltol of the solver, and so there are even some cases where the derivative gives unbounded error even if the primal is well-behaved.

Chris Rackauckas  15 minutes ago
So ggnore.

Chris Rackauckas  14 minutes ago
One idea would be to say "screw it, just fall back to the autodiff norm of not changing the algorithm", but that would mean you could get inaccurate derivatives in real use cases

Chris Rackauckas  12 minutes ago
and it's even worse with stiff ODE solvers, the Sundials docs even mention it on page 139: https://computing.llnl.gov/sites/default/files/cvs_guide-5.7.0.pdf

Chris Rackauckas  11 minutes ago
on page 28 you can see the forward sensitivities are just the jvp done out by hand, so it's equivalent to forward-mode AD. And by default they make what would essentially be the dual terms change the norm for precisely this reason.

Chris Rackauckas  11 minutes ago
But now if you generate those equations by AD, what's the right behavior? :shrug: AD hell.

PR #437, opened 2021-07-01
Non-Intrusive Least Square Shadowing (NILSS)

frankschae:
Implements the NILSS algorithm introduced in https://arxiv.org/abs/1611.00880 with a (manual) segmentation routine that takes the number of segments and steps on each segment. Code is based on https://github.com/nishaChandramoorthy/energies and https://github.com/niangxiu/nilss .

It has a new overload for `ForwardDiff()` which solves an augmented vector for `vstar` (1 x inhomogenous equation in the paper, i.e., standard forward sensitivity) and `w` (number of unstable components x homogenous equations) . 

Related issue: #102


src/forward_sensitivity.jl
```diff
@@ -98,6 +114,9 @@ end
 function ODEForwardSensitivityProblem(f::DiffEqBase.AbstractODEFunction,u0,
                                     tspan,p=nothing,
                                     alg::ForwardSensitivity = ForwardSensitivity();
+                                    nus=nothing, # determine if Nilss is used
+                                    w0=nothing,
+                                    v0=nothing,
```
ChrisRackauckas, line 119: what are these?

frankschae: They allow us to pass the renormalized input values of the sensitivity. For example, `v0` starts at `zeros(numindvar)`. After the first segment, the qr decomposition is used to renormalize the sensitivity. Then we want to pass that as the initial condition.

ChrisRackauckas: oh I see.

ChrisRackauckas (approved):

PR #438, opened 2021-07-03
Optimize ForwardSensitivity

YingboMa:


src/forward_sensitivity.jl
```diff
@@ -57,6 +59,26 @@ function ODEForwardSensitivityFunction(f,analytic,tgrad,jac,jac_prototype,sparsi
...
+  jac_prototype === nothing && (jac_prototype = u0 .* u0' .* false)
+  jac_prototype = ForwardSensitivityJacobian(jac_prototype)
+  mass_matrix = mass_matrix === I ? I : BlockDiagonal(mass_matrix, Diagonal(I, size(J, 1)))
+  colorvec !== nothing && (colorvec = [colorvec; colorvec])
+  newjac = if DiffEqBase.has_jac(sense)
+    let jac=sense.jac, n=length(u0)
+      (J, u, p, t) -> begin
+        @assert J isa DiffEqBase.ForwardSensitivityJacobian
+        J = parent(J)
+        jac(J, u[1:n], p, t)
```
ChrisRackauckas, line 73: view?

src/forward_sensitivity.jl
```diff
@@ -57,6 +59,26 @@ function ODEForwardSensitivityFunction(f,analytic,tgrad,jac,jac_prototype,sparsi
   return sensefun
 end
 
+function sensefun2odefun(sense::ODEForwardSensitivityFunction, u0)
+  @unpack jac_prototype, jac, sparsity, uf, J, jac_config, alg, f_cache, mass_matrix, colorvec = sense
+  jac_prototype === nothing && (jac_prototype = u0 .* u0' .* false)
+  jac_prototype = ForwardSensitivityJacobian(jac_prototype)
+  mass_matrix = mass_matrix === I ? I : BlockDiagonal(mass_matrix, Diagonal(I, size(J, 1)))
+  colorvec !== nothing && (colorvec = [colorvec; colorvec])
```
ChrisRackauckas, line 67: if it is nothing, `repeat(1:div(n,2),2)` ?

YingboMa: Yeah, something like that. This isn't quite right. It should only take one Jacobian. So we should always define an analytic Jacobian for a sensefun. The mass matrix is also incorrect.

ChrisRackauckas: Yes on both parts.

PR #444, opened 2021-07-10
Handle additional callback case

frankschae:
This fixes the broken tests for the case when the callback doesn't change the parameters but depends on it, e.g., in

`affect!(integrator) = (integrator.u .+= integrator.p[2]//8*sin.(integrator.u))`

`QuadratureAdjoint()` is a bit nasty to fix here because I think one needs the value of the adjoint \lambda before the loss function is applied on it. However, this value is not necessarily the left or right `continuity` limit of the adjoint solution. In this first attempt, I tried to subtract it manually for the additional vjp.

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -133,7 +133,9 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback,DiscreteCallback,
                               dgrad=dgrad, dy=dy)
 
         λ .= dλ
-
+        if !(sensealg isa QuadratureAdjoint)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Yeah QuadratureAdjoint and discontinuities is a somewhat odd combination anyways. Good the fix was found though.

ChrisRackauckas (approved):

frankschae:
oh -- something went wrong. It only passed the simple loss function but not the MSE tests. Maybe it's due to the indexing of the duplicate time points.

Issue #383, opened 2021-01-21
ContinuousCallback + save_positions=(true,true) + slicing


ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/pull/381 found that if the position is saved, i.e., save_positions=(true,true). Then, gradients are dropped when slices on the solution are used

```julia
du01,dp1 = Zygote.gradient(
  (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,
   callback=cb,
   sensealg=BacksolveAdjoint())[end]),u0,p)
```

then the gradient is incorrect. Likely there's an off-by-one kind of thing where we need to separate this save from the `saveat` and keep it separate from the points where the data is checked.

PR #445, opened 2021-07-12
State-dependent Continuous Callbacks for BacksolveAdjoint

frankschae:
The update rule is given by the implicit function theorem. We need to compute Eqs. (9) and (12) in https://arxiv.org/abs/2011.03902 (I checked those terms), i.e., essentially an additional term that depends on the loss and the condition (and derivatives of both) must be added to the gradient of the loss function. 

Some details wrt what I have so far:
- we need to adjust the loss function gradient only for implicitly defined events that enter *directly* into the loss function. That is, no correction is needed in the case of `save_positions=(false,false)` or if slices exclude the saved states (like in `abs2(_sol[1, end] - target)`). 
- if `save_positions=(true,true)`, we must adjust `λ` before and after the vjp handling of the affect function. To select the proper loss gradient we need to pass the `cur_time` Ref from the loss function callback(s) also to the reverse callback.
- The event condition needs to be differentiated with respect to `u` and `t`. Therefore, `sensealg` should also be passed to the reverse callback (e.g., for the choice between AD and FiniteDiff).
- The `ImplicitCorrection` struct holds all values that are necessary to differentiate the condition and extract the loss gradient value. It is `mutable` because `cur_time` is not known yet when `TrackedAffect` is defined. `cur_time`  is then re-set, once the reverse callback is constructed.

I haven't looked closely into `InterpolatingAdjoint()` yet. I just tried to fix extraction errors (such as reduction over empty collection, ..) for the use of a `DiscreteCallback` vs  a `ContinuousCallback` . I think it will still need some fine-tuning wrt. `checkpoints`, `tstops`, `dg`, etc. 

One problem which I however already noticed is that while for a `DiscreteCallback` the separation for the events work, the `ContinuousCallback` was triggered twice. I'll look into this.

related issues: #383 #374

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -38,12 +38,12 @@ if GROUP == "All" || GROUP == "Core2"
     @time @safetestset "Autodiff Events" begin include("autodiff_events.jl") end
     @time @safetestset "Null Parameters" begin include("null_parameters.jl") end
     @time @safetestset "Forward Mode Prob Kwargs" begin include("forward_prob_kwargs.jl") end
-    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
     @time @safetestset "Steady State Adjoint" begin include("steady_state.jl") end
     @time @safetestset "Concrete Solve Derivatives of Second Order ODEs" begin include("second_order_odes.jl") end
 end
 
 if GROUP == "All" || GROUP == "Core3"
+    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
     @time @safetestset "Shadowing Tests" begin include("shadowing.jl") end
 end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -43,7 +43,6 @@ if GROUP == "All" || GROUP == "Core2"
 end
 
 if GROUP == "All" || GROUP == "Core3"
-    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
     @time @safetestset "Shadowing Tests" begin include("shadowing.jl") end
 end
 
@@ -69,6 +68,10 @@ if GROUP == "DiffEqFlux"
     @time @safetestset "ForwardDiff Sparsity Components" begin include("downstream/forwarddiffsensitivity_sparsity_components.jl") end
     @time @safetestset "SDE - Neural" begin include("downstream/sde_neural.jl") end
     @time @safetestset "Complex No u" begin include("downstream/complex_no_u.jl") end
+end
+
+if GROUP == "Callbacks"
+    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
 end
 
 if GROUP == "GPU"
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -71,9 +71,9 @@ if GROUP == "DiffEqFlux"
 end
 
 if GROUP == "Callbacks"
-    activate_downstream_env()
     @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
 end
+
 
 
 if GROUP == "GPU"
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -71,8 +71,10 @@ if GROUP == "DiffEqFlux"
 end
 
 if GROUP == "Callbacks"
+    activate_downstream_env()
     @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
 end
+
 
 if GROUP == "GPU"
     activate_downstream_env()
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,4 +1,18 @@
 steps:
+  - label: "Callbacks"
+    plugins:
+      - JuliaCI/julia#v1:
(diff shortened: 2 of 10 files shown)


frankschae:
I think that point:
> we need to adjust the loss function gradient only for implicitly defined events that enter directly into the loss function. That is, no correction is needed in the case of save_positions=(false,false) or if slices exclude the saved states (like in abs2(_sol[1, end] - target)).

might be wrong. I did a too quick initial check of that claim with an unlucky choice of the time span, where it works that way.  

```julia
function f(du,u,p,t)
  du[1] = u[2]
  du[2] = -p[1]
end

u0 = [5.0,0.0]
tspan = (0.0,2.0) #(0.0,5.0)
p = [10, 0.8]
prob = ODEProblem(f,u0,tspan,p)

### compute gradients based on discrete sensitivities (AD)
target = 1.0

##
### include events

# ContinuousCallback (implicit event)
condition(u,t,integrator) = u[1] # Event when condition(u,t,integrator) == 0
affect!(integrator) = integrator.u[2] = -integrator.p[2]*integrator.u[2]
cb = ContinuousCallback(condition,affect!,save_positions=(true,true))

# solve forward trajectory
sol = solve(prob,Tsit5(),callback=cb,
 save_everystep=false, save_start=false
 #saveat=[tspan[1],tspan[2]]
 )

### include dependence on discontinuity
function loss1(u0,p; sensealg=ReverseDiffAdjoint())
  _prob = remake(prob, u0=u0, p=p)
  _sol = Array(solve(_prob,Tsit5(),callback=cb,sensealg=sensealg,
             #saveat=[tspan[2]]
             save_everystep=false, save_start=false
             ))
  sum(abs2, _sol[1,:] .- target)
end

loss1(u0,p)

bounceFD1 = ForwardDiff.gradient((θ)->loss1(θ[1:2],θ[3:4]),[u0;p])

bounceBacksolve1 = Zygote.gradient((u0,p)->loss1(u0,p,sensealg=BacksolveAdjoint()),u0,p)
bounceInterpolating1 = Zygote.gradient((u0,p)->loss1(u0,p,sensealg=InterpolatingAdjoint(checkpointing=false)),u0,p)
bounceQuadrature1 = Zygote.gradient((u0,p)->loss1(u0,p,sensealg=QuadratureAdjoint()),u0,p)

@show bounceFD1 bounceBacksolve1 bounceInterpolating1 bounceQuadrature1
```
result:
```julia
bounceFD1 = [4.0000000000000195, 0.8000000000000179, -0.7999999999999997, 39.999999999999794]
bounceBacksolve1 = ([4.000000000000005, 0.8000000000000265], [-0.7999999999999861, 39.999999999999865])
bounceInterpolating1 = ([4.000000000000005, 0.8000000000000265], [-0.7999999999999827, 39.999999999999865])
bounceQuadrature1 = ([4.000000000000005, 0.8000000000000265], [-0.8000000000000007, 39.999999999999865])
```
changing  `tspan = (0.0,2.0)`, e.g., to `tspan = (0.0,1.5)`, then yields wrong gradient (`du0`) results.

```
bounceFD1 = [0.3500000000000114, -1.0499999999999923, 0.7874999999999948, 17.499999999999993]
bounceBacksolve1 = ([3.499999999999986, 2.100000000000013], [-0.7874999999999964, 17.499999999999794])
bounceInterpolating1 = ([3.499999999999986, 2.1000000000000134], [-0.787499999999998, 17.499999999999787])
bounceQuadrature1 = ([3.499999999999986, 2.1000000000000134], [-0.7875000000000079, 17.499999999999787])
```

frankschae:
The tests look good! :) I think it's ready for a review @ChrisRackauckas.
```julia
Test Summary:           | Pass  Broken  Total
Callbacks with Adjoints |  386      99    485
```
[ The error is just: ERROR: Package DiffEqSensitivity errored during testing (received signal: KILL) ..]
Should we maybe move the callback tests to a new buildkite group with juliacpu queue?

ChrisRackauckas:
yeah, move them.

test/runtests.jl
```diff
@@ -71,6 +70,10 @@ if GROUP == "DiffEqFlux"
     @time @safetestset "Complex No u" begin include("downstream/complex_no_u.jl") end
 end
 
+if GROUP == "Callbacks"
+    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
+end
```
ChrisRackauckas, line 75:
```suggestion
if GROUP == "Callbacks"
    activate_downstream_env()
    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
end

```

test/runtests.jl
```diff
@@ -71,6 +70,12 @@ if GROUP == "DiffEqFlux"
     @time @safetestset "Complex No u" begin include("downstream/complex_no_u.jl") end
 end
 
+if GROUP == "Callbacks"
+    activate_downstream_env()
+    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
+end
```
ChrisRackauckas, line 76:
```suggestion
if GROUP == "Callbacks"
    @time @safetestset "Callbacks with Adjoints" begin include("callbacks.jl") end
end

```

PR #458, opened 2021-08-01
Test master

ChrisRackauckas:


ChrisRackauckas:
@frankschae is this just RNG sensitivity now?

frankschae:
I would think so.. IIRC some bounds were pretty tight.

ChrisRackauckas:
Alright, let's take a pass through this in the near future to try and bump some bounds. Both StochasticDiffEq.jl and this repo are going to just upper bound RandomNumbers.jl for now.

PR #467, opened 2021-08-06
test ChainRules 1

ChrisRackauckas:


ChrisRackauckas:
Held back by https://github.com/TuringLang/DistributionsAD.jl/pull/184

DhairyaLGandhi:
I think @oxinabox  is tracking the changes in CR 1 that broke the tests here.

oxinabox:
I am. Though my time for digging into them is limited. I hope to have time next week.

I agree it probably was not a new change, in ChainRulesCore or in Zygote.
It might have been a new rule in ChainRules, exposing a underlying problem in the Zygote-ChainRules interface.
Or just that the Zygote tests hadn't been run in so long.

ChrisRackauckas:
Awesome.

PR #469, opened 2021-08-08
Add missing time update for uf

ArnoStrouwen:
If this is a bug, it is pretty bad it slips past tests.

### DIFF
--- a/test/forward.jl
+++ b/test/forward.jl
@@ -5,14 +5,14 @@ using DiffEqSensitivity: ForwardSensitivity
 
 function fb(du,u,p,t)
   du[1] = dx = p[1]*u[1] - p[2]*u[1]*u[2]
-  du[2] = dy = -p[3]*u[2] + u[1]*u[2]
+  du[2] = dy = -p[3]*u[2] + t*u[1]*u[2]
 end
 function jac(J,u,p,t)
   (x, y, a, b, c) = (u[1], u[2], p[1], p[2], p[3])
   J[1,1] = a + y * b * -1
-  J[2,1] = y
+  J[2,1] = t*y
   J[1,2] = b * x * -1
-  J[2,2] = c * -1 + x
+  J[2,2] = c * -1 + t*x
 end
 
 f = ODEFunction(fb,jac=jac)
--- a/test/forward.jl
+++ b/test/forward.jl
@@ -26,7 +26,7 @@ probvecmat = ODEForwardSensitivityProblem(fb,[1.0;1.0],(0.0,10.0),p,
 sol = solve(prob,Tsit5(),abstol=1e-14,reltol=1e-14)
 @test_broken solve(probInpl,KenCarp4(),abstol=1e-14,reltol=1e-14).retcode == :Success
 solInpl = solve(probInpl,KenCarp4(autodiff=false),abstol=1e-14,reltol=1e-14)
-solInpl2 = solve(probInpl,Rodas4(autodiff=false),abstol=1e-14,reltol=1e-14)
+solInpl2 = solve(probInpl,Rodas4(autodiff=false),abstol=1e-10,reltol=1e-10)
 solnoad = solve(probnoad,KenCarp4(autodiff=false),abstol=1e-14,reltol=1e-14)
 solvecmat = solve(probvecmat,Tsit5(),abstol=1e-14,reltol=1e-14)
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Yeah... could you add a test that catches this?

test/forward.jl
```diff
@@ -3,16 +3,18 @@ using DiffEqSensitivity,OrdinaryDiffEq, RecursiveArrayTools, DiffEqBase,
 using Test
 using DiffEqSensitivity: ForwardSensitivity
 
+using OrdinaryDiffEq
+using DiffEqSensitivity
```
ChrisRackauckas, line 7:
```suggestion
```

ArnoStrouwen:
For the test error, lower tol or increase iterations?

ChrisRackauckas:
It's off by 10^10. Something looks bad there.

ArnoStrouwen:
Is it perhaps because the stiff solvers work with the same jacobian object?
https://github.com/SciML/SciMLBase.jl/blob/4edec99cdabcec8266a628dc49e778b4a3ce41a7/src/function_wrappers.jl#L15
And messing with this object inside the (extended) ODE system causes trouble?
Everything works fine on Tsit5(), so it is specifically the stiff solvers causing the issue.

ArnoStrouwen:
I think the issue was just that Rodas4(autodiff=false) was only able to solve until 0.1s before maximum iterations was reached. 
The massive error in the test then was due to extrapolation to 5s?

ChrisRackauckas:
The fact that it diverged is pretty iffy.

ArnoStrouwen:
I can't find any sensitivity specific cause for the divergence.
It could be that the finite difference based jacobian (of the extended system) is iffy?

ChrisRackauckas:
It shouldn't be that bad on this system. Check the full forwarddiff one to see how stable the new model is.

ArnoStrouwen:
I tested if it was the finite difference based jacobian of the extended system by using ForwardDiffSensitivity instead of ForwardSensitivity:
```jl
probInpl3 = ODEForwardSensitivityProblem(fb,[1.0;1.0],(0.0,10.0),p,ForwardDiffSensitivity())
solInpl3 = solve(probInpl3,Rodas4(),abstol=1e-10,reltol=1e-10)
solInpl3 = solve(probInpl3,Rodas4(autodiff=false),abstol=1e-10,reltol=1e-10)      
```

```jl
julia> solInpl3 = solve(probInpl3,Rodas4(),abstol=1e-10,reltol=1e-10)
retcode: Success
Interpolation: specialized 3rd order "free" stiffness-aware interpolation
t: 4167-element Vector{Float64}:
  0.0
  0.0008227854154406597
  0.0018796649056780181
  0.0032608543701877967
  0.004916470576327273
  ⋮
  9.994272386460981
  9.996865914249916
  9.999470174260512
 10.0
u: 4167-element Vector{Vector{ForwardDiff.Dual{ForwardDiff.Tag{ODEFunction{true, typeof(fb), 
...

julia> solInpl3 = solve(probInpl3,Rodas4(autodiff=false),abstol=1e-10,reltol=1e-10)
retcode: Success
Interpolation: specialized 3rd order "free" stiffness-aware interpolation
t: 415556-element Vector{Float64}:
  0.0
  3.1827379411029245e-5
  4.3374170126018126e-5
  5.827269493778959e-5
  7.317121974956106e-5
  ⋮
  9.999858353275464
  9.999911268508027
  9.999964187597156
 10.0
u: 415556-element Vector{Vector{ForwardDiff.Dual{ForwardDiff.Tag{ODEFunction{true, typeof(fb), 
...
```

ChrisRackauckas:
Try TRBDF2? That does look unstable under finite differences.

ArnoStrouwen:
TRBDF2 works fine?
```jl
julia> solInpl3 = solve(probInpl3,TRBDF2(autodiff=false),abstol=1e-10,reltol=1e-10)
retcode: Success
Interpolation: 3rd order Hermite
t: 2959-element Vector{Float64}:

julia> solInpl3 = solve(probInpl3,TRBDF2(),abstol=1e-10,reltol=1e-10)
┌ Warning: `vendor()` is deprecated, use `BLAS.get_config()` and inspect the output instead
│   caller = isopenblas at linear_nonlinear.jl:86 [inlined]
└ @ Core C:\Users\arno\.julia\packages\DiffEqBase\uQlhE\src\linear_nonlinear.jl:86
retcode: Success
Interpolation: 3rd order Hermite
t: 2959-element Vector{Float64}:
```

ChrisRackauckas:
Yup, that's expected. So yes, it is a Jacobian accuracy with finite difference issue (SDIRK methods do not need accurate Jacobians, while non-W Rosenbrock methods, such as Rodas4, do).

ArnoStrouwen:
I rebased, please merge in if everything looks fine to you.

ChrisRackauckas:
There's no change to the package code?

ArnoStrouwen:
No the change to package code was already included in https://github.com/SciML/DiffEqSensitivity.jl/pull/470
Which is why I had to rebase, since I wrote this fix before that pr was completed.
Thus all that remains is writing a test.

PR #470, opened 2021-08-09
[WIP] Jacobian matrix products for ForwardSensitivity

ArnoStrouwen:


src/forward_sensitivity.jl
```diff
@@ -127,12 +135,20 @@ function ODEForwardSensitivityProblem(f::DiffEqBase.AbstractODEFunction,u0,
...
+  isautojacmat = get_jacmat(alg)
   isautojacvec = get_jacvec(alg)
   p == nothing && error("You must have parameters to use parameter sensitivity calculations!")
   uf = DiffEqBase.UJacobianWrapper(f,tspan[1],p)
   pf = DiffEqBase.ParamJacobianWrapper(f,tspan[1],copy(u0))
-
-  if isautojacvec
+  if isautojacmat
+    if alg_autodiff(alg)
+      jac_config_seed = ForwardDiff.Dual{typeof(uf)}.(u0,[ntuple(x -> 0, length(p)) for i in eachindex(u0)])
```
ArnoStrouwen, line 145:
```suggestion
      jac_config_seed = ForwardDiff.Dual{typeof(uf)}.(u0,[ntuple(x -> zero(eltype(u0)), length(p)) for i in eachindex(u0)])
```

ChrisRackauckas:
still WIP?

ArnoStrouwen:
The CI complains one of the lines I added is not covered by tests, while it definitely does get run.
Is this because the function got inlined maybe?

Other than that you can merge it, if it looks fine to you.

ChrisRackauckas:
@YingboMa we should add this to the benchmarks.

And actually, we should just transfer those repo benchmarks over to SciMLBenchmarks now that it's all open.

PR #474, opened 2021-08-16
QuadratureAdjoint() for ContinuousCallback

frankschae:
There are at least two options in `QuadratureAdjoint()` which are more difficult to support than in `BacksolveAdjoint()` 

- when one doesn't save the event time, the loop for the parameter gradients https://github.com/SciML/DiffEqSensitivity.jl/blob/b2d33c3e62fd8571e0d0028360f8534352c2af59/src/quadrature_adjoint.jl#L268 misses the necessary update. So we would need to run a check if there was an additional saved point in the adjoint which is not in `t`, push that into `t`, and check in the loop for those values. 
- terminating via `terminate!(integrator)` would also need some fixes because for `QuadratureAdjoint()` https://github.com/SciML/DiffEqSensitivity.jl/blob/b2d33c3e62fd8571e0d0028360f8534352c2af59/src/concrete_solve.jl#L128 provides a wrong time range if the full simulation time interval is not a multiple of `saveat`. (same for `InterpolatingAdjoint()`)

related issue: #374

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -178,6 +178,12 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback,DiscreteCallback,
 
         du = first(get_tmp_cache(integrator))
         λ,grad,y,dλ,dgrad,dy = split_states(du,integrator.u,integrator.t,S)
+        # if save_positions[2] = false, then the right limit is not saved. Thus, for 
+        # the QuadratureAdjoint we would need to lift y from the left to the right limit.
+        # However, one also needs to update dgrad later on.
+        if sensealg isa QuadratureAdjoint && !cb.save_positions[2]
+          w(y,y,integrator.p,integrator.t)
+        end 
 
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/pull/474/checks?check_run_id=3342289462#step:6:611

Looks like you might be relying on exact floating point ending?

frankschae:
I thought the same thing when I saw the error.... but now with the printed times:
 ``` LoadError: Endpoints do not match. Likely your time range is not a multiple of `saveat`. End times: 1.233895325143395 10.0```
it doesn't look like floating point error. (Shouldn't it be okay at the endpoint anyway, because the integrator steps to tspan[2] ..?) It's 1.233895325143395 vs 10.0.
I am wondering where the 1.23 comes from in https://github.com/SciML/DiffEqFlux.jl/blob/c8ee324b278865b35f3faf037da71ddef4aaa313/test/galacticoptim.jl#L138

ChrisRackauckas:
Nice, the message reads well.

PR #478, opened 2021-08-20
[WIP] test adjoint sensitivity explicit t dependence

ArnoStrouwen:
Test if adjoint suffers from the same issue as forward.

Also, the lack of p[4] in the jac seems weird?
Is this for testing derivatives involving the first three parameters but not the fourth?

### DIFF
--- a/test/adjoint.jl
+++ b/test/adjoint.jl
@@ -4,19 +4,19 @@ using Test
 
 function fb(du,u,p,t)
   du[1] = dx = p[1]*u[1] - p[2]*u[1]*u[2]*t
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Handling a zero derivative is hard for some AD systems (Zygote... which returns nothing instead of zero), so that's on purpose.

ChrisRackauckas:
oh wait, I see what you're saying, that's a typo. The zero parameter tests are a different spot.

ChrisRackauckas (approved):

PR #481, opened 2021-08-23
Improve checking for nothing

ArnoStrouwen:


ChrisRackauckas (approved):

devmotion (approved):

ArnoStrouwen:
That test passes locally for me.
I don't understand what the issue is.

frankschae:
yeah that's odd.. it crashes when computing the Jacobian and the vjp with FiniteDifferences. 

```
res2c = adjoint_sensitivities(sol2,DynamicSS(Rodas5()),sensealg=SteadyStateAdjoint(autodiff=false),g,nothing)
```

Can you try rerunning it to see if that was a random failure? (`res2g` passes which, however, also has `autodiff=false`)

ArnoStrouwen:
Seems to be a random failure.

PR #488, opened 2021-09-03
NILSAS

frankschae:
I have started to implement [NILSAS](https://arxiv.org/abs/1801.08674). The tests for the Lorenz system are still failing at the moment. I'll have to check if that's due to the use of `BacksolveAdjoint` as the `sensealg` for the adjoint sensitivities. 

related issue: #102


frankschae:
The stability of `BacksolveAdjoint`  wasn't the issue but the construction of `B` for the least squares problem had a bug.  The tolerance of ~10 % in the tests of the loss function depending on the parameters is a bit large.. but because of the memory consumption of LSS we can't further increase the trajectory length..

PR #493, opened 2021-09-23
Check master

ChrisRackauckas:


ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/pull/493/checks?check_run_id=3687059298#step:6:295

Something happened maybe in Zygote that makes it so that it now misses some adjoints?

@DhairyaLGandhi @mcabbott this is a pretty big master failure and I hope we can identify a fix ASAP. I'll try bounding.

mcabbott:
Error is from
https://github.com/SciML/DiffEqSensitivity.jl/blob/master/test/literal_adjoint.jl#L23
which I can't reproduce locally, as I get a segfault:
```
julia> Zygote.gradient(test,u0,p)

signal (11): Segmentation fault: 11
in expression starting at REPL[9]:1
union_isinlinable at /Applications/Julia-1.7.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.7.dylib (unknown line)
```
My guess from the stack trace is that what's new is
https://github.com/FluxML/Zygote.jl/blob/master/src/compiler/chainrules.jl#L200
which you can try removing to confirm:
```
@eval Zygote function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f_args...; kwargs...)
    y, pb = if !isempty(kwargs)
        kwf() = first(f_args)(Base.tail(f_args)...; kwargs...)
        _y, _pb = _pullback(config.context, kwf)
        _y, Δ -> first(_pb(Δ)).f_args  # `first` should be `only`
    else
        _pullback(config.context, f_args...)
    end
    ad_pullback(Δ) = zygote2differential(pb(wrap_chainrules_output(Δ)), f_args)
    return y, ad_pullback
end
```

DhairyaLGandhi:
Yeah, this doesn't look good. Which rule does this miss specifically? It might also explain https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/364 failing. Maybe we can avoid checking for an `rrule` first.

DhairyaLGandhi:
Found the issue - its `@inline wrap_chainrules_output(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))` from https://github.com/FluxML/Zygote.jl/pull/1044 removing that solves the issue, but would need to unthunking separately

ChrisRackauckas:
So will there be a hotfix patch removing it or should I upper bound?

DhairyaLGandhi:
This will need to be fixed, since otherwise we start leaking a lot of the ChainRules's types in the returned gradients.

```Julia
(Thunk(ProjectTo{AbstractArray}(element = ProjectTo{Float64}(), axes = (Base.OneTo(2),)) ∘ DiffEqSensitivity.var"#380#386"{0, ODESolution{Any, 2, Vector{Any}, Nothing, Nothing, Vector{Float64}, Nothing, ODEProblem{Vector{Float64}, Tuple{Float64, Float64}, true, Vector{Float64}, ODEFunction{true, typeof(lv!), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, ...
```

DhairyaLGandhi:
This still means there are ChainRules' types in the adjoints, so it would still error on unthunking

DhairyaLGandhi:
@ChrisRackauckas mind running the tests here with FluxML/Zygote.jl#1075 ?

ChrisRackauckas:
Seems like there are still failures

DhairyaLGandhi:
Did a little bit of sleuthing. Turns out, RecursiveArrayTools has been broken as well. https://github.com/FluxML/Zygote.jl/pull/1044 added restrictions on how to represent the gradient. For example
```Julia
julia> va = RecursiveArrayTools.VectorOfArray([rand(3,3), rand(3,3)]);

julia> gradient(va) do va
         sum(va[1])
       end
ERROR: DimensionMismatch("variable with size(x) == (3, 3, 2) cannot have a gradient with size(dx) == (2,)")
Stacktrace:
 [1] (::ChainRulesCore.ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ChainRulesCore.ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}}}})(dx::Vector{AbstractMatrix{Float64}})
   @ ChainRulesCore ~/.julia/packages/ChainRulesCore/ChM7X/src/projection.jl:219
 [2] _project
   @ ~/Downloads/mwes/diffeqsensitivity/Zygote.jl/src/compiler/chainrules.jl:141 [inlined]
 [3] map(f::typeof(Zygote._project), t::Tuple{VectorOfArray{Float64, 3, Vector{Matrix{Float64}}}}, s::Tuple{Vector{AbstractMatrix{Float64}}})
   @ Base ./tuple.jl:232
 [4] gradient(f::Function, args::VectorOfArray{Float64, 3, Vector{Matrix{Float64}}})
   @ Zygote ~/Downloads/mwes/diffeqsensitivity/Zygote.jl/src/compiler/interface.jl:77
 [5] top-level scope
   @ REPL[30]:1
```

Also, rather than initialising gradients using `zero(arg)` (specifically in the `getindex` adjoint), we now get something different due to projection

```Julia
x = Any[ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), [1.0, 0.0]]
```
which of course RAT doesn't know how to handle as the gradient of a `VectorOfArray`. Thankfully, we are still preserving the indices that we need.

Also notice that we get a type instability here, earlier, this was inferred type stable, now it isn't.

Since the array is initialised using eltype from the partial, we should be producing correct types of arrays by not using a `_zero`, and a regular `zero` instead, but that would break `OneElement`. Removing the unnecessary projection solves the issue reported in this issue, but not RAT.

mcabbott:
RecursiveArrayTools's problem is that it claims to be a 3-array of numbers, but when you iterate it, is instead a list of matrices. I think that's going to break a lot of code that believes what it says about itself:
```
julia> va = RecursiveArrayTools.VectorOfArray([rand(3,3), rand(3,3)]);

julia> va isa AbstractArray
true

julia> size(va)
(3, 3, 2)

julia> eachindex(va)
Base.OneTo(2)

julia> CartesianIndices(va)
3×3×2 CartesianIndices{3, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}}:
[:, :, 1] =
 CartesianIndex(1, 1, 1)  CartesianIndex(1, 2, 1)  CartesianIndex(1, 3, 1)
...

julia> eltype(va)
Float64

julia> first(va)
3×3 Matrix{Float64}:
 0.261312  0.429784   0.177689
 0.588208  0.548013   0.0232512
 0.965871  0.0722182  0.551001
```
Zygote could special-case it one way or another, because it's well-established. But this is pretty surprising behaviour.

ChrisRackauckas:
Nothing that breaks the iteration interface of the AbstractArray interface, so it's not surprising to the interfaces and is something code should be compatible with if it's not making untrue assumptions.

DhairyaLGandhi:
I wouldn't want to special case it in Zygote. Established or not, it obeys regular Julia semantics and implements the same array interface as any other array.

mcabbott:
So what gradients are acceptable for this object? It's a requirement that the tangent vector can be added to the original, right?
```
julia> va = RecursiveArrayTools.VectorOfArray([rand(3,4), rand(3,4)]);

julia> va + ones(size(va)) |> summary  # seems like gradient can be a 3-array of same size
"3×4×2 Array{Float64, 3}"

julia> va .+ ones(size(va)) |> summary  # the same
"3×4×2 Array{Float64, 3}"

julia> y, bk = pullback(va -> sum(va[1]), va); 

julia> summary(y)
"Float64"

julia> bk(1.0)[1] |> summary  # Zygote produces a vector of matrices
"2-element Vector{AbstractMatrix{Float64}}"

julia> va + bk(1.0)  # adding that to original is an error
ERROR: MethodError: no method matching +(::VectorOfArray{Float64, 3, Vector{Matrix{Float64}}}, ::Tuple{Vector{AbstractMatrix{Float64}}})

julia> va .+ bk(1.0)  # maybe .+ ? different error.
ERROR: MethodError: no method matching +(::Float64, ::Vector{AbstractMatrix{Float64}})
```
So what Zygote was producing seems wrong. Must it produce a Tangent? What consumes this, if not `+`?

ChrisRackauckas:
What's your suggestion on what I should do then to make it not error? I would've assumed ProjectTo with just the first two overloads would be correct.

mcabbott:
> assumed ProjectTo with just the first two overloads 

This will break 3-array gradients like `gradient((x,y) -> sum(x ./ y), va, rand(3))`. And won't accept the VofA gradient producded by the `(::ProjectTo{VectorOfArray})(dx::AbstractVector{<:AbstractArray})` method.

Whether what's written will "make it not error" I don't know. This is a real question:

> What consumes this, if not +?

i.e. what happens downstream? Did it previously handle both 3-array and vector of array gradients, or only the latter? Does it just use iteration or will it expect indexing / broadcasting to work exactly like a vector of matrices?

ChrisRackauckas:
It broadcasts like an Array under the interpretation of its multidimensional indexing. It just uses the AbstractArray interface fallbacks for all of that.

mcabbott:
Sure. What I mean is, what broadcasting behaviour is expected _of the gradient_? Changing a vector of arrays into a VectorOfArrays (as the above projection does) will change this.

ChrisRackauckas:
It should broadcast like a VectorOfArray

PR #498, opened 2021-10-05
fuse adjoint - with =

ChrisRackauckas:
This should both make things a little bit faster while also fixing a GPU issue.


ChrisRackauckas:
Looks like this is a Zygote issue? https://github.com/SciML/DiffEqSensitivity.jl/pull/498/checks?check_run_id=3806840253#step:6:381

@DhairyaLGandhi

PR #507, opened 2021-11-01
Fix: swap u and p in TrackedAffect

frankschae:
The main error:
`MethodError: no method matching Vector{Float32}(::Matrix{Float32})`
in https://github.com/SciML/DiffEqSensitivity.jl/issues/503 is due to a swap between `typeof(u)` and `typeof(p)`. 

There are follow up errors like: 
`type FakeIntegrator has no field tprev`

@ChrisRackauckas Should we copy `tprev` to the fakeintegrator as well (or is there a better way to solve it)?

### DIFF
--- a/src/sensitivity_interface.jl
+++ b/src/sensitivity_interface.jl
@@ -14,9 +14,9 @@ function _adjoint_sensitivities(sol,sensealg,alg,g,t=nothing,dg=nothing;
                                    kwargs...)
 
   if sol.prob isa ODEProblem
-    adj_prob = ODEAdjointProblem(sol,sensealg,g,t,dg,checkpoints=checkpoints,
+    adj_prob = ODEAdjointProblem(sol,sensealg,g,t,dg; checkpoints=checkpoints,
                                  callback = callback,
-                                 abstol=abstol,reltol=reltol)
+                                 abstol=abstol,reltol=reltol, kwargs...)
 
   elseif sol.prob isa SDEProblem
     adj_prob = SDEAdjointProblem(sol,sensealg,g,t,dg,checkpoints=checkpoints,
--- a/test/callbacks.jl
+++ b/test/callbacks.jl
@@ -482,7 +482,7 @@ end
           condition(u,t,integrator) = t == 5
           affect!(integrator) = (@show integrator.tprev; integrator.u[1] += integrator.t-integrator.tprev)
           cb = DiscreteCallback(condition,affect!)
-          tstops=[5.0]
+          tstops = [4.999, 5.0]
           test_discrete_callback(cb,tstops,g,dg!,nothing,true)
         end
       end
@@ -544,7 +544,7 @@ end
           condition(u,t,integrator) = t == 5
           affect!(integrator) = (@show integrator.tprev; integrator.u[1] += integrator.t-integrator.tprev)
           cb = DiscreteCallback(condition,affect!)
-          tstops=[5.0]
+          tstops = [4.999, 5.0]
           test_discrete_callback(cb,tstops,g,dg!,nothing,true)
         end
       end
--- a/test/callbacks.jl
+++ b/test/callbacks.jl
@@ -5,7 +5,7 @@ abstol=1e-12
 reltol=1e-12
 savingtimes=0.5
 
-function test_discrete_callback(cb, tstops, g, dg!, cboop=nothing)
+function test_discrete_callback(cb, tstops, g, dg!, cboop=nothing, tprev=false)
   function fiip(du,u,p,t)
     du[1] = dx = p[1]*u[1] - p[2]*u[1]*u[2]
     du[2] = dy = -p[3]*u[2] + p[4]*u[1]*u[2]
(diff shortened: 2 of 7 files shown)


ChrisRackauckas:
>@ChrisRackauckas Should we copy tprev to the fakeintegrator as well (or is there a better wat to solve it)?

Yeah we should.

ChrisRackauckas:
Rebased.

frankschae:
Nice, thanks! Status here is that it almost works, but the example from issue https://github.com/SciML/DiffEqSensitivity.jl/issues/503 has some remaining problems (didn't look closely, but I think due to mutation)

ChrisRackauckas:
Rebased.

ChrisRackauckas:
Looks like those are real test failures since master tests are passing.

frankschae:
yup, agreed. I'll run some tests locally to see what's going on.

frankschae:
I am surprised that the stepping behaviour seems to be different for mutating vs. non-mutationg ODEProblems:
```julia
condition(u,t,integrator) = t == 5
affect!(integrator) = (@show integrator.tprev; integrator.u[1] += integrator.t-integrator.tprev)
cb = DiscreteCallback(condition,affect!)
tstops=[5.0]
function fiip(du,u,p,t)
end

function foop(u,p,t)
  dx = p[1]*u[1] - p[2]*u[1]*u[2]
  dy = -p[3]*u[2] + p[4]*u[1]*u[2]
  [dx,dy]
end

sol1 = solve(prob,Tsit5(),u0=u0,p=p,callback=cb,tstops=tstops,abstol=abstol,reltol=reltol,saveat=savingtimes)
sol2 = solve(proboop,Tsit5(),u0=u0,p=p,callback=cb,tstops=tstops,abstol=abstol,reltol=reltol,saveat=savingtimes)
```
```
integrator.tprev = 4.998981564680802 (for sol1)
integrator.tprev = 4.998924417947237 (for sol2)
```

ChrisRackauckas:
The difference could be just that in-place forces simd ivdep

test/downstream/HybridNODE.jl
```diff
@@ -115,6 +115,9 @@ end
 mutable struct Affect{T}
       callback_data::T
 end
+function (cb::Affect)(integrator)
+    integrator.u .= integrator.u .+ cb.callback_data[:,Int(integrator.t÷60+1),1] * (integrator.t-integrator.tprev)
```
ChrisRackauckas, line 119: `@view(cb.callback_data[:,Int(integrator.t÷60+1),1])`?

ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/runs/5044152833?check_suite_focus=true#step:6:780 looks like a real test failure?

test/downstream/HybridNODE.jl
```diff
@@ -112,6 +113,85 @@ function test_hybridNODE2(sensealg)
     println("  ")
 end
 
+mutable struct Affect{T}
+      callback_data::T
+end
+compute_index(t) = Int(t ÷ 1 + 1)
+Zygote.@nograd compute_index #avoid taking grads of this function
+function (cb::Affect)(integrator)
+    indx = compute_index(integrator.t)
```
frankschae, line 122: The test failure is related to the index assignment (`t ÷ 1`). I had hoped that `nograd` could handle this. 
```
MethodError: no method matching div(::ReverseDiff.TrackedReal{Float32, Float32, ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}}, ::ReverseDiff.TrackedReal{Float32, Float32, ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}}, ::RoundingMode{:ToZero})
```

test/callbacks.jl
```diff
@@ -466,6 +478,13 @@ end
           tstops=[5.1]
           test_discrete_callback(cb,tstops,g,dg!,cboop)
         end
+        @testset "tprev dependent callback" begin
+          condition(u,t,integrator) = t == 5
+          affect!(integrator) = (@show integrator.tprev; integrator.u[1] += integrator.t-integrator.tprev)
+          cb = DiscreteCallback(condition,affect!)
+          tstops = [4.999, 5.0]
```
frankschae, line 485: The first `tstop` value here forces the integrator to step to 4.999. This allows for a unified stepping behavior.

src/sensitivity_interface.jl
```diff
@@ -14,9 +14,9 @@ function _adjoint_sensitivities(sol,sensealg,alg,g,t=nothing,dg=nothing;
                                    kwargs...)
 
   if sol.prob isa ODEProblem
-    adj_prob = ODEAdjointProblem(sol,sensealg,g,t,dg,checkpoints=checkpoints,
+    adj_prob = ODEAdjointProblem(sol,sensealg,g,t,dg; checkpoints=checkpoints,
                                  callback = callback,
-                                 abstol=abstol,reltol=reltol)
+                                 abstol=abstol,reltol=reltol, kwargs...)
```
frankschae, line 19: passing `kwargs` to the `ODEAdjointProblem` seems to be necessary to force the stepping also in the checkpointed interpolating adjoint. Otherwise, `tstops` that do not save anything are dropped.

test/downstream/HybridNODE.jl
ChrisRackauckas: `round(Int,t)+1`?

PR #509, opened 2021-11-16
Try moving to `juliaecosystem`

staticfloat:
We don't have GPU runners on the sandbox runners yet, but we may eventually get them.  Until then, we can continue to use `juliagpu`.


staticfloat:
Well, the buildkite build itself worked....

PR #511, opened 2021-12-02
Fix the enzyme default check

ChrisRackauckas:
```julia
using OrdinaryDiffEq, Zygote, DiffEqSensitivity
function fiip(du,u,p,t)
end
p = [1.5,1.0,3.0,1.0]; u0 = [1.0; 1.0]
prob = ODEProblem(fiip,complex(u0),(0.0,10.0),complex(p))

function sum_of_solution(u0, p)
    _prob = remake(prob,u0=u0,p=p)
    real(sum(solve(_prob,Tsit5(),reltol=1e-6,abstol=1e-6,saveat=0.1)))
end

dx = Zygote.gradient(sum_of_solution, complex(u0), complex(p))
```

now works.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: 1
+          version: 1.6
       - uses: actions/cache@v1
         env:
           cache-name: cache-artifacts
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -29,7 +29,7 @@ function DiffEqBase._concrete_solve_adjoint(prob::Union{ODEProblem,SDEProblem},
         try
           Enzyme.autodiff(Enzyme.Duplicated(du, du),
                           u0,p,prob.tspan[1]) do out,u,_p,t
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
It looks like Zygote on v1.7 is missing derivatives for gemv! coming from some higher level vector math call? @maleadt @DhairyaLGandhi I'm going to merge and ignore this but that's a little odd.

https://buildkite.com/julialang/diffeqsensitivity-dot-jl/builds/510#ac82ea28-d297-45b5-bb5f-01f490e5759d/668-1284

DhairyaLGandhi:
Seems to be a matrix-vector-scalar call somewhere in cublas. @maleadt is there a change in the code path for CUDA in 1.7?

maleadt:
Nothing comes to mind, no.

devmotion:
The seemingly same error showed up also in ApproximateGPs (https://github.com/JuliaGaussianProcesses/ApproximateGPs.jl/runs/4385202539?check_suite_focus=true#step:5:31, the corresponding issue is https://github.com/JuliaGaussianProcesses/ApproximateGPs.jl/issues/82). Since the package and its tutorials don't use CUDA, it seems to be a more general BLAS/Zygote issue.

PR #513, opened 2021-12-04
ignore branching in `Base.OneTo` constructor

simeonschaub:
In 1.7, a check was added to the `Base.OneTo` constructor. This should
fix https://github.com/SciML/DiffEqFlux.jl/pull/659#issuecomment-986041454.


ChrisRackauckas:
Thanks!

PR #519, opened 2021-12-13
[WIP] Optimize Steady State Adjoint

avik-pal:
Currently, VecJac operator repeatedly call the forward pass with the same arguments.
(I know this is not the right place to keep the operators and stuff. But wanted to share a temporary solution which should speed things up for big neural networks)

### DIFF
--- a/src/steadystate_adjoint.jl
+++ b/src/steadystate_adjoint.jl
@@ -1,14 +1,7 @@
-struct SteadyStateAdjointSensitivityFunction{
-    C<:AdjointDiffCache,
-    Alg<:SteadyStateAdjoint,
-    uType,
-    SType,
-    fType<:ODEFunction,
-    CV,
-    λType,
-    VJPType,
-    LS,
-} <: SensitivityFunction
+include("operator.jl")
+
+struct SteadyStateAdjointSensitivityFunction{C<:AdjointDiffCache,Alg<:SteadyStateAdjoint,uType,SType,fType<:ODEFunction,
+                                             CV,λType,VJPType,LS} <: SensitivityFunction
     diffcache::C
     sensealg::Alg
     discrete::Bool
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
We should define an operator in this package with all of the vjp choices.

src/steadystate_adjoint.jl
```diff
@@ -21,88 +14,38 @@ struct SteadyStateAdjointSensitivityFunction{
...
-        sol,
-        dg,
-        f;
-        quad = false,
-        needs_jac = needs_jac,
-    )
+    diffcache, y = adjointdiffcache(g, sensealg, discrete, sol, dg, f; quad=false, needs_jac=needs_jac)
 
     λ = zero(y)
     linsolve = needs_jac ? nothing : sensealg.linsolve
```
ChrisRackauckas, line 23: it should always just use `sensealg.linsolve` and default that to `nothing`.

src/steadystate_adjoint.jl
```diff
@@ -21,88 +14,38 @@ struct SteadyStateAdjointSensitivityFunction{
...
-        discrete,
-        sol,
-        dg,
-        f.colorvec,
-        needs_jac,
-    )
+    sense = SteadyStateAdjointSensitivityFunction(g, sensealg, discrete, sol, dg, f.colorvec, needs_jac)
     @unpack diffcache, y, sol, λ, vjp, linsolve = sense
 
     if needs_jac
```
ChrisRackauckas, line 43: `needs_jac = linsolve !== nothing && needs_concrete_A(linsolve)` using https://github.com/SciML/LinearSolve.jl/pull/82

avik-pal:
Not modifying this branch for the sake of the paper. New PR #542

PR #527, opened 2021-12-22
only hold back DiffEqBase

ChrisRackauckas:


ChrisRackauckas:
Hmm, so the change was just a tiny floating point change.

https://github.com/SciML/DiffEqBase.jl/pull/689

is the test just unstable @frankschae ?

frankschae:
The shadowing tests from https://github.com/SciML/DiffEqSensitivity.jl/pull/525? The LSS test looks good (at least it makes perfect sense, because there were indeed significant fluctuations). The NILSS test is a bit of a stretch - although I recall similar things happen with shorter trajectories but very randomly.

ChrisRackauckas:
How do we get tests passing again then. This should only be changing like the 14th decimal place 😅, but chaos is chaotic.

frankschae:
I fixed the seed for the original trajectory. Then, when I saw stronger fluctuations, I used a longer time span and a larger number of segments and steps per segment (that worked very reliable once the original trajectory had been fixed).

PR #537, opened 2022-01-09
Change non existing du to Jv

ArnoStrouwen:
What is du?
Is this not tested?


ChrisRackauckas:
Interesting. I would've thought the default would be to jacvec with autodiff?

ArnoStrouwen:
The names of the different methods are not always informative.

autodiff = true with autojacvec = true calculates a directional derivative using AD without explicitly forming the jac
autodiff = false with autojacvec = true calculates a directional derivative using finite differences without explicitly forming the jac

ChrisRackauckas:
Yeah... we'll need to finish https://github.com/JuliaDiff/AbstractDifferentiation.jl and then change the names to something more informative 😅

PR #542, opened 2022-01-14
[WIP] Optimize Steady State Adjoint

avik-pal:


ChrisRackauckas:
Shouldn't the operator just be defined via `mul!` to use the `vecjacobian!` function? That would then make all of the `sensealg` choices work?

avik-pal:
The `vecjacobian!` function would run the forward pass each time `mul!` is called. This one does 1 forward pass and `n` backward passes.

ChrisRackauckas:
I wonder if there's an effective way to reuse the code. New kwarg for the forward?

Issue #543, opened 2022-01-15
Sign of the gradient w.r.t. initial condition


ilysym:
`du0, dp = adjoint_sensitivities(sol, alg, dg, ts)`
it seems that `du0` has wrong sign here: `du0 == -du0_true`.
There is a test for `du0`, but is uses wrong `dg`:
https://github.com/SciML/DiffEqSensitivity.jl/blob/547104399ed173d3a6e077714090bc81235e3044/test/adjoint.jl#L41
`dg(u) = 2.0.-u` while it should be `dg(u) = u.-1.0`.

ChrisRackauckas:
Yeah I think @frankschae and @YingboMa noticed this too. We end up correcting the sign in `concrete_solve`, so people who use AD for the derivatives (99.99% of users) will get the right gradient, but it is still odd. I think we should probably just correct this with a breaking release ASAP.

ChrisRackauckas:
The signs were corrected in the update to SciMLSensitivity.jl

Issue #544, opened 2021-11-22
Minimisation of a mixed-type loss function


nhcho91:
Hi. I am opening a new issue following the response to my question in [here](https://discourse.julialang.org/t/minimisation-of-a-bolza-form-loss-function-in-diffeqflux-framework/71920). For your convenience, I copy-pasted my original question below:

# Brief Summary
* Problem at hand:
minimise $J \left( p  \right) = \phi \left( x \left(t_{f} \right) \right) + \int_{t_{0}}^{t_{f}} L \left( \tau, x\left(\tau\right), p \right) d\tau + R\left( p \right)$
subject to $\dot{x}\left(t\right) = f\left( t, x\left(t\right), p\right)$
with $x\left(t_{0}\right) = x_{0}$, $t_{0}$, $t_{f}$  fixed,
where $p$ is the parameter for a neural network.

* How can I compute the correct gradient for the above Bolza form loss function (= continuous functional (Lagrange term) + terminal cost (Mayer term) + neural network parameter-dependent function ) through continuous adjoint sensitivity analysis using DiffEqSensitivity or DiffEqFlux?

* Is there any recommendation about the way of constructing the Bolza form loss function, e.g., including $\dot{J}_{cont}\left( t \right) = L\left( t, x\left(t\right), p \right)$ in the forward pass?

# More Details
Hello. My name is Namhoon Cho.

I was trying to solve a nonlinear deterministic optimal control problem in continuous-time domain by optimising a neural feedback policy. The context of my application requires specifying the terminal cost to be a function of final state in the form $\phi \left( x \left(t_{f}\right) \right)$ besides the continuous functional.

I have been following the codebase developed for models combining ODE and NNs.
At the first glance, I tried to just specify the loss function in such mixed form and apply DiffEqFlux.sciml_train (or others) for training.

[...]

PR #545, opened 2022-01-16
swap signs adjoint

ArnoStrouwen:
Looking into https://github.com/SciML/DiffEqSensitivity.jl/issues/543
Example for once branch, all the other branches need to have their signs swapped also.

I think the issue is that you can either define the adjoint equation as:
`d lambda = -A*lambda -B`
or reparametrized as `gamma = -lambda `
`d gamma = -A*gamma +B`
and the code uses both versions (but calls everything lambda), the signs in `d grad` depend  on which version you use, for continuous cost functions gamma was assumed and was also correctly passed, while for discrete cost functions lambda was assumed, but gamma was passed. I suggest switching everything to assuming lambda (because that is how I see it done in optimal control books). If there is a hack in concrete solve that fixes this issue it will have to be removed, but I don't know where that is located.

```jl
using DiffEqSensitivity
using OrdinaryDiffEq
using ForwardDiff
function fiip(du,u,p,t)
end

function cost_forward(input)
    u0 = input[1:2]
    p = input[3:end]

    sol = solve(prob,Tsit5(),abstol=1e-14,reltol=1e-14)
    cost = zero(eltype(p))
    for t = 0:0.01:10.0
       cost += (sol(t)[1]^2 + p[1])*0.01
    end
    cost
end
p = [1.5,1.0,3.0,1.0];
u0 = [1.0;1.0];
input = vcat(u0,p)

ForwardDiff.gradient(cost_forward,input)
#= 4-element Vector{Float64}:
  -71.92545923275867
  -18.013666752777436
  -19.780018103016378
  -18.01366675277761
   91.50591475972585
 -330.49918845227745
 =#

sol = solve(prob,Tsit5(),reltol=1e-14,abstol=1e-14)
g(u,p,t) = u[1]^2 + p[1]
function dgdu(out,u,p,t)
  out[1] = 2u[1]
  out[2] = 0.0
end
function dgdp(out,u,p,t)
  out[1] = 1.0
  out[2] = 0.0
  out[2] = 0.0
  out[2] = 0.0
end

du0,dp = adjoint_sensitivities(sol,Tsit5(),g,nothing,(dgdu,dgdp),sensealg=QuadratureAdjoint(),abstol=1e-14,reltol=1e-14)
#([-71.95379975016145, -18.015832635730415], [-19.809511715740577 -18.01613240466981 91.49995776597893 -330.5063153397706])


function cost_forward(input)
  u0 = input[1:2]
  p = input[3:end]

  sol = solve(prob,Tsit5(),abstol=1e-14,reltol=1e-14)
  cost = zero(eltype(p))
  for t = 1.0:9.0
     cost += sol(t)[1]^2 #+ p[1]
  end
  cost
end
ForwardDiff.gradient(cost_forward,input)
# with p[1]
#= 6-element Vector{Float64}:
  -45.75266688665985
  -21.217655577397284
   40.388411700615876
  -21.217655577397277
  101.53385018713378
 -302.9410227368751
 =#
# without p[1]
#=  6-element Vector{Float64}:
 -45.75266688665985
 -21.217655577397284
  31.388411700615876
 -21.217655577397277
 101.53385018713378
-302.9410227368751
 =#

g(u,p,t) = u[1]^2 #+ p[1]
function dgdu(out,u,p,t,i)
  out[1] = 2u[1]
  out[2] = 0.0
end
function dgdp(out,u,p,t,i) # don't know how to pass this so can't be sure if more sign changes are needed for this
  out[1] = 1.0
  out[2] = 0.0
  out[2] = 0.0
  out[2] = 0.0
end

du0,dp = adjoint_sensitivities(sol,Tsit5(),dgdu,1.0:9.0,abstol=1e-3,reltol=1e-3,sensealg=QuadratureAdjoint())
# ([-45.43427809297227, -21.209701202168834], [31.885996129674837 -21.176751357966683 101.5737226156943 -302.5700510411537])
```

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -283,7 +283,7 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::Bool, dgrad, dy, W
         end
       end
     end
-    mul!(dgrad',λ',-pJ)
+    mul!(dgrad',λ',pJ)
   end
   if dy !== nothing
     if W===nothing
@@ -323,10 +323,10 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::TrackerVJP, dgrad,
 
     # Grab values from `_dy` before `back` in case mutated
     dy !== nothing && (dy[:] .= vec(Tracker.data(_dy)))
-    
+
     tmp1, tmp2 = Tracker.data.(back(λ))
     dλ[:] .= vec(tmp1)
-    dgrad !== nothing && (dgrad[:] .= -vec(tmp2))
+    dgrad !== nothing && (dgrad[:] .= vec(tmp2))
   else
     if W===nothing
       _dy, back = Tracker.forward(y, p) do u, p
@@ -343,7 +343,7 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::TrackerVJP, dgrad,
 
     tmp1, tmp2 = Tracker.data.(back(λ))
     dλ[:] .= vec(tmp1)
-    dgrad !== nothing && (dgrad[:] .= -vec(tmp2))
+    dgrad !== nothing && (dgrad[:] .= vec(tmp2))
   end
   return
 end
@@ -419,7 +419,7 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::ReverseDiffVJP, dg
   ReverseDiff.increment_deriv!(output, λ)
   ReverseDiff.reverse_pass!(tape)
   copyto!(vec(dλ), ReverseDiff.deriv(tu))
-  dgrad !== nothing && copyto!(vec(dgrad), -ReverseDiff.deriv(tp))
+  dgrad !== nothing && copyto!(vec(dgrad), ReverseDiff.deriv(tp))
   ReverseDiff.pull_value!(output)
   dy !== nothing && copyto!(vec(dy), ReverseDiff.value(output))
   return
@@ -450,7 +450,7 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::ZygoteVJP, dgrad, 
 
     tmp1,tmp2 = back(λ)
     dλ[:] .= vec(tmp1)
-    dgrad !== nothing && tmp2 !== nothing && (dgrad[:] .= -vec(tmp2))
+    dgrad !== nothing && tmp2 !== nothing && (dgrad[:] .= vec(tmp2))
   else
     if W===nothing
       _dy, back = Zygote.pullback(y, p) do u, p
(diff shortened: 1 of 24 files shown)


ChrisRackauckas:
This will fail unless the concrete_solve.jl files are also updated, since all of the AD overloads are designed to be correct under the different choice of sign from `adjoint_sensitivities`.

This is a breaking change and so it will need a major bump. We should do it at the same time as the fix for https://github.com/SciML/DiffEqSensitivity.jl/issues/544 since both focus on non-AD usage of adjoints.

frankschae:
To Do:

- [ ] Swap all signs + fix all tests (including callbacks and shadowing methods).
- [ ] change signature of `adjoint_sensitivity` to support mixed costs. Do we want: `adjoint_sensitivity(sol, alg, ts=nothing, dg_discrete=nothing, dg_continuous=nothing, g=nothing)`?

ChrisRackauckas:
> Do we want: adjoint_sensitivity(sol, alg, ts=nothing, dg_discrete=nothing, dg_continuous=nothing, g=nothing)?

I think that's good. I don't think this should be the same PR though. We should do it as part of the same breaking change, but one after the other.

frankschae:
yup, I agree (I only mentioned it again because of the name of the test set - I think having a comprehensive set of tests that are sensitive to the sign change is good, and that one can then be updated to reflect the new signature + mixed contributions).

src/derivative_wrappers.jl
```diff
@@ -574,12 +572,12 @@ function _jacNoise!(λ, y, p, t, S::TS, isnoise::Bool, dgrad, dλ, dy) where TS<
 
     if StochasticDiffEq.is_diagonal_noise(prob)
       pJt = transpose(λ).*transpose(pJ)
-      dgrad[:] .= vec(pJt)
+      dgrad[:] .= -vec(pJt)
```
frankschae, line 575: Instead of these changes, we could also have a 

```julia
dgrad .*= -1
```
similar to the 
```julia
dλ .*= -1
```
within `BacksolveAdjoint` and `InterpolatingAdjoint`. Might actually be easier to read along the typical definition given in https://github.com/SciML/DiffEqSensitivity.jl/pull/545#issue-1105053984.

ChrisRackauckas: `-vec` allocates, so we shouldn't do that.

frankschae: oh yes. Then the other option is definitely better. (I guess `.-vec` could be fine but that's definitely harder to read I think)

ChrisRackauckas:
Yeah, that last commit is much less "intrusive" and makes it match the derivation much more closely.

src/steadystate_adjoint.jl
```diff
@@ -176,9 +176,9 @@ end
         dg_dp_config = build_grad_config(sensealg, dg_dp, p, p)
         gradient!(dg_dp_val, dg_dp, p, sensealg, dg_dp_config)
 
-        @. dg_dp_val = dg_dp_val + vjp
+        @. dg_dp_val = dg_dp_val - vjp
         return dg_dp_val
     else
-        return vjp
+        return -vjp
```
ChrisRackauckas, line 182: change this to non-allocating by `vjp .*= -1`

ChrisRackauckas:
There was a patch release. https://github.com/EnzymeAD/Enzyme.jl/releases/tag/v0.10.1

frankschae:
oh wait, it must be related to `ReverseDiffVJP` as we didn't fix the `@view` issue https://github.com/SciML/DiffEqSensitivity.jl/issues/630.

frankschae:
Core5 is solved by https://github.com/SciML/DiffEqSensitivity.jl/pull/644.

frankschae:
LGTM once the non-constant types error messages are fixed.

ChrisRackauckas:
Those are fixed on master, so merging 🎉

PR #553, opened 2022-01-24
consolidate automated vjp choice and throw more warnings

ChrisRackauckas:
Currently builds on https://github.com/SciML/DiffEqSensitivity.jl/pull/551.

The idea is:

- Default autojacvec to `nothing`, remove all weird extra `autojacvec = true` handling
- If autojacvec was originally nothing, try a VJP, and fallback to numerical if that fails (with a warning)

This should try better VJPs more often with less code and be more clear to the user, with more warnings and fallbacks to accidentally be good enough more of the time.

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -3,7 +3,7 @@
 # Here is where we can add a default algorithm for computing sensitivities
 # Based on problem information!
 
-function inplace_vjp(prob,u0,p,verbose)
+function inplace_vjp(prob,u0,p,has_cb,verbose)
   du = copy(u0)
   ez = try
     Enzyme.autodiff(Enzyme.Duplicated(du, du),
@@ -15,24 +15,28 @@ function inplace_vjp(prob,u0,p,verbose)
   catch
     false
   end
-  if ez
+  if ez && !has_cb
     return EnzymeVJP()
   end
 
   # Determine if we can compile ReverseDiff
-  compile = try
-      if DiffEqBase.isinplace(prob)
-        !hasbranching(prob.f,copy(u0),u0,p,prob.tspan[1])
-      else
-        !hasbranching(prob.f,u0,p,prob.tspan[1])
-      end
-  catch
+  compile = if has_cb
       false
-  end
+    else
+      try
+        if DiffEqBase.isinplace(prob)
+          !hasbranching(prob.f,copy(u0),u0,p,prob.tspan[1])
+        else
+          !hasbranching(prob.f,u0,p,prob.tspan[1])
+        end
+      catch
+        false
+      end
+    end
   vjp = try
-    ReverseDiff.GradientTape((_y, _p, [t])) do u,p,t
+    ReverseDiff.GradientTape((copy(u0), p, [prob.tspan[1]])) do u,p,t
       du1 = similar(u, size(u))
-      f(du1,u,p,first(t))
+      prob.f(du1,u,p,first(t))
       return vec(du1)
     end
     ReverseDiffVJP(compile)
@@ -106,7 +110,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
                                  saveat = eltype(prob.tspan)[],
                                  save_idxs = nothing,
                                  kwargs...)
- 
+
   if !(typeof(p) <: Union{Nothing,SciMLBase.NullParameters,AbstractArray}) || (p isa AbstractArray && !Base.isconcretetype(eltype(p)))
     throw(AdjointSensitivityParameterCompatibilityError())
   end
@@ -278,7 +282,7 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg, sensealg::AbstractForward
                                             u0, p, args...;
                                             save_idxs=nothing,
                                             kwargs...)
-  
+
   if !(typeof(p) <: Union{Nothing,SciMLBase.NullParameters,AbstractArray}) || (p isa AbstractArray && !Base.isconcretetype(eltype(p)))
     throw(ForwardSensitivityParameterCompatibilityError())
   end
@@ -335,7 +339,7 @@ function DiffEqBase._concrete_solve_forward(prob,alg,
    out,_concrete_solve_pushforward
 end
 
-const FORWARDDIFF_SENSITIVITY_PARAMETER_COMPATABILITY_MESSAGE = 
+const FORWARDDIFF_SENSITIVITY_PARAMETER_COMPATABILITY_MESSAGE =
 """
 ForwardDiffSensitivity assumes the `AbstractArray` interface for `p`. Thus while
 DifferentialEquations.jl can support any parameter struct type, usage
@@ -343,8 +347,8 @@ with ForwardDiffSensitivity requires that `p` could be a valid
 type for being the initial condition `u0` of an array. This means that
 many simple types, such as `Tuple`s and `NamedTuple`s, will work as
 parameters in normal contexts but will fail during ForwardDiffSensitivity
-construction. To work around this issue for complicated cases like nested structs, 
-look into defining `p` using `AbstractArray` libraries such as RecursiveArrayTools.jl 
+construction. To work around this issue for complicated cases like nested structs,
+look into defining `p` using `AbstractArray` libraries such as RecursiveArrayTools.jl
 or ComponentArrays.jl.
 """
 
@@ -359,7 +363,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
                                  sensealg::ForwardDiffSensitivity{CS,CTS},
                                  u0,p,args...;saveat=eltype(prob.tspan)[],
                                  kwargs...) where {CS,CTS}
-  
+
   if !(typeof(p) <: Union{Nothing,SciMLBase.NullParameters,AbstractArray}) || (p isa AbstractArray && !Base.isconcretetype(eltype(p)))
     throw(ForwardDiffSensitivityParameterCompatibilityError())
   end
@@ -690,7 +694,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,sensealg::TrackerAdjoint,
   DiffEqBase.sensitivity_solution(sol,u,Tracker.data.(sol.t)),tracker_adjoint_backpass
 end
 
-const REVERSEDIFF_ADJOINT_GPU_COMPATABILITY_MESSAGE = 
+const REVERSEDIFF_ADJOINT_GPU_COMPATABILITY_MESSAGE =
 """
 ReverseDiffAdjoint is not compatible GPU-based array types. Use a different
 sensitivity analysis method, like InterpolatingAdjoint or TrackerAdjoint,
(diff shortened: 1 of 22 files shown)


ChrisRackauckas:
This PR needs Enzyme segfaults to be fixed before it can be viable. Somewhat like https://github.com/wsmoses/Enzyme.jl/issues/175 but I think caused by a view?

ChrisRackauckas:
@frankschae can you look into helping finish this one?

frankschae:
hmm... `InterpolatingAdjoint()`, which uses `ReverseDiffVJP(false)` here, and `ForwardDiffSensitivity()` fail for the first set of chunking tests.  

```julia
using DiffEqSensitivity, OrdinaryDiffEq, Zygote, Test, ForwardDiff

function fiip(du,u,p,t)
  du[1] = dx = p[1]*u[1] - p[2]*p[51]*p[75]*u[1]*u[2]
  du[2] = dy = -p[3]*p[81]*p[25]*u[2] + (sum(@view(p[4:end]))/100)*u[1]*u[2]
end
function foop(u,p,t)
  dx = p[1]*u[1] - p[2]*p[51]*p[75]*u[1]*u[2]
  dy = -p[3]*p[81]*p[25]*u[2] + (sum(@view(p[4:end]))/100)*p[4]*u[1]*u[2]
  [dx,dy]
end

p = reshape(vcat(p,ones(100)),4,26)

loss = (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,abstol=1e-14,reltol=1e-14,saveat=0.1,sensealg=ForwardDiffSensitivity()))
@time du01,dp1 = Zygote.gradient(loss,u0,p)

loss = (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,abstol=1e-14,reltol=1e-14,saveat=0.1,sensealg=InterpolatingAdjoint()))
@time du02,dp2 = Zygote.gradient(loss,u0,p)

loss = (u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,abstol=1e-14,reltol=1e-14,saveat=0.1,sensealg=ForwardDiffSensitivity(chunk_size=104)))
@time du03,dp3 = Zygote.gradient(loss,u0,p)

dp = ForwardDiff.gradient(p->loss(u0,p),p)
du0 = ForwardDiff.gradient(u0->loss(u0,p),u0)

@test du01 ≈ du0 rtol=1e-12
@test du01 ≈ du02 rtol=1e-12
@test du01 ≈ du03 rtol=1e-12
@test dp1 ≈ dp rtol=1e-12
@test dp1 ≈ dp2 rtol=1e-12
@test dp1 ≈ dp3 rtol=1e-12
```

frankschae:
oh yeah ...
```julia
loss = (u0, p) -> sum(solve(prob, Tsit5(), u0=u0, p=p, abstol=1e-14, reltol=1e-14, saveat=0.1, sensealg=InterpolatingAdjoint()))
@time du1, dp1 = Zygote.gradient(loss, u0, p)
@time du2, dp2 = Zygote.gradient(loss, u0, p) # are not the same
@show p # p is mutated
```
```
p = [1.5 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 0.96136454502692 ..... 0.96136454502692 ]
```

frankschae:
I am not sure if we want to keep the additional `noise=` kwarg to `BacksolveAdjoint` and `InterpolatingAdjoint`. Although I can think of many cases where the user would like to use a different AD backend for the drift and diffusion functions, I don`t think the current interface is flexible enough (since the parameters are shared and since both functions are always equal regarding inplace/out-of-place, there isn`t a good way to make an individual decision).

ChrisRackauckas:
I agree. It might as well just match autojacvec.

src/concrete_solve.jl
```diff
@@ -3,45 +3,58 @@
...
+                    copy(u0),copy(p),prob.tspan[1]) do out,u,_p,t
       prob.f(out, u, _p, t)
       nothing
     end
     true
   catch
     false
   end
-  if ez
+  if ez && !has_cb
```
ChrisRackauckas, line 18: what's the issue with callbacks here?

src/concrete_solve.jl
```diff
@@ -3,45 +3,58 @@
 # Here is where we can add a default algorithm for computing sensitivities
 # Based on problem information!
 
-function inplace_vjp(prob,u0,p)
+function inplace_vjp(prob,u0,p,has_cb,verbose)
   du = copy(u0)
   ez = try
```
ChrisRackauckas, line 8: if we know it has issues with callbacks, then we should do `!has_cb` checks before we try it.

src/sensitivity_algorithms.jl
```diff
@@ -768,34 +754,33 @@ Base.@pure function NILSS(nseg, nstep; nus=nothing, rng=Xorshifts.Xoroshiro128Pl
...
+each time step and thus should generally be preferred (for large system sizes) over `AdjointLSS`.
+`NILSAS` is favourable over `NILSS` for many parameters because NILSAS computes the gradient
+with respect to multiple parameters with negligible additional cost.
 
 ## Constructor
 
 ```julia
 NILSAS(nseg, nstep, M=nothing; rng = Xorshifts.Xoroshiro128Plus(rand(UInt64)),
-                                adjoint_sensealg = BacksolveAdjoint(),
+                                adjoint_sensealg = BacksolveAdjoint(autojacvec=ReverseDiffVJP()),
```
ChrisRackauckas, line 781: why hardcode reversediffvjp into the defaults here?

frankschae: `NILSAS` does not have a corresponding `adjoint_sensitivities` function yet. So, I think there are three possibilities:
- hardcode it (basically chosen because I just wanted to see if all tests still pass)
- add an algorithm selection within NILSAS 
```julia
adjoint_sensealg = if isinplace(sol.prob)
      setvjp(adjoint_sensealg, inplace_vjp(sol.prob, sol.prob.u0, sol.prob.p, false, false))
    else
      setvjp(adjoint_sensealg, ZygoteVJP())
    end
```
(with the downside that it does not have the fallback to a numerical vjp)
- add a corresponding `adjoint_sensitivities` function dispatch for `NILSAS` with the try and catch functionality. 

Should we go for the last one?

ChrisRackauckas: Let's go with what we have now then, and open an issue to improve this in the future.

src/concrete_solve.jl
```diff
@@ -98,7 +95,10 @@ function DiffEqBase._concrete_solve_adjoint(prob::Union{ODEProblem,SDEProblem},
   else
     has_cb = false
   end
-  default_sensealg = automatic_sensealg_choice(prob,u0,p,has_cb,verbose)
+  default_sensealg = automatic_sensealg_choice(prob,u0,p,verbose)
+  if has_cb
+    default_sensealg = setvjp(default_sensealg, ReverseDiffVJP())
```
ChrisRackauckas, line 100: no tape compilation?

src/concrete_solve.jl
ChrisRackauckas: can you isolate the enzyme issue?

src/concrete_solve.jl
frankschae: yeah.. tape compilation seems to fail for instance in:
```julia
using OrdinaryDiffEq, Zygote
using DiffEqSensitivity, Test, ForwardDiff

abstol = 1e-12
reltol = 1e-12
savingtimes = 0.5

# test the continuous callbacks wrt to the equivalent discrete callback
function f!(du, u, p, t)
  #Bouncing Ball
  du[1] = u[2]
  du[2] = -p[1]
end

# no saving in Callbacks; prescribed vafter and vbefore; loss on the endpoint

tstop = 3.1943828249997
vbefore = -31.30495168499705
vafter = 25.04396134799764

u0 = [50.0, 0.0]
tspan = (0.0, 5.0)
p = [9.8, 0.8]

prob = ODEProblem(f!, u0, tspan, p)

condition2(u, t, integrator) = t == tstop
cb2 = DiscreteCallback(condition2, affect!, save_positions=(false, false))

du01, dp1 = Zygote.gradient(
  (u0, p) -> sum(solve(prob, Tsit5(), u0=u0, p=p,
    callback=cb2,
    tstops=[tstop],
    sensealg=BacksolveAdjoint(autojacvec=ReverseDiffVJP(true)),
    saveat=tspan[2], save_start=false)), u0, p)
# ([-0.0, 1.0], [-9.435743866328323, 0.0])

du01, dp1 = Zygote.gradient(
  (u0, p) -> sum(solve(prob, Tsit5(), u0=u0, p=p,
    callback=cb2,
    tstops=[tstop],
    sensealg=BacksolveAdjoint(autojacvec=ReverseDiffVJP(false)),
    saveat=tspan[2], save_start=false)), u0, p)

# ([1.0, 5.999999999999999], [-17.499999999999964, 0.0]) (correct value)
```

src/concrete_solve.jl
frankschae: It happens in: 
https://github.com/SciML/DiffEqSensitivity.jl/blob/e4fcd60b8bf9853b3d79134a4643043a410ba4ea/src/derivative_wrappers.jl#L471
(Should this be `S.f` here?)

coming from 
https://github.com/SciML/DiffEqSensitivity.jl/blob/e4fcd60b8bf9853b3d79134a4643043a410ba4ea/src/callback_tracking.jl#L278

with the new `CallbackSensitivityFunction`
https://github.com/SciML/DiffEqSensitivity.jl/blob/master/src/callback_tracking.jl#L200-L203

frankschae: `S.f` fails a few lines later https://github.com/SciML/DiffEqSensitivity.jl/blob/e4fcd60b8bf9853b3d79134a4643043a410ba4ea/src/derivative_wrappers.jl#L501
because of `Enzyme.Duplicated(y, tmp1)`
```julia
julia> typeof(y)
SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}

julia> typeof(tmp1)
Vector{Float64} (alias for Array{Float64, 1})
```
So 
```
struct CallbackSensitivityFunction{fType,Alg<:DiffEqBase.AbstractSensitivityAlgorithm,C<:AdjointDiffCache,pType} <: SensitivityFunction
    f::fType
    sensealg::Alg
    diffcache::C
    prob::pType
end
```
should probably get a `y` of `u` type. 

```julia
ERROR: LoadError: MethodError: no method matching Enzyme.Duplicated(::SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, ::Vector{Float64})
Closest candidates are:
  Enzyme.Duplicated(::T, ::T) where T at ~/.julia/packages/Enzyme/7MHm8/src/Enzyme.jl:66
Stacktrace:
  [1] _vecjacobian!(dλ::SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{I
```

src/concrete_solve.jl
ChrisRackauckas: oh yes, because the callback machinery has branching.

src/concrete_solve.jl
ChrisRackauckas: Open an issue on this? It's not clear to me what causes this, but this is probably an important one because Enzyme should be a good default here.

PR #557, opened 2022-02-23
throw an error when NILSAS is used in `concrete_solve`

frankschae:
changes `dg(out,u,p,t,i)` to `dg(out,u,p,t)` for `NILSAS` to match convention in https://diffeq.sciml.ai/stable/analysis/sensitivity/#Example-continuous-adjoints-on-an-energy-functional.

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -675,7 +675,7 @@ end
 
 
 function DiffEqBase._concrete_solve_adjoint(prob,alg,
-                                 sensealg::Union{AdjointLSS,ForwardLSS,NILSS},
+                                 sensealg::AbstractShadowingSensitivityAlgorithm,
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Interestingly, the merge had some red tests?

frankschae:
Where did you see it? I see all tests passed above. Some of them I didn't run on my laptop because they exceed the available RAM but other than that I didn't notice anything.

ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/actions/runs/1916815779 yeah it only failed after merge.

frankschae:
odd.. but it fails for neural SDEs. I think that should be independent of the changes here. I can take a look if I can reproduce it somehow.

ChrisRackauckas:
Maybe just a stochastic failure?

PR #561, opened 2022-03-16
Update for Enzyme 0.9

ChrisRackauckas:

PR #562, opened 2022-03-16
Bump Enzyme

wsmoses:
@ChrisRackauckas I presume this should close most of your "does not throw Julia error issues", among others


wsmoses:
Potential dup of https://github.com/SciML/DiffEqSensitivity.jl/pull/561

Issue #555, opened 2022-02-07
Callbacks with initial values as a matrix


timkimd:
The following code uses `BacksolveAdjoint()` and uses callbacks to perturb ODE trajectories at certain time points. The initial condition given to the solver is a matrix, rather than a vector.

```
using DifferentialEquations, DiffEqFlux, DiffEqSensitivity

const T = 50    
const L = 3           
const N = 10

tspan = (0.0f0,1.0f0)
trange = range(tspan[1],tspan[2],length=T)
u0 = Float32[1.; 1.]

sense = BacksolveAdjoint()
f_ = FastChain(FastDense(L, 17, tanh), FastDense(17,L)) # changing f_ simply to FastDense(L,L) works
p1 = initial_params(f_)

function f!(du,u,p,t) # this works, but f with (u, p, t) does NOT work
    du .= f_(u, p[1:length(p1)])
end

function callback_(callback_data)
    function callback_each_trial(j, callback_data)
        affect!(integrator) = integrator.u[:, j] = integrator.u[:, j] .+ 0.1 * tanh.(integrator.p[length(p1)+1:length(p1)+L])
        cbj = PresetTimeCallback(trange[callback_data[j,:]],affect!,save_positions=(false,false))
        return cbj
    end
    
    cbs = CallbackSet(ntuple(j -> callback_each_trial(j, callback_data), size(callback_data, 1))...)
    return cbs
end

affect_times = randn(N, T) .> 0

ode_data = randn(L, N, T)

θ = Float32[p1; zeros(L)]
u0s = randn(L, N)
prob = ODEProblem(f!, u0s, tspan, θ)

function loss_nn_ode(p)
    _prob = remake(prob, u0=u0s, p=p, callback=callback_(affect_times))
    x = Array(solve(_prob, Tsit5(), saveat=trange, 
                reltol=1e-6, abstol=1e-6, sensealg=sense))
    loss = sum(abs2, x .- ode_data)
    return loss, x
end

cb = function (p, l, pred)
    display(l)
    return false
end

# Burn in loss
cb(θ,loss_nn_ode(θ)...)

res1 = DiffEqFlux.sciml_train(loss_nn_ode, θ, ADAM(0.001), cb = cb, maxiters = 5)
```

When the above code is run, this error was encountered:

```
[...]

PR #565, opened 2022-04-04
Reshape state of BacksolveAdjoint to match shape of forward pass

frankschae:
Fixes: https://github.com/SciML/DiffEqSensitivity.jl/issues/555

For matrix-valued initial conditions, callbacks can be defined with respect to a matrix interpretation. In `BacksolveAdjoint()`, we solve the vectorized form backwards. This resulted in the BoundsError in https://github.com/SciML/DiffEqSensitivity.jl/issues/555. Therefore, we need to reshape the state to match the definition within the callback. 

In addition, this PR corrects the selection of the affect function for continuous callbacks.

CC: @timkimd

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -173,7 +173,7 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback,DiscreteCallback,
         tprev = get_tprev(cb,indx,pos_neg)
 
         function w(du,u,p,t,tprev,pos_neg)
-          _affect! = get_affect!(cb.affect!,pos_neg)
+          _affect! = get_affect!(cb,pos_neg)
           fakeinteg = FakeIntegrator([x for x in u],[x for x in p],t,tprev)
           _affect!(fakeinteg)
           du .= fakeinteg.u
@@ -238,7 +238,7 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback,DiscreteCallback,
           if !(sensealg isa QuadratureAdjoint)
             
             function wp(dp,p,u,t,tprev,pos_neg)
-              _affect! = get_affect!(cb.affect!,pos_neg)
+              _affect! = get_affect!(cb,pos_neg)
               fakeinteg = FakeIntegrator([x for x in u],[x for x in p],t,tprev)
               _affect!(fakeinteg)
               dp .= fakeinteg.p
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Looks like real test failures.

frankschae:
@ChrisRackauckas  can you have a look at these test errors? I don't see how they are related to the changes in the PR. Callback tests in the buildkite queue are passing.

ChrisRackauckas:
https://github.com/SciML/GalacticOptim.jl/issues/251 not a good day for GO 😓 .

ChrisRackauckas:
Was this superseded by https://github.com/SciML/DiffEqSensitivity.jl/pull/580 ? @frankschae

src/callback_tracking.jl
```diff
@@ -214,6 +215,12 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback,DiscreteCallback,
         end
 
         update_p = copy_to_integrator!(cb,y,integrator.p,integrator.t,indx,pos_neg)
+        # reshape u and du (y and dy) to match forward pass (e.g., for matrices as initial conditions). Only needed for BacksolveAdjoint
+        if sensealg isa BacksolveAdjoint
+          _size = pos_neg ? size(cb.affect!.uleft[indx]) : size(cb.affect_neg!.uleft[indx])
```
frankschae, line 220: @ChrisRackauckas, no it's not superseded by #580. This addition reshaping the state for matrix-valued initial conditions (#555) is independent.

ChrisRackauckas:
Okay, then this needs to get rebased.

ChrisRackauckas:
That looks like a real test failure.

frankschae:
Yes, it looks like a real test failure -- never saw it before because other tests were failing earlier. I'll try to solve it.

frankschae:
`get_affect` called `(S::PeriodicCallbackAffect)(integrator)`

 https://github.com/SciML/DiffEqCallbacks.jl/blob/3c827b4faf8966e4f1335a7b50b061f385be3f2e/src/iterative_and_periodic.jl#L66

instead of its `affect!` field

```julia
struct PeriodicCallbackAffect{A,dT,Ref1,Ref2}
    affect!::A
   ...
end
```

https://github.com/SciML/DiffEqCallbacks.jl/blob/3c827b4faf8966e4f1335a7b50b061f385be3f2e/src/iterative_and_periodic.jl#L60

PR #568, opened 2022-04-08
Fix forward mode resizing

ChrisRackauckas:


frankschae (commented):
it looks like the shadowing tests fail here as well. So probably there is a memory problem again.

PR #570, opened 2022-04-13
Fix shadowing tests

frankschae:
Shadowing tests were found to fail with 

```julia
ERROR: Package DiffEqSensitivity errored during testing (received signal: KILL)
Stacktrace:
 [1] pkgerror(msg::String)
```
in recent PRs, see #568 and #565. This PR is intended to fix this test failure, which is (probably) a memory problem.

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -690,6 +690,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
                                  u0,p,args...;save_start=true,save_end=true,
                                  saveat = eltype(prob.tspan)[],
                                  save_idxs = nothing,
+                                 g = nothing,
                                  t0skip=zero(eltype(prob.tspan)), t1skip=zero(eltype(prob.tspan)),
                                  kwargs...)
 
@@ -697,12 +698,6 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
     error("Sensitivity analysis based on Least Squares Shadowing is not compatible with callbacks. Please select another `sensealg`.")
   else
     _prob = remake(prob,u0=u0,p=p)
-  end
-
-  if haskey(kwargs, :g)
-    g = kwargs[:g]
-  else
-    g = nothing
   end
 
   # some shadowing sensealgs require knowledge of g
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -690,8 +690,6 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
                                  u0,p,args...;save_start=true,save_end=true,
                                  saveat = eltype(prob.tspan)[],
                                  save_idxs = nothing,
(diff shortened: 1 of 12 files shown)


ChrisRackauckas:
Move the shadowing tests to an exclusive buildkite

src/concrete_solve.jl
```diff
@@ -690,6 +690,7 @@ function DiffEqBase._concrete_solve_adjoint(prob,alg,
                                  u0,p,args...;save_start=true,save_end=true,
                                  saveat = eltype(prob.tspan)[],
                                  save_idxs = nothing,
+                                 g = nothing,
                                  t0skip=zero(eltype(prob.tspan)), t1skip=zero(eltype(prob.tspan)),
```
ChrisRackauckas, line 694: I'm not sure what these kwargs are? Should these be in the algorithm? I think the kwarg checker found a real issue 😅

frankschae: I am a bit confused -- I tried to fix the kwargs warning 

> Warning: Unrecognized keyword arguments found 

in 
https://buildkite.com/julialang/diffeqsensitivity-dot-jl/builds/743#3af4314a-160c-4c3d-adae-7d99ec5da380.
I thought that this warning got triggered because `g` (the instantaneous objective function) was a kwarg and not recognized by `solve`.

Do you have an idea where the error in the Core1 test group comes from? 
```julia

du06,dp6 = Zygote.gradient((u0,p)->sum(solve(prob,Tsit5(),u0=u0,p=p,abstol=1e-14,reltol=1e-14,saveat=0.1,sensealg=ForwardSensitivity())),u0,p)

@test du06 isa ChainRulesCore.NotImplemented
```
Evaluates to `nothing isa ChainRulesCore.NotImplemented` and fails. Seems to be independent of the shadowing methods and the changes in this PR?

ChrisRackauckas: > I thought that this warning got triggered because g (the instantaneous objective function) was a kwarg and not recognized by solve.

Yes, it's not. The list is: https://github.com/SciML/DiffEqBase.jl/blob/master/src/solve.jl#L13-L67 `t0skip` is also not in the list. I'm not sure what they do either 😅

frankschae: Oh, is the warning triggered because the original solution call contains a "g=g"? Then I think I should add `g`
`t0skip`, and `t1skip` as options to `LSS()`. The times `t0skip` and `t1skip` are needed for the dilation term  (to reduce side effects)  as alternative options for windowing
https://github.com/SciML/DiffEqSensitivity.jl/blob/4151ba26d5032d7698fe4d5224ff11f3f2677d41/src/lss.jl#L308

ChrisRackauckas: Yes exactly. It's trying to solidify the common args, with non-common args being pieces documented and held by specific algorithms.

PR #572, opened 2022-05-05
Test master

ChrisRackauckas:


frankschae:
Didn't see how to fix the DiffEqFlux one yet.
https://github.com/SciML/DiffEqSensitivity.jl/runs/6301207611?check_suite_focus=true
The failure in the core 1 group can be solved by the change
NotImplemented -> nothing
in the test file according to
https://github.com/FluxML/Zygote.jl/pull/1205

ChrisRackauckas:
Zygote must be treating one of these differently:

https://github.com/SciML/RecursiveArrayTools.jl/blob/master/src/zygote.jl

Trigger the recursivearraytools tests and try to isolate which one it could be.

PR #573, opened 2022-05-06
NILSAS fix to work with multiple parameters

vleon1234:
Before this fix, NILSAS only worked if `size(p) == size(u)`. In any other case, NILSAS would throw error. 

Now NILSAS outputs the correct sensitivity for any `size(p)` and `size(u)` combination. I have confirmed that NILSS still outputs correct sensitivity for input `size(p) == 1`. Tests were done on Lorenz system varying number of input parameters.


frankschae (approved):
LGTM.

PR #574, opened 2022-05-10
NILSS fix for multiple parameter input

vleon1234:
Before this fix, in any system with more than one free parameter, NILSS would throw error because `vstar` dimension doesn't match `v` dimension. After this change, NILSS still outputs the correct sensitivity for single parameter `p` as in test cases. 

However, this fix reveals that something else unrelated is wrong with how NILSS handles multiple parameters. As number of parameters increases, slope noise increases until `Cinv` becomes un-invertible. I haven't been able to find the cause of this second bug yet. 

I tried setting parameters in this loop back to zero (initial state) at each parameter run, but it didn't help:
https://github.com/vleon1234/DiffEqSensitivity.jl/blob/799efb165b9684f7b2241ad2546f666073959441/src/nilss.jl#L540


frankschae (commented):
good catch! Does it work now for the three parameters of the Lorenz system? Could you add a test example too?

vleon1234:
Currently, in the Lorenz system, NILSS only outputs correct sensitivities for any parameter when there is only one free parameter. For any combination of 2 free parameters, both returned slopes are wrong. 

Example code to get wrong slopes from 3 parameter Lorenz:

`
println("Import libraries")
using OrdinaryDiffEq, DiffEqSensitivity

function lorenz!(du,u,p,t)
    du[1] = p[2]*(u[2]-u[1])
    du[2] = u[1]*(p[1]-u[3]) - u[2]
    du[3] = u[1]*u[2] - (p[3])*u[3]
end

println("Solve transient problem")
tspan_transient = (0.0,100.0)
prob_transient = ODEProblem(lorenz!,u0_trans,tspan_transient,p)
sol_transient = solve(prob_transient, Tsit5())

u0 = sol_transient.u[end]

println("Solve attractor problem")
tspan_attractor = (100.0,150.0)
prob_attractor = ODEProblem(lorenz!,sol_transient[end],tspan_attractor,p)
sol_attractor = solve(prob_attractor,Tsit5())

println("NILSS")
g(u,p,t) = u[end]
nseg = 100
nstep = 101
nilss_prob = NILSSProblem(prob_attractor, NILSS(nseg, nstep), g)
res = shadow_forward(nilss_prob, Tsit5())

`

vleon1234:
I figured out what was wrong with NILSS. Before, NILSS was written to recalculate `w` for each parameter and output wrong values for sensitivities. Based on the original NILSS paper, when doing multiple parameters, we can reuse `w` weights for each parameter. So, I implemented this and fixed some indexing problems that were causing the bugs mentioned above.

Now, NILSS now works for multiple parameters (checked with Lorenz system with combinations of 1, 2, and 3 parameters) and NILSS should be faster for multiple parameters, since `w` is re-used instead of recalculated for each parameter.

My edits should be checked to make sure that they follow best practice for speed.

src/nilss.jl
```diff
@@ -160,7 +160,7 @@ function NILSSProblem(prob, sensealg::NILSS, t=nothing, dg = nothing; nus = noth
     end
   end
   vstar0 = zeros(eltype(u0), numindvar*numparams)
-  w0 = vec(w[:,:,1,1,:])
+  w0 = vec(w[1,:,1,1,:])
```
frankschae, line 163: good catch. Though based on the sentence in the paper: 
"so they could be reused for the new parameter"
it sounds like it should only have an impact on performance.

src/nilss.jl
```diff
@@ -289,20 +289,27 @@ function store_y_w_vstar!(y, w, vstar, sol, nus, numindvar, numparams, iseg)
...
 
       _w = @view w[i,:,:,iseg, j]
       copyto!(_w, (@view sol[indx1:indx2,:]))
     end
   end
 
+  # copying w for all parameters to use same w
+  for i=2:numparams
+    _w = @view w[i,:,:,iseg,:]
+    copyto!(_w, w[1,:,:,iseg,:])
```
frankschae, line 306: w is initialized here: 
https://github.com/SciML/DiffEqSensitivity.jl/blob/749648549bb0068ea5f15952d22e70a9f83f93c0/src/nilss.jl#L148
you can drop the 'numparams' dimension and avoid copying.

frankschae (changes requested):
Please also add a test case that checks the derivative with respect to `LSS`.

vleon1234:
Ok, I added requested changes:
- I dropped `w` dimension `numparams` and fixed associated type definitions in NILSSProblems + some more indexing problems that arose due to the change
- added test case comparing LSS vs NILSS output for multiple parameters. Parameter 3 has a larger tolerance as LSS is quite noise for parameter 3 due to relatively short time integrated due to computational cost. I confirmed with different random seeds that the test is consistent across random initializations and NILSS is less noisy.
- tests pass on my computer

ChrisRackauckas:
Rebase to master?

test/shadowing.jl
```diff
@@ -302,6 +302,57 @@ end
     dp1 = Zygote.gradient((p)->G(p),p)
     @test res1 ≈ dp1[1] atol=1e-10
   end
+
+  @testset "Lorentz" begin
```
frankschae, line 306:
```suggestion 
@testset "Lorenz" begin
```

src/nilss.jl
```diff
@@ -354,7 +355,7 @@ function perp!(w_perp, vstar_perp, w, vstar, dudt, iseg, numparams, nsteps, nus)
     for indx_params=1:numparams
       for indx_nus=1:nus
         _w_perp = @view w_perp[indx_params,:,indx_steps,iseg,indx_nus]
```
frankschae, line 357: Can `w_perp` not be reused for each parameter too?

src/nilss.jl
```diff
@@ -452,11 +453,11 @@ end
 function compute_v!(v,v_perp,vstar,vstar_perp,w,w_perp,a,nseg,nus,indxp)
   _vstar = @view vstar[indxp,:,:,:]
   _vstar_perp = @view vstar_perp[indxp,:,:,:]
-  _w = @view w[indxp,:,:,:,:]
+  _w = @view w[:,:,:,:]
```
frankschae, line 456: This `@view` can be removed.

src/nilss.jl
```diff
@@ -289,20 +289,21 @@ function store_y_w_vstar!(y, w, vstar, sol, nus, numindvar, numparams, iseg)
   copyto!(_y, (@view sol[1:numindvar,:]))
 
   # fill w
+  # only calculate w one time, w can be reused for each parameter
   for j=1:nus
-    for i=1:numparams
-      indx1 = (j-1)*numindvar*numparams + i*numindvar+1
-      indx2 = (j-1)*numindvar*numparams + (i+1)*numindvar
+   for i=1
```
frankschae, line 294: Is something still missing here? Otherwise, I think you can remove the loop?

test/shadowing.jl
frankschae: (Looks like you can change this throughout the file if you want 👍 ).

vleon1234:
Nice catch on reusing `w_perp` Frank. Your requested changes have been implemented. Redundant `@view` have been removed and some misc. fixing of indexing. Tests still pass (confirmed with multiple random initializations).

src/nilss.jl
```diff
@@ -289,20 +290,19 @@ function store_y_w_vstar!(y, w, vstar, sol, nus, numindvar, numparams, iseg)
   copyto!(_y, (@view sol[1:numindvar,:]))
 
   # fill w
+  # only calculate w one time, w can be reused for each parameter
   for j=1:nus
-    for i=1:numparams
-      indx1 = (j-1)*numindvar*numparams + i*numindvar+1
-      indx2 = (j-1)*numindvar*numparams + (i+1)*numindvar
+   indx1 = (j-1)*numindvar*1 + 1*numindvar+1
```
frankschae, line 295: I think the `1*` can go away and instead of (1+1) I would directly write 2 for better readability.

frankschae (commented):
Nice! Looks almost done to me!

frankschae (approved):
All good! (+ we just confirmed that the forward sensitivities are properly resized.)

ChrisRackauckas:
I think the callbacks were fixed and this was just not rebased? Anyways, the failure is at least not related.

Issue #579, opened 2022-05-27
Hybrid Differential Equations (example) not working for implicit solvers (at the example of Rosenbrock23)


ThummeTo:
Dear team,

I just found out the example here [Training Neural Networks in Hybrid Differential Equations](https://diffeqflux.sciml.ai/dev/examples/hybrid_diffeq/#Training-Neural-Networks-in-Hybrid-Differential-Equations) doesn't work for implicit solvers in general. So this might be the reason for issue #564. If I **only change the solver in line 34** from it's original:
`Array(solve(_prob,Tsit5(),u0=z0,p=p,callback=cb,saveat=t,sensealg=ReverseDiffAdjoint()))[1:2,:]`
to an implicit solver, it fails with the following errors:

(1) Rosenbrock, no autodiff: `Array(solve(_prob,Rosenbrock23(autodiff=false),u0=z0,p=p,callback=cb,saveat=t,sensealg=ReverseDiffAdjoint()))[1:2,:]`
Error:
```
ERROR: MethodError: Cannot `convert` an object of type 
[...]

PR #580, opened 2022-05-27
Fix BacksolveAdjoint() for implicit solvers

frankschae:
This PR adds a few more tests for adjoint sensitivity methods using implicit solvers to ensure multiple `sensealg` choices are covered by the tests. `BacksolveAdjoint()` failed when the state was copied to its cache. 

ReverseDiffAdjoint seems to be broken too. Until we fix that, it won't work for the callbacks either:
https://github.com/SciML/DiffEqSensitivity.jl/issues/579

### DIFF
--- a/src/backsolve_adjoint.jl
+++ b/src/backsolve_adjoint.jl
@@ -21,12 +21,16 @@ function (S::ODEBacksolveSensitivityFunction)(du,u,p,t)
 
   λ,grad,_y,dλ,dgrad,dy = split_states(du,u,t,S)
 
-  copyto!(vec(y), _y)
+  if eltype(_y) <: ForwardDiff.Dual # handle implicit solvers
+    copyto!(vec(y), ForwardDiff.value.(_y))
+  else
+    copyto!(vec(y), _y)
+  end
 
   if S.noiseterm
     if length(u) == length(du)
-      vecjacobian!(dλ, y, λ, p, t, S, dgrad=dgrad,dy=dy)
-    elseif length(u) != length(du) &&  StochasticDiffEq.is_diagonal_noise(prob) && !isnoisemixing(S.sensealg)
+      vecjacobian!(dλ, y, λ, p, t, S, dgrad=dgrad, dy=dy)
+    elseif length(u) != length(du) && StochasticDiffEq.is_diagonal_noise(prob) && !isnoisemixing(S.sensealg)
       vecjacobian!(dλ, y, λ, p, t, S, dy=dy)
       jacNoise!(λ, y, p, t, S, dgrad=dgrad)
     else
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -79,7 +79,6 @@ if GROUP == "Shadowing"
     @time @safetestset "Shadowing Tests" begin include("shadowing.jl") end
 end
 
-
 if GROUP == "GPU"
     activate_downstream_env()
     @time @safetestset "Standard DiffEqFlux GPU" begin include("gpu/diffeqflux_standard_gpu.jl") end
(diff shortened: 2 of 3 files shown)


ChrisRackauckas (approved):

frankschae:
`_y` which is copied to `y` is of the form:

`ForwardDiff.Dual{ForwardDiff.Tag{OrdinaryDiffEq.OrdinaryDiffEqTag, Float64}, Float64, 9}[Dual{ForwardDiff.Tag{OrdinaryDiffEq.OrdinaryDiffEqTag, Float64}}(3.2966043681397785,0.0,0.0,0.0,0.0,0.0,0.0,1.0,0.0,0.0), Dual{ForwardDiff.Tag{OrdinaryDiffEq.OrdinaryDiffEqTag, Float64}}(2.202041540486022,0.0,0.0,0.0,0.0,0.0,0.0,0.0,1.0,0.0), Dual{ForwardDiff.Tag{OrdinaryDiffEq.OrdinaryDiffEqTag, Float64}}(1.4600600883917199,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,1.0)]))`

(and with the `ForwardDiff.value.()` call, it gets converted to a "normal" array)

frankschae:
Steady-state sensitivity test looks like a real test failure (independent of this PR -- also happened in #565):
https://github.com/SciML/DiffEqSensitivity.jl/runs/6629976019?check_suite_focus=true 
any suspects?

ChrisRackauckas:
We need to define the new algorithm traits from SciMLBase in SteadyStateDiffEq. See the SciMLAlgorithms page in SciMLBase.sciml.ai for the list of traits. I thought all were covered so it was non-breaking but I guess I missed one.

ChrisRackauckas:
Fixed, but there's a NaN in the derivative. 

https://github.com/SciML/DiffEqSensitivity.jl/runs/6636605630?check_suite_focus=true#step:6:402

can you check that out?

frankschae:
Yes, I believe there is one test that sometimes randomly resulted in a NaN. I'll try to fix it/see where it comes from.

frankschae:
The NaN comes from 
https://github.com/SciML/DiffEqSensitivity.jl/blob/68feef223f85e6741c6236f85020b1684022f24c/src/derivative_wrappers.jl#L795
In some cases, the cache contains NaNs, e.g.
```julia
[ Info: FiniteDiff.GradientCache{Nothing, Nothing, Nothing, Vector{Float64}, Val{:central}(), Float64, Val{true}()}(nothing, nothing, nothing, [2.8160187127e-314, 1.55196e-319, NaN, 0.0])
```
and then the NaNs propagate through the other functions after:
https://github.com/SciML/DiffEqSensitivity.jl/blob/master/src/steadystate_adjoint.jl#L127-L135
```julia
diffcache.dg_val = [0.0, 0.0]
diffcache.g = DiffEqSensitivity.UGradientWrapper{var"#g#10", Nothing, Vector{Float64}}(var"#g#10"(), nothing, [2.0, -2.0, 1.0, -4.0])
y = [0.9999999999999982, 0.24999999999999897]
diffcache.g_grad_config = FiniteDiff.GradientCache{Nothing, Nothing, Nothing, Vector{Float64}, Val{:central}(), Float64, Val{true}()}(nothing, nothing, nothing, [2.8160187127e-314, 1.55196e-319, NaN, 0.0])
diffcache.dg_val = [NaN, NaN]
```
Is that a bug in FiniteDiff.jl?

ChrisRackauckas:
It's uninitialized memory. If it's supposed to be zeros then it should be zero'd.

ChrisRackauckas:
@Tokazama see the test failure here.

PR #588, opened 2022-05-31
update ForwardLSS and AdjointLSS

frankschae:
This PR updates the docs for ForwardLSS and AdjointLSS.
I slightly restructured the keyword arguments. Before a numerical value of the weight `alpha` of the dilation term determined whether dilation (with `t0skip` and `t1skip`) or windowing is used. Now, the weight and the two truncation times are moved to the `TimeDilation` struct and the keyword is called `LSSregularizer`.

(I'll do `NILSS` and `NILSAS` in a follow-up PR)

### DIFF
--- a/src/lss.jl
+++ b/src/lss.jl
@@ -198,7 +198,7 @@ function LSSSchur(dt,u0,numindvar,Nt,Ndt,LSSregularizer::TimeDilation)
   LSSSchur(wBinv,wEinv,B,E)
 end
 
-function LSSSchur(dt,u0,numindvar,Nt,Ndt,LSSregularizer::Union{CosWindowing,Cos2Windowing})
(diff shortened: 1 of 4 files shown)


src/lss.jl
```diff
@@ -263,9 +267,9 @@ function B!(S::LSSSchur,dt,umid,sense,sensealg)
   return nothing
 end
 
-E!(S::LSSSchur,dudt,alpha::Union{CosWindowing,Cos2Windowing}) = nothing
+E!(S::LSSSchur,dudt,LSSregularizer::Union{CosWindowing,Cos2Windowing}) = nothing
```
ChrisRackauckas, line 270: Make an AbstractCosWindowing?

frankschae:
The callback test failures are unrelated I think. I'll clean them up after https://github.com/JuliaDiff/FiniteDiff.jl/pull/136 is merged.

Issue #582, opened 2022-05-29
Clean up automatic differentiation and sensitivity analysis infrastructure and documentation


ChrisRackauckas:
- [x] Make DiffEqSensitivity.jl a fully-documented automatic differentiation extension system for SciML
  - [x] Move as much as possible to docstrings, and try to document as many functions as possible 
- [x] Rename DiffEqSensitivity.jl to SciMLSensitivity.jl
- [ ] Move other automatic differentiation overloads to this package to allow this to be the only package with AD dependencies (no Requires.jl)
  - [ ] Quadrature.jl 
- [x] Simplify DiffEqFlux.jl documentation to be just about its special extension layers
- [x] Make the DiffEqDocs.jl simply point to this documentation for sensitivity analysis
- [x] Add the high-level automatic choice algorithm of `sciml_train` to GalacticOptim.jl
- [x] Remove all uses of `sciml_train` and make it all just use GalacticOptim.jl directly
- [x] Move all DiffEqFlux.jl tests which don't use the DiffEqFlux.jl layers to DiffEqSensitivity.jl
- [x] Enable doctests as much as possible (as much as compute allows)
- [x] Deprecate FastChain and just use Lux.jl instead

ChrisRackauckas:
@frankschae can you finish up the docstrings on the shadowing methods and the noise choices?

ChrisRackauckas:
@avik-pal can you flush out the SteadyStateAdjoint docs and docstrings?

ChrisRackauckas:
@Abhishek-1Bhatt can cover the removal of `sciml_train` from the doc examples? 

I'll create the new GalacticOptim subpackage

ChrisRackauckas:
This metaissue has served its purpose.

PR #592, opened 2022-06-02
replaced deprecated functions

Abhishek-1Bhatt:
[SciML/DiffEqSensitivity.jl/issues/582](https://github.com/SciML/DiffEqSensitivity.jl/issues/582)


docs/src/sde_fitting/neural_sde.md
```diff
@@ -85,12 +90,12 @@ Let's see what that looks like:
...
-prediction0 = neuralsde(u0)
+prediction0, st1, st2 = neuralsde(u0,p1,p2,st1,st2)
 
-drift_(u, p, t) = drift_dudt(u, p[1:neuralsde.len])
-diffusion_(u, p, t) = diffusion_dudt(u, p[(neuralsde.len+1):end])
+drift_(u, p, t) = drift_dudt(u, p1, st1)[1]
+diffusion_(u, p, t) = diffusion_dudt(u, p2, st2)[1]
 
-prob_neuralsde = SDEProblem(drift_, diffusion_, u0,(0.0f0, 1.2f0), neuralsde.p)
+prob_neuralsde = SDEProblem(drift_, diffusion_, u0,(0.0f0, 1.2f0), [p1;p2])
```
Abhishek-1Bhatt, line 98: How should we handle passing the parameters of both chains here? Because doing `[p1,p2]` as above returns a flattened vector of the parameters which won't be restructured by `SDEProblem` I guess to call `drift_` and `diffusion_`

ChrisRackauckas:
Can you setup doctests to ensure correctness?

docs/src/sde_fitting/neural_sde.md
ChrisRackauckas: Create a ComponentArray of the two ComponentArrays. They should nest fine, and then dereference them?

Abhishek-1Bhatt:
Yes, it doesn't make sense to check all of these locally, so its kinda necessary to automate it. Same for SciML/DiffEqFlux.jl/pull/711

ChrisRackauckas:
This looks good, so I'll merge, but let's look back around for doctesting

PR #593, opened 2022-06-03
NILSS fix to reduce memory allocations and remove initial time step estimate

vleon1234:
Before this fix, the shadow forward ODEs would be solved with fixed `dt` dictated by `nseg` and `nstep`. This meant that solver behavior was really inconsistent and output sensitivities could be noisy. Also, memory allocation was very high.

Now, the shadow forward ODEs are solved with adaptive timestep. For larger systems, I've seen this fix reduce computational time and memory usage by 50 to 100x. For the Lorenz 1 parameter system, I've seen it reduce solve times by 10x. This fix is also simultaneously implementing the "snapshotting" method discussed in the NILSS and FD-NILSS paper, since `nstep` is now decoupled from the shadow forward ODE's solver timestep. `nstep` can now be optimized by user to just improve integration estimates of NILSS algorithm internal states instead of trying to optimize for solver behavior and integration estimates simultaneously.

I've confirmed output slopes are correct on the Lorenz 3 parameter system, and tests still pass on my computer.

### DIFF
--- a/src/nilss.jl
+++ b/src/nilss.jl
@@ -263,7 +263,7 @@ function forward_sense(prob::NILSSProblem,nilss::NILSS,alg)
     # compute y, w, vstar
     # _sol is a numindvar*(1+nus+1) x nstep matrix
                
-    dt = (t2 - t1) / floor(nstep-1)
+    dt = (t2 - t1) / (nstep-1)
     _sol = Array(solve(_prob, alg, saveat=t1:dt:t2))
                     
     store_y_w_vstar!(y, w, vstar, _sol, nus, numindvar, numparams, iseg)
--- a/src/nilss.jl
+++ b/src/nilss.jl
@@ -262,7 +262,10 @@ function forward_sense(prob::NILSSProblem,nilss::NILSS,alg)
   for iseg=1:nseg
     # compute y, w, vstar
     # _sol is a numindvar*(1+nus+1) x nstep matrix
-    _sol = Array(solve(_prob, alg, saveat=dtsave))
+               
+    dt = (t2 - t1) / floor(nstep-1)
+    _sol = Array(solve(_prob, alg, saveat=t1:dt:t2))
+                    
     store_y_w_vstar!(y, w, vstar, _sol, nus, numindvar, numparams, iseg)
 
     # store dudt, objective g (gsave), and its derivative wrt. to u (dgdu)
--- a/src/nilss.jl
+++ b/src/nilss.jl
@@ -262,7 +262,7 @@ function forward_sense(prob::NILSSProblem,nilss::NILSS,alg)
   for iseg=1:nseg
     # compute y, w, vstar
     # _sol is a numindvar*(1+nus+1) x nstep matrix
-    _sol = Array(solve(_prob, alg, saveat=dtsave, dt=dtsave)(_prob.tspan[1]:dtsave:_prob.tspan[2]))
+    _sol = Array(solve(_prob, alg, saveat=dtsave))
     store_y_w_vstar!(y, w, vstar, _sol, nus, numindvar, numparams, iseg)
 
     # store dudt, objective g (gsave), and its derivative wrt. to u (dgdu)


ChrisRackauckas:
This looks correct to me, I'm curious why it was added in the first place @frankschae

src/nilss.jl
```diff
@@ -262,7 +262,7 @@ function forward_sense(prob::NILSSProblem,nilss::NILSS,alg)
   for iseg=1:nseg
     # compute y, w, vstar
     # _sol is a numindvar*(1+nus+1) x nstep matrix
-    _sol = Array(solve(_prob, alg, saveat=dtsave, dt=dtsave)(_prob.tspan[1]:dtsave:_prob.tspan[2]))
+    _sol = Array(solve(_prob, alg, saveat=dtsave))
```
frankschae, line 265: IIRC this addition was necessary because if I ran NILSS many times, there could be endpoint problems (the saved states (`v`, `w`) are preallocated, and then there was a size mismatch).

frankschae (commented):
If the size mismatch no longer occurs, I'm fine with the change (I think this might be the origin of the speedup, since there's no extra copy). However, I do not understand how the description in the PR fits? 
The solver should also do adaptive steps in the current version? The kwarg `dt=dtsave` should only give an initial estimate of the step size -- I thought I confirmed in the implementation that this can help to make very few steps on a segment?
 
@vleon1234  could you point me to the lines in the paper where the "snapshotting" method is discussed?

vleon1234:
> If the size mismatch no longer occurs, I'm fine with the change (I think this might be the origin of the speedup, since there's no extra copy). 

I haven't seen size mismatch with the proposed change, so I think this is not an issue.

> However, I do not understand how the description in the PR fits? The solver should also do adaptive steps in the current version? The kwarg dt=dtsave should only give an initial estimate of the step size -- I thought I confirmed in the implementation that this can help to make very few steps on a segment?

Good point! I checked the solver behavior and it is adaptive as you said. With `dt` option it just sets the initial stepsize. However, taking that initial fixed stepsize seems to make the solver behavior worse with regards to solver stability and noise. I did a quick experiment with 1 parameter Lorenz getting `d<z>/drho` to demonstrate this, but I observe that this decrease in solver stability and noise get worse with a larger system: 

Experiment is run parameter sweep `rho = 14.0:1.0:28.0` 10 times. So reported values below are 140 NILSS calls in chaotic and non-chaotic regime.

without `dt`, sensitivities are average (1.03) and std dev (0.12).
with `dt`, sensitivities are average (0.91) and std dev(0.31).

I guess a more accurate PR would be remove initial step estimate to improve solver stability?

> @vleon1234 could you point me to the lines in the paper where the "snapshotting" method is discussed?

I realized that snapshotting is discussed in older versions of NILSS and not the most recent. It was removed it from NILSS paper after they published the standalone FD-NILSS paper. They discuss snapshotting concept in pg 11 around equation 46: https://arxiv.org/pdf/1711.06633.pdf

Essentially, they define snapshotting as just doing integral estimates using the start and end point values and not with the middle values. With the fix I did, you can basically do this by setting `nseg=2`, since the solver then stores only the start and end point. However, I find that this is really noisy, likely because the integral estimates are quite bad. For 1 parameter Lorenz, I've found that `nseg=10` works nicely, but you can increase more to decrease noise in the integral estimate.

src/nilss.jl
```diff
@@ -262,7 +262,10 @@ function forward_sense(prob::NILSSProblem,nilss::NILSS,alg)
   for iseg=1:nseg
     # compute y, w, vstar
     # _sol is a numindvar*(1+nus+1) x nstep matrix
-    _sol = Array(solve(_prob, alg, saveat=dtsave, dt=dtsave)(_prob.tspan[1]:dtsave:_prob.tspan[2]))
+               
+    dt = (t2 - t1) / (nstep-1)
```
frankschae, line 266: LGTM!

frankschae (approved):
Nice! Thanks for the precise analysis and the re-examination of the size mismatch! 

> Essentially, they define snapshotting as just doing integral estimates using the start and end point values of a segment and not with the middle values. With the fix I propose, you can basically do this by setting nseg=2, since the solver then stores only the start and end point. However, I find that this is really noisy, likely because the integral estimates are quite bad. For 1 parameter Lorenz, I've found that nseg=10 works nicely, but you can increase more to decrease noise in the integral estimate.

I see. We actually have a lot of good integration tools at our disposal (and already in use in DiffEqSensitivity). We can maybe see if we can also use some of them to compute Eq. (38). In general, implementing FD-NILSS based on the NILSS implementation we already have might be good idea too (not in this PR).

ChrisRackauckas:
The integral can be computed using pseudo values in the integration process. Or, you can implement the analytical solution to the interpolation and just add the values. Anyways, this can merge.

PR #599, opened 2022-06-06
test master

ChrisRackauckas:


ChrisRackauckas:
@frankschae master has the same issue, so I'll merge the dependency removal but this should get looked into

PR #600, opened 2022-06-06
split callback tests

frankschae:
Split tests into two buildkite groups and independent tests over different files.


ChrisRackauckas:
oh it's using powerpc. https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/.buildkite/pipeline.yml#L29 let's set the arch

frankschae:
for all buildkite tests?

Issue #564, opened 2022-03-29
`type TrackedAffect has no field funciter` when training hybrid models with implicit solvers


ThummeTo:
The error (s. title) occurs when training a hybrid model (DiffEqFlux.jl) with Backward-Sensitivity-Methods like `ReverseDiffAdjoint` and using `rootfind=RightRootFind`. If this is more a DiffEqCallbacks-Issue, feel free to re-organize the thread.

Please note, that the provided MWE is a copy of the official [DiffEqFlux-Tutorial for hybrid modelling](https://diffeqflux.sciml.ai/stable/examples/hybrid_diffeq/). The only ~~three~~ four things modified are:

- a VectorContinuousCallback for state events was added (with option `rootfind=RightRootFind`)
- a dummy-condition `condition` for state callback triggering was added
- both callbacks are added to the `solve`-call
- **EDIT:** I forgot to say that I am using an implicit solver like `Rosenbrock23(autodiff=false)`. For other solvers like `Rodas4(autodiff=false)` I still get errors, but other ones (`MethodError: no method matching OrdinaryDiffEq.RodasTableau(::ReverseDiff.TrackedReal{Float32...` or `MethodError: no method matching (::DiffEqSensitivity.TrackedAffect{Float32...`). There seems no issue with explicit solvers like e.g. `Tsit5()`.

Thanks in advance and best regards!

MWE (teste in Julia 1.6.5 LTS with current library releases):

```
using DiffEqFlux, DifferentialEquations, Plots
import SciMLBase: RightRootFind

u0 = Float32[2.; 0.]
datasize = 100
tspan = (0.0f0,10.5f0)
dosetimes = [1.0,2.0,4.0,8.0]

function affect!(integrator)
    integrator.u = integrator.u.+1
end
cb_ = PresetTimeCallback(dosetimes,affect!,save_positions=(false,false))
function trueODEfunc(du,u,p,t)
    du .= -u
end
t = range(tspan[1],tspan[2],length=datasize)

prob = ODEProblem(trueODEfunc,u0,tspan)
ode_data = Array(solve(prob,Tsit5(),callback=cb_,saveat=t))
dudt2 = Chain(Dense(2,50,tanh),
             Dense(50,2))
p,re = Flux.destructure(dudt2) # use this p as the initial condition!

function dudt(du,u,p,t)
    du[1:2] .= -u[1:2]
    du[3:end] .= re(p)(u[1:2]) #re(p)(u[3:end])
end
[...]

PR #602, opened 2022-06-07
Test and fix `VectorContinuousCallback`

frankschae:
`VectorContinuousCallback`s are not fully implemented or tested yet (see e.g. https://github.com/SciML/DiffEqSensitivity.jl/issues/564). This PR adds a test and

- [x]  in contrast to `ContinuousCallback`s, we also have to track event indices for `VectorContinuousCallback`s. 
- [x]   `get_indx(cb::Union{ContinuousCallback,VectorContinuousCallback}, t)` always checked first for an event in `affect!`. However, if the last event is in `affect_neg!`, `searchsortedfirst` provided an index > `len(event_times)`. This is now solved by the `min()` function. 
- [x]  the condition function `g` in `VectorContinuousCallback`s operates on a vector. (in contrast to a scalar condition in `ContinuousCallback`s.). Therefore, the wrappers for computing the derivative `dg/dt` and the gradient `dg/du` must be adjusted.
- [x]  need to think about how to handle several true conditions for a single event?

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -314,16 +314,18 @@ end
 
 get_indx(cb::DiscreteCallback,t) = (searchsortedfirst(cb.affect!.event_times,t), true)
 function get_indx(cb::Union{ContinuousCallback,VectorContinuousCallback}, t)
-  if !isempty(cb.affect!.event_times)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
>need to think about how to handle several true conditions for a single event?

VCCs just error in that case right now, so it's fine to not handle its derivative and document that for now. But if you want to solve that issue...

PR #604, opened 2022-06-08
Partially revert removal of DiffEqSensitivity from deps

DhairyaLGandhi:
https://github.com/SciML/DiffEqSensitivity.jl/blob/058c6f5532c2f0f1eef45af0c98512a8cada0f23/src/steadystate_adjoint.jl#L140 still uses DiffEqOperators, which needs to be updated in favour of SparseDiffTools. Until then, allow DiffEqOperators.jl to be in the deps so as to allow downstream packages' CI to work (ref https://github.com/Chemellia/AtomicGraphNets.jl/pull/108)


thazhemadam:
Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/605.

PR #632, opened 2022-06-18
Allow for alternative AD front-ends

ChrisRackauckas:


ChrisRackauckas:
A bunch of unexpected passes. Nice.

PR #637, opened 2022-06-19
Avoid aliasing in NeuralODE training

Abhishek-1Bhatt:
Copies the function output to avoid aliasing issues in Neural ODE Training with SimpleChains. Related to PumasAI/SimpleChains.jl/pull/89


ChrisRackauckas:
No this isn't the right solution, I think I see the right one.

PR #638, opened 2022-06-19
Mutation robust VJPs for SimpleChains

ChrisRackauckas:


ChrisRackauckas:
Since this is at least not bad to do I'll merge, but @Abhishek-1Bhatt please follow up with a simplechains test

PR #643, opened 2022-06-20
try a few more docs example building

ChrisRackauckas:


ChrisRackauckas:
@frankschae why is ForwardSensitivity trying to set a vjp? https://github.com/SciML/DiffEqSensitivity.jl/runs/6974538189?check_suite_focus=true#step:5:94 That seems like a bug.

frankschae:
yeah that looks strange. I'll have a look.

PR #644, opened 2022-06-20
Fix tests for out-of-place constant requirement

ChrisRackauckas:


ChrisRackauckas:
The GPU one is weird 😅

PR #648, opened 2022-06-21
Doctests for training tips

Abhishek-1Bhatt:


docs/src/training_tips/multiple_nn.md
```diff
@@ -78,18 +79,18 @@ optprob = Optimization.OptimizationProblem(optf, p)
...
-first neural network's evolution is written like `NN_1([v,w], p[1:length(p1)])`.
+ComponentVector `p` and then train on this parameter
+vector. Whenever we need to evaluate the neural networks, we dereference the
+vector and grab the key that corresponds to the neural network.
+For example, the `p1` portion is `p.p1`, which is why the
+first neural network's evolution is written like `NN_1([v,w], p.p1)`.
 
 This method is flexible to use with many optimizers and in fairly
-optimized ways. The allocations can be reduced by using `@view p[1:length(p1)]`.
+optimized ways. The allocations can be reduced by using `@view p.p1`.
```
ChrisRackauckas, line 94: ComponentArrays already does `p.p1` as a view, so `@view is unnecessary here.

Abhishek-1Bhatt, line 94:
```suggestion
optimized ways.
```

ChrisRackauckas:
Hmm, just noticed that the divergence one uses DiffEqFlux, which we don't want to depend on. Can you do a direct definition of the neural network and drop the dependency there?

Abhishek-1Bhatt:
Oh yeah, it's the other two, they use DiffEqFlux for NeuralODE, we can define an ODEProblem directly here

PR #649, opened 2022-06-21
Even more docs clean up

ChrisRackauckas:
- Neural SDE -> DiffEqFlux
- Delete SGLD: people should use Turing instead of a handrolled thing
- Make shadow adjoints a doc example


ChrisRackauckas:
https://github.com/SciML/DiffEqSensitivity.jl/runs/6987504279?check_suite_focus=true#step:5:37 @frankschae can you help finish up the shadow adjoint one?

PR #652, opened 2022-06-22
remove some hard-coded images

baggepinnen:


ChrisRackauckas:
Why only those two? Or in other words, why don't the others display?

baggepinnen:
I removed all the others in the PEM tutorial in the previous PR, but had missed these two.

PR #655, opened 2022-06-22
Add code coverage from documentation builds

ChrisRackauckas:


ChrisRackauckas:
Code coverage on docs here makes it too slow 😅 . Ouch, we need to split that build.

PR #657, opened 2022-06-22
Allow for mixed-type cost contributions in continuous adjoints

frankschae:
This PR introduces a new signature to the `adjoint_sensitivities` interface
```julia
adjoint_sensitivity(sol, alg, ts=nothing, dg_discrete=nothing, dg_continuous=nothing, g=nothing)
```
Upon completion, we should be able to handle mixed discrete and continuous loss formulations within the direct `adjoint_sensitivities` interface.

Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/544 and #545


ChrisRackauckas:
Make them keyword arguments instead? Since you don't need to define `dg_discrete` to do continuous.

ChrisRackauckas:
Amazing.

ChrisRackauckas:
It's much easier to read like this too.

Issue #661, opened 2022-06-23
SDE Controls tutorial fix


ChrisRackauckas:
@frankschae can you take this on? I left this out of https://github.com/SciML/DiffEqSensitivity.jl/pull/659, but it's the last tutorial to fix other than https://github.com/SciML/DiffEqSensitivity.jl/pull/635

frankschae:
Should we switch to Lux + Optimization.jl for this example too?

ChrisRackauckas:
If you want. I don't think the choice is so important for Lux vs Flux right now. Optimization.jl, yes we should.

ChrisRackauckas:
I removed the running in #662 to execute the DiffEqSensitivity -> SciMLSensitivity change with correct docs, but once that's done, let's open up the PR again and see why it stalled.

frankschae:
hmm it was finished on my machine over lunch and I checked the plot. I'll check how expensive it actually is.

ChrisRackauckas:
Too high of a peak memory requirement will easily kill the CI computers. That might be it

ChrisRackauckas:
Looks like it's fixed, other metaissue

PR #662, opened 2022-06-23
More doctests: SDE control tutorials

frankschae:
see #661


frankschae:
It currently fails on a different example: 
```
Error: failed to run `@example` block in src/training_tips/multiple_nn.md:9-83
```
with 
```
 value = DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 21 and 6")
```

ChrisRackauckas:
Huh, it was working, and works locally. Looks like it was just a bad RNG seed that diverged.

ChrisRackauckas:
Killed... is the new tutorial expensive 😅

PR #665, opened 2022-06-25
Rename and set as a breaking change

ChrisRackauckas:
There are a lot of breaking changes in this one, especially the sign change and the mixed adjoint (continuous + discrete) causing changes to the direct version. But the docs are also all fixed so yay!

The reason for the name change is because this package is the core package for all of SciML's integration with AD and sensitivity analysis. For example, the issue with Integrals.jl is that it cannot handle all of the vjps... which are defined here. Optimization.jl needs adjoints, defined here for vjps. Etc. so the name change reflects that, and since we also had a massive breaking API change, it makes sense to pair this together as essentially a new package.


ChrisRackauckas:
Downstream tests should fail with this change?

PR #675, opened 2022-06-27
Support `dgdp` in discrete costs within `adjoint_sensitivities` interface

frankschae:
See: 
https://github.com/SciML/SciMLSensitivity.jl/issues/286
https://github.com/SciML/SciMLSensitivity.jl/issues/302

Continuous cost version: 
https://github.com/SciML/SciMLSensitivity.jl/pull/285 

Throws an error when a discrete cost is used but only a `g` is provided. 
Also computes `dgdp_continuous` via AD/FiniteDiff correctly now if only a `g` is provided.

### DIFF
--- a/src/backsolve_adjoint.jl
+++ b/src/backsolve_adjoint.jl
@@ -11,9 +11,9 @@ struct ODEBacksolveSensitivityFunction{C <: AdjointDiffCache, Alg <: BacksolveAd
     noiseterm::Bool
 end
 
-function ODEBacksolveSensitivityFunction(g, sensealg, discrete, sol, dg, f;
+function ODEBacksolveSensitivityFunction(g, sensealg, discrete, sol, dgdu, dgdp, f;
                                          noiseterm = false)
-    diffcache, y = adjointdiffcache(g, sensealg, discrete, sol, dg, f; quad = false,
+    diffcache, y = adjointdiffcache(g, sensealg, discrete, sol, dgdu, dgdp, f; quad = false,
                                     noiseterm = noiseterm)
 
     return ODEBacksolveSensitivityFunction(diffcache, sensealg, discrete,
@@ -131,9 +131,6 @@ end
                with a discrete cost function but no specified `dgdu_discrete` or `dgdp_discrete`.
                Please use the higher level `solve` interface or specify these two contributions.")
 
-    dg_discrete = (dgdu_discrete, dgdp_discrete)
-    dg_continuous = (dgdu_continuous, dgdp_continuous)
-
     @unpack f, p, u0 = sol.prob
 
     # check if solution was terminated, then use reduced time span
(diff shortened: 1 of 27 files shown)


src/backsolve_adjoint.jl
```diff
@@ -205,16 +216,25 @@ end
...
+                                     kwargs...) where {DG1, DG2, DG3, DG4, G}
+    dgdu_discrete === nothing && dgdu_continuous === nothing && g === nothing &&
+        error("Either `dgdu_discrete`, `dgdu_continuous`, or `g` must be specified.")
+    t !== nothing && dgdu_discrete === nothing && dgdp_discrete === nothing &&
+        error("It looks like you're using the direct `adjoint_sensitivities` interface
+               with a discrete cost function but no specified `dgdu_discrete` or `dgdp_discrete`.
+               Please use the higher level `solve` interface or specify these two contributions.")
+
+    dg_discrete = (dgdu_discrete, dgdp_discrete)
+    dg_continuous = (dgdu_continuous, dgdp_continuous)
```
ChrisRackauckas, line 237: why put them input tuples instead of keeping them separate?

frankschae: I thought it reads slightly easier internally if we have, e.g. only `pg`, `pg_config` and `dg_val` in: 
```julia
pgpu = UGradientWrapper(g, tspan[2], p)
pgpu_config = build_grad_config(sensealg, pgpu, u0, p)
pgpp = ParamGradientWrapper(g, tspan[2], u0)
pgpp_config = build_grad_config(sensealg, pgpp, p, p)
pg = (pgpu, pgpp)
pg_config = (pgpu_config, pgpp_config)
dg_val = (similar(u0, numindvar), similar(u0, numparams))
dg_val[1] .= false
dg_val[2] .= false
```
in the `diffcache` and was not entirely sure where to stop (discrete and continuous never mix in a function, that's why I thought keeping them grouped from the beginning might be a good idea .. but yeah in the end we need every time the unpacked version of `dg_discrete` and `dg_continuous`, respectively.)

frankschae:
I think it would make sense to also update `SteadyStateAdjoint` and the shadowing methods to `dgdu` and `dgdp` (instead of `dg`). @ChrisRackauckas do you agree?

ChrisRackauckas:
yes

test/adjoint_param.jl
```diff
@@ -85,13 +90,12 @@ function model(p)
...
-                              jac = true,
-                              abstol = 1e-12,
-                              reltol = 1e-12),
+                              abstol = abstol,
+                              reltol = reltol),
                    Tsit5(),
-                   jac = true,
                    saveat = collect(0:0.1:7),
-                   sensealg = QuadratureAdjoint())
+                   sensealg = QuadratureAdjoint(autojacvec=ReverseDiffVJP()),
```
frankschae, line 97: This test file wasn't linked in runtests.jl. For the default `EnzymeVJP`, this test fails, see 
https://github.com/SciML/SciMLSensitivity.jl/runs/7171310194?check_suite_focus=true  

@ChrisRackauckas, do you see what's the issue with Enzyme here?

ChrisRackauckas: Probably some of the allocations don't compile away. I'd test the ReverseDiffVJP tape compilation works here, and leave Enzyme as `@test_broken` for now.

PR #680, opened 2022-07-05
Out of place QuadratureAdjoint for Working with StaticArrays

Abhishek-1Bhatt:
Tested on PumasAI/SimpleChains.jl/pull/97 , so should be merged after it. 
NeuralODE example

```julia
using SimpleChains, StaticArrays, OrdinaryDiffEq, SciMLSensitivity, Optimization, OptimizationFlux, Plots

u0 = @SArray Float32[2.0, 0.0]
datasize = 30
tspan = (0.0f0, 1.5f0)
tsteps = range(tspan[1], tspan[2], length = datasize)

function trueODE(u, p, t)
    true_A = @SMatrix Float32[-0.1 2.0; -2.0 -0.1]
    ((u.^3)'true_A)'
end

prob = ODEProblem(trueODE, u0, tspan)
data = Array(solve(prob, Tsit5(), saveat = tsteps))

sc = SimpleChain(
                static(2),
                Activation(x -> x.^3),
                TurboDense{true}(tanh, static(50)),
                TurboDense{true}(identity, static(2))
            )

p_nn = SimpleChains.init_params(sc)

f(u,p,t) = sc(u,p)

prob_nn = ODEProblem(f, u0, tspan)

function predict_neuralode(p)
    Array(solve(prob_nn, Tsit5();p=p,saveat=tsteps,sensealg=QuadratureAdjoint(autojacvec=ZygoteVJP())))
end

function loss_neuralode(p)
    pred = predict_neuralode(p)
    loss = sum(abs2, data .- pred)
    return loss, pred
end

callback = function (p, l, pred; doplot = true)
    display(l)
    plt = scatter(tsteps, data[1,:],label="data")
    scatter!(plt, tsteps, pred[1,:], label = "prediction")
    if doplot
        display(plot(plt))
    end
    return false
end

optf = Optimization.OptimizationFunction((x,p)->loss_neuralode(x), Optimization.AutoZygote())
optprob = Optimization.OptimizationProblem(optf, p_nn)

res = Optimization.solve(optprob, ADAM(0.05),callback=callback,maxiters=300)
```


src/concrete_solve.jl
```diff
@@ -334,6 +334,211 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg,
...
+            outtype = typeof(_out) <: SubArray ?
+                      DiffEqBase.parameterless_type(_out.parent) :
+                      DiffEqBase.parameterless_type(_out)
+            if only_end
+                eltype(Δ) <: NoTangent && return
+                if typeof(Δ) <: AbstractArray{<:AbstractArray} && length(Δ) == 1 && i == 1
+                    # user did sol[end] on only_end
+                    if typeof(_save_idxs) <: Number
+                        x = vec(Δ[1])
+                        _out[_save_idxs] .= adapt(outtype, @view(x[_save_idxs]))
```
frankschae, line 463: Did you check these different cases? Does it really work with the `.=`?

src/concrete_solve.jl
```diff
@@ -334,6 +334,211 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg,
...
+                else
+                    if typeof(_save_idxs) <: Number
+                        _out[_save_idxs] = adapt(outtype,
+                                                 reshape(Δ, prod(size(Δ)[1:(end - 1)]),
+                                                         size(Δ)[end])[_save_idxs, i])
+                    elseif _save_idxs isa Colon
+                        _out = vec(adapt(outtype,
+                                         reshape(Δ, prod(size(Δ)[1:(end - 1)]),
+                                         size(Δ)[end])[:, i]))######Required Assignment#################
+                    else
```
Abhishek-1Bhatt, line 503: This is the only assignment that is required for the neural ode with SArrays to work. The others are indexing on `_out` , so, won't be compatibe with SArrays anyway. Let me know if I should simplify it to just contain this assignment.

src/concrete_solve.jl
```diff
@@ -334,6 +334,211 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg,
...
+                                            u0::StaticArrays.SVector, p, originator::SciMLBase.ADOriginator,
+                                            args...; save_start = true, save_end = true,
+                                            saveat = eltype(prob.tspan)[],
+                                            save_idxs = nothing,
+                                            kwargs...)
+    if !(typeof(p) <: Union{Nothing, SciMLBase.NullParameters, AbstractArray}) ||
+        (p isa AbstractArray && !Base.isconcretetype(eltype(p)))
+        throw(AdjointSensitivityParameterCompatibilityError())
+    end
+
```
frankschae, line 348: maybe add an `@assert sensealg isa QuadratureAdjoint` if we only implement it for this case and throw an error otherwise.

src/concrete_solve.jl
```diff
@@ -334,6 +334,211 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg,
...
+                                                  [_out[i][save_idxs]
+                                                   for i in 1:length(_out)], ts)
+        end
+        only_end = length(ts) == 1 && ts[1] == _prob.tspan[2]
+    end
+
+    _save_idxs = save_idxs === nothing ? Colon() : save_idxs
+
+    function adjoint_sensitivity_backpass(Δ)
+        function df(_out, u, p, t, i)
```
frankschae, line 453: we could also write it as `df(u,p,t,i)`. I think that would make it clearer that it's not mutating.

src/derivative_wrappers.jl
```diff
@@ -868,6 +911,29 @@ function accumulate_cost!(dλ, y, p, t, S::TS,
     return nothing
 end
 
+function accumulate_cost(dλ, y, p, t, S::TS,
+                         dgrad = nothing) where {TS <: SensitivityFunction}
+    @unpack dg, dg_val, g, g_grad_config = S.diffcache
+    if dg !== nothing
+        if !(dg isa Tuple)
+            dg(dg_val, y, p, t)
```
frankschae, line 919: do you have a test case where this is actually triggered (i.e, a continuous and a discrete cost contribution)? I'd think it makes more sense to write this part also fully non-mutating.

frankschae (commented):
Please also add a test set. 

To format the files, you can run
```julia
using JuliaFormatter, SciMLSensitivity
format(joinpath(dirname(pathof(SciMLSensitivity)), ".."))
```

src/adjoint_common.jl
```diff
@@ -419,8 +424,12 @@ function (f::ReverseLossCallback)(integrator)
     if F !== nothing
         F !== I && F !== (I, I) && ldiv!(F, Δλd)
     end
-
-    u[diffvar_idxs] .+= Δλd
+    
+    if u isa StaticArrays.SArray
+        integrator.u += Δλd
```
ChrisRackauckas, line 429: comment that this drops diffvar_idxs because it assumes QuadratureAdjoint, or even better, put an `@assert` in here for that.

src/concrete_solve.jl
ChrisRackauckas: throw a custom error like above but with a custom error message describing that `!ArrayInterface.ismutable(u0)` types require QuadratureAdjoint. Then we should make an issue to extend it afterwards, but put a low priority on it because of the issue with parameter sizing.

src/concrete_solve.jl
```diff
@@ -334,6 +334,211 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg,
     out, adjoint_sensitivity_backpass
 end
 
+function DiffEqBase._concrete_solve_adjoint(prob, alg,
+                                            sensealg::AbstractAdjointSensitivityAlgorithm,
+                                            u0::StaticArrays.SVector, p, originator::SciMLBase.ADOriginator,
+                                            args...; save_start = true, save_end = true,
+                                            saveat = eltype(prob.tspan)[],
+                                            save_idxs = nothing,
+                                            kwargs...)
```
ChrisRackauckas, line 343: this seems excessive, why does it have a completely separate dispatch instead of a few targeted `ArrayInterface.ismutable` checks?

src/concrete_solve.jl
ChrisRackauckas: No it's not, it's dependent on the arguments. Did you try the other arguments?

src/quadrature_adjoint.jl
```diff
@@ -187,8 +210,13 @@ end
 function (S::AdjointSensitivityIntegrand)(out, t)
     @unpack y, λ, pJ, pf, p, f_cache, dgdp_cache, paramjac_config, sensealg, sol, adj_sol = S
     f = sol.prob.f
-    sol(y, t)
-    adj_sol(λ, t)
+    if eltype(sol.u) <: StaticArrays.SArray
```
ChrisRackauckas, line 213:
```suggestion
    if ArrayInterfaceCore.ismutable(eltype(sol.u))
```

ChrisRackauckas:
This doesn't need a whole new dispatch. You just need to handle the mutating cases with a branch, not copy paste the entire code and change like 30 lines.

src/adjoint_common.jl
```diff
@@ -401,9 +401,15 @@ function (f::ReverseLossCallback)(integrator)
...
-    # Warning: alias here! Be careful with λ
-    gᵤ = isq ? λ : @view(λ[1:idx])
-    g(gᵤ, y, p, t[cur_time[]], cur_time[])
+    if ArrayInterfaceCore.ismutable(u)
+        # Warning: alias here! Be careful with λ
+        gᵤ = isq ? λ : @view(λ[1:idx])
+        g(gᵤ, y, p, t[cur_time[]], cur_time[])
+    else
+        @assert sensealg isa QuadratureAdjoint
+        gᵤ = isq ? λ : @view(λ[1:idx])
```
ChrisRackauckas, line 410:
```suggestion
        gᵤ = isq ? λ : λ[1:idx]
```

No views are required on immutables.

src/quadrature_adjoint.jl
```diff
@@ -48,6 +62,18 @@ function split_states(du, u, t, S::ODEQuadratureAdjointSensitivityFunction; upda
...
+function split_states(u, t, S::ODEQuadratureAdjointSensitivityFunction; update = true)
+    @unpack y, sol = S
+
+    if update
+        y = sol(t, continuity = :right)
+    end
+
+    λ = u
+
+    λ, nothing, y, nothing, nothing
```
ChrisRackauckas, line 74: why 5 instead of 3 returns?

src/adjoint_common.jl
Abhishek-1Bhatt: @assert

ChrisRackauckas: true, even better

src/concrete_solve.jl
```diff
@@ -834,7 +913,7 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg, sensealg::ReverseDiffAdjo
 
     function reversediff_adjoint_forwardpass(_u0, _p)
         if (convert_tspan(sensealg) === nothing &&
-            ((haskey(kwargs, :callback) && has_continuous_callback(kwargs[:callback])))) ||
+            ((haskey(kwargs, :callback) && has_a_callback(kwargs[:callback])))) ||
```
ChrisRackauckas, line 916: was this a mistake?

src/derivative_wrappers.jl
```diff
@@ -573,6 +579,43 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::ZygoteVJP, dgrad,
...
+    end
+
+    if dgrad !== nothing
+        if tmp2 === nothing && !sensealg.autojacvec.allow_nothing
+            throw(ZygoteVJPNothingError())
+        elseif tmp2 !== nothing
+            (dgrad[:] .= vec(tmp2))
+        end
+    end
+    return dλ
```
ChrisRackauckas, line 616: There's 3 quantities this function calculates, dy, dlambda, and dgrad

src/concrete_solve.jl
```diff
@@ -283,7 +291,11 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg,
                     if typeof(_save_idxs) <: Number
                         _out[_save_idxs] = @view(x[_save_idxs])
                     elseif _save_idxs isa Colon
-                        vec(_out) .= vec(x)
+                        if ArrayInterfaceCore.ismutable(u)
```
ChrisRackauckas, line 294: You want to revert the changes to this anonymous function, and just if else whether to do this one or the one below.

src/concrete_solve.jl
```diff
@@ -305,6 +323,67 @@ function DiffEqBase._concrete_solve_adjoint(prob, alg,
                     end
                 end
             end
+            if !(ArrayInterfaceCore.ismutable(u0))
+                return _out
+            end
+        end
+
+        function df(u, p, t, i;outtype=nothing)
```
ChrisRackauckas, line 331: to make this work, you probably need to change the dgdu accmulation code in adjoint_common inside of the `affect!` definition

src/concrete_solve.jl
Abhishek-1Bhatt, line 916:
```suggestion
            ((haskey(kwargs, :callback) && has_continuous_callback(kwargs[:callback])))) ||
```

Abhishek-1Bhatt: yeah, pressed `a` accidentally, thought I corrected it, but it crept in

src/concrete_solve.jl
Abhishek-1Bhatt: Oh, so its inferred as a redefinition of the first df, which is causing the tests using the first one to fail

test/adjoint.jl
```diff
@@ -866,3 +866,76 @@ using LinearAlgebra, SciMLSensitivity, OrdinaryDiffEq, ForwardDiff, QuadGK
...
     end
 end
+
+####Fully oop Adjoint
+
+u0 = @SArray Float32[2.0, 0.0]
+datasize = 30
+tspan = (0.0f0, 1.5f0)
+tsteps = range(tspan[1], tspan[2], length = datasize)
+
```
ChrisRackauckas, line 876: This is a StaticArray/SimpleChain regression test for optimization, not an adjoint test. Please add a proper adjoint test. It should probably be its own file `adjoint_oop.jl` where it similarly tests the direct and AD interfaces for gradient correctness.

This test should be a separate SimpleChains regression test.

ChrisRackauckas:
Rebase this onto master?

I think this looks generally correct now. It needs to change the tests as commented above, and of course fix tests, but it looks like it's in the right direction.

Abhishek-1Bhatt:
We'll need to wait for the SimpleChains branch linked at the top to be merged. The tests are passing on it. SimpleChains master doesn't return SArray in the reverse pass so the reverse ode throws the type not constant error, it's fixed in the above branch

test/adjoint_oop.jl
```diff
@@ -0,0 +1,107 @@
+using SciMLSensitivity, OrdinaryDiffEq, SimpleChains, StaticArrays, QuadGK, ForwardDiff,
```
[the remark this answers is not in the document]
ChrisRackauckas: Still missing.

ChrisRackauckas:
Can you add the requested accuracy tests?

test/adjoint_oop.jl
```diff
@@ -0,0 +1,209 @@
...
+                    sensealg = SensitivityADPassThrough(), abstol = 1e-12, reltol = 1e-12))
+    return sum(((1 .- u) .^ 2) ./ 2)
+end
+
+G_p(p)
+G_u(u0)
+n_dp = ForwardDiff.gradient(G_p, p)
+n_du0 = ForwardDiff.gradient(G_u, u0)
+
+@test n_du0≈du0 rtol=1e-3
```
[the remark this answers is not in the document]
Abhishek-1Bhatt: The tests on master are between autodiff and autodiff(different adjoint algorithms combined with different `autojacvec`s) here the tests are between autodiff and forwarddiff, so the agreement is quite low, like these on master 
https://github.com/SciML/SciMLSensitivity.jl/blob/a22e87cf6b5ecb6cee5c2c0ce0a93be3099b0d76/test/adjoint.jl#L332-L334
https://github.com/SciML/SciMLSensitivity.jl/blob/a22e87cf6b5ecb6cee5c2c0ce0a93be3099b0d76/test/adjoint.jl#L596-L597

Abhishek-1Bhatt:
For Lotka Volterra the issue is that making a function like 
```julia
function lotka(u, p, t)
    du1 = p[1]*u[1] - p[2]*u[1]*u[2]
    du2 = -p[3]*u[2] + p[4]*u[1]*u[2]
    SVector(du1, du2)
end
```
doesn't go well with Zygote when calling 
```julia
du0, dp = adjoint_sensitivities(sol, Tsit5(); t = tsteps, dgdu_discrete = dg_disc,
                                sensealg = QuadratureAdjoint(autojacvec = ZygoteVJP()))
```
on the solve as constructing an SVector as a return doesn't have an adjoint and throws
```
ERROR: Need an adjoint for constructor SVector{2, Float32}. Gradient is of type SVector{2, Float32}
```
So, I had to make a function which returns an SVector as output as a result of the computation
```julia
u0 = @SVector [1.0f0, 1.0f0]
p = @SMatrix [1.5f0 -1.0f0; 3.0f0 -1.0f0]

function f(u, p, t)
    p * u
end
```

ChrisRackauckas:
> ERROR: Need an adjoint for constructor SVector{2, Float32}. Gradient is of type SVector{2, Float32}

Add a fix for that in ChainRules?

test/adjoint_oop.jl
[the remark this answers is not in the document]
ChrisRackauckas: Thanks for pointing out that issue. I ran it locally and those match to 1e-10, so added here: https://github.com/SciML/SciMLSensitivity.jl/pull/691 . But I don't see how that would justify a high tolerance here?

ChrisRackauckas:
Yes

ChrisRackauckas:
Is this being finished? What's left?

Abhishek-1Bhatt:
The StaticArrays AD stuff is almost done here https://github.com/JuliaArrays/StaticArrays.jl/pull/1068 , hopefully it will get merged.

ChrisRackauckas:
That error is https://github.com/SciML/SciMLSensitivity.jl/issues/707 which is hopefully fixed in https://github.com/SciML/SciMLSensitivity.jl/pull/706. I'll rebase after that fix is merged and merge if tests pass.

PR #684, opened 2022-07-08
use adjoints from SteadyStateProblems also for NonlinearProblems

frankschae:
Converts a `NonlinearFunction` to an `ODEfunction` and `SteadyStateProblem` to `Union{SteadyStateProblem, NonlinearProblem}` to use steady-state adjoint definitions also for nonlinear solvers.

requires: 
https://github.com/SciML/NonlinearSolve.jl/pull/77
https://github.com/SciML/DiffEqBase.jl/pull/793
https://github.com/SciML/SciMLBase.jl/pull/202


ChrisRackauckas:
Bump the lower bounds when bumping for the tests. Other than that it looks good.

PR #688, opened 2022-07-16
Test master

ChrisRackauckas:


ChrisRackauckas:
The KLU thing is https://github.com/SciML/LinearSolve.jl/pull/157

PR #695, opened 2022-07-25
Allow mixing Enzyme with Rodas via PreallocationTools

ChrisRackauckas:
Should fix https://github.com/xzackli/Bolt.jl/issues/59


src/sensitivity_algorithms.jl
```diff
@@ -1021,10 +1021,24 @@ like BLAS/LAPACK are used) and this will be the most efficient adjoint implement
...
+  the `DualCaches` for PreallocationTools.jl. It defaults to 12, the maximum default
+  chunksize of ForwardDiff.jl, for safety, but could be decreased if this value is known to
+  be lower to conserve memory. Future optimizations can make this more automatic.
 """
-struct EnzymeVJP <: VJPChoice end
+struct EnzymeVJP <: VJPChoice
+    chunksize::Int
+end
+
+EnzymeVJP(; chunksize = 12) = EnzymeVJP(12)
```
prbzrg, line 1041: Shouldn't it be:
```suggestion
EnzymeVJP(; chunksize = 12) = EnzymeVJP(chunksize)
```
?

ChrisRackauckas: yes, good catch

ChrisRackauckas: Fixed on master

Issue #714, opened 2022-08-23
Allow provided jacobians with DAE adjoint


taylormcd:
The adjoint solve fails when the user provides their own jacobian.  This is applicable when solving for the sensitivities of semi-explicit DAEs.  If you look at derivative_wrappers.jl, there is no implementation for custom jacobians, only implementations for ForwardDiff and FiniteDiff.

taylormcd:
The line of the code that needs to change is here: https://github.com/SciML/SciMLSensitivity.jl/blob/b4fd1ce3c0cfa1db744f71319875cf7c8b2bd8d1/src/adjoint_common.jl#L469
I think the best solution is to add an if statement like the the one in the steady state adjoint: https://github.com/SciML/SciMLSensitivity.jl/blob/b4fd1ce3c0cfa1db744f71319875cf7c8b2bd8d1/src/steadystate_adjoint.jl#L91-L107
I can probably put together a basic pull request with this.

taylormcd:
Here's a MWE based on the semi-explicit version of the pendulum problem:

```
using OrdinaryDiffEq, LinearAlgebra, SciMLSensitivity, Zygote

function f!(du, u, p, t)
    x, y, vx, vy, T = u
    g, L = p
    du[1] = vx
    du[2] = x*T
    du[3] = vy
    du[4] = y*T - g
    du[5] = 2*(vx^2 + vy^2 + y*(y*T - g) + T*x^2)
    return nothing
end

function jac!(J, u, p, t)
    x, y, vx, vy, T = u
    g, L = p
    J .= 0
    J[1,3] = 1
    J[2,1] = T
    J[2,5] = x
    J[3,4] = 1
    J[4,2] = T
    J[4,5] = y
    J[5,1] = 4*T*x
    J[5,2] = 2*(2*y*T - g)
    J[5,3] = 4*vx
    J[5,4] = 4*vy
    J[5,5] = 2*(y^2 + x^2)
    return nothing
end

func = ODEFunction(f!, mass_matrix=collect(Diagonal([1,1,1,1,0])), jac=jac!)
u0 = [1.0, 0, 0, 0, 0]; p = [9.8, 1]; tspan = (0, 10.0)
prob = ODEProblem(func, u0, tspan, p)
solve(prob,Rodas5())

function sum_of_solution(p)
    _prob = remake(prob,p=p)
[...]

PR #717, opened 2022-08-24
Allow custom jacobians with DAE adjoints

taylormcd:
See #714

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -469,8 +469,11 @@ function (f::ReverseLossCallback)(integrator)
         if J isa DiffCache
             J = get_tmp(J, y)
         end
-
-        jacobian!(J, uf, y, f_cache, sensealg, jac_config)
+        if DiffEqBase.has_jac(f)
+            f.jac(J, y, p, t)
+        else 
+            jacobian!(J, uf, y, f_cache, sensealg, jac_config)
+        end
         dhdd = J[algevar_idxs, diffvar_idxs]
         dhda = J[algevar_idxs, algevar_idxs]
         # TODO: maybe need a `conj`
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -473,7 +473,7 @@ function (f::ReverseLossCallback)(integrator)
             J = get_tmp(J, y)
         end
         if DiffEqBase.has_jac(f.f)
-            f.f.jac(J, y, p, t)
+            f.f.jac(J, y, p, t[cur_time[]])
         else 
             jacobian!(J, uf, y, f_cache, sensealg, jac_config)
         end
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -405,7 +405,7 @@ inplace_sensitivity(S::SensitivityFunction) = isinplace(getprob(S))
 
 struct ReverseLossCallback{λType, timeType, yType, RefType, FMType, AlgType, dg1Type,
                            dg2Type,
-                           cacheType, solType}
+                           cacheType, fType, solType}
     isq::Bool
     λ::λType
     t::timeType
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Rebased onto current master for v1.8 test fixes.

ChrisRackauckas (approved):

src/adjoint_common.jl
```diff
@@ -427,13 +427,8 @@ function ReverseLossCallback(sensefun, λ, t, dgdu, dgdp, cur_time)
...
     idx = length(prob.u0)
-    if ArrayInterfaceCore.ismutable(y)
-        return ReverseLossCallback(isq, λ, t, y, cur_time, idx, factorized_mass_matrix,
-                                   sensealg, dgdu, dgdp, sensefun.diffcache, nothing)
-    else
-        return ReverseLossCallback(isq, λ, t, y, cur_time, idx, factorized_mass_matrix,
-                                   sensealg, dgdu, dgdp, sensefun.diffcache, sensefun.sol)
-    end
+    return ReverseLossCallback(isq, λ, t, y, cur_time, idx, factorized_mass_matrix,
+                                sensealg, dgdu, dgdp, sensefun.diffcache, sensefun.sol)
```
ChrisRackauckas, line 431: Looks like this slid by because BacksolveAdjoint is not used with DAEs (for stability reasons), but if this is always checked now then it needs to be more exact.

ChrisRackauckas, line 431:
```suggestion
                                sensealg, dgdu, dgdp, sensefun.diffcache, sensefun.f)
```

src/adjoint_common.jl
```diff
@@ -469,8 +464,11 @@ function (f::ReverseLossCallback)(integrator)
         if J isa DiffCache
             J = get_tmp(J, y)
         end
-
-        jacobian!(J, uf, y, f_cache, sensealg, jac_config)
+        if DiffEqBase.has_jac(sol.prob.f)
+            sol.prob.f.jac(J, y, p, t)
```
ChrisRackauckas, line 468:
```suggestion
        if DiffEqBase.has_jac(f.f)
            f.f.jac(J, y, p, t)
```

src/adjoint_common.jl
```diff
@@ -427,13 +427,8 @@ function ReverseLossCallback(sensefun, λ, t, dgdu, dgdp, cur_time)
...
-                                   sensealg, dgdu, dgdp, sensefun.diffcache, nothing)
-    else
-        return ReverseLossCallback(isq, λ, t, y, cur_time, idx, factorized_mass_matrix,
-                                   sensealg, dgdu, dgdp, sensefun.diffcache, sensefun.sol)
-    end
+    return ReverseLossCallback(isq, λ, t, y, cur_time, idx, factorized_mass_matrix,
+                                sensealg, dgdu, dgdp, sensefun.diffcache, sensefun.sol)
 end
 
 function (f::ReverseLossCallback)(integrator)
```
ChrisRackauckas, line 434: I cannot edit suggest, but the `sol` down there should change.

src/adjoint_common.jl
ChrisRackauckas: oh wait hmm, some cases need `sol`. This should be algorithm dependent, and the `f` should be kept separate.

taylormcd:
The latest commit treats `f` and `sol` separately, it also reintroduces the `if` statement that was previously there for determining whether `sol` should be passed in..  I believe this should address your comments.

ChrisRackauckas:
Just needs the formatter to be ran and this is good to go. The DiffEqFlux failure is a known Zygote thing (with an incoming patch). 

https://github.com/SciML/SciMLStyle#juliaformatter

ChrisRackauckas:
I rebased and formatted. After tests pass this should be good to go.

ChrisRackauckas:
Beautiful all green, I like it.

PR #719, opened 2022-08-24
fix tag error for fallback case, fixes #718

### LINKED ISSUE #718: ForwardDiff tag error when computing Jacobian for DAE adjoint
Here's a MWE:

```julia
using OrdinaryDiffEq, LinearAlgebra, SciMLSensitivity, Zygote

function f!(du, u, p, t)
    x, y, vx, vy, T = u
    g, L = p
    du[1] = vx
    du[2] = x*T
    du[3] = vy
    du[4] = y*T - g
    du[5] = 2*(vx^2 + vy^2 + y*(y*T - g) + T*x^2)
    return nothing
end

func = ODEFunction(f!, mass_matrix=collect(Diagonal([1,1,1,1,0])))
u0 = [1.0, 0, 0, 0, 0]; p = [9.8, 1]; tspan = (0, 10.0)
prob = ODEProblem(func, u0, tspan, p)
solve(prob,Rodas5())

function sum_of_solution(p)
    _prob = remake(prob,p=p)
    sum(solve(_prob,Rodas5(),saveat=0.1,sensealg=QuadratureAdjoint(autojacvec=false)))
end

dp = Zygote.gradient(sum_of_solution,p)

ERROR: Invalid Tag object:
  Expected ForwardDiff.Tag{SciMLBase.UJacobianWrapper{ODEFunction{true, typeof(f!), Matrix{Int64}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED), Nothing, Nothing}, ForwardDiff.Dual{ForwardDiff.Tag{DiffEqBase.OrdinaryDiffEqTag, Float64}, Float64, 1}, Vector{Float64}}, ForwardDiff.Dual{ForwardDiff.Tag{DiffEqBase.OrdinaryDiffEqTag, Float64}, Float64, 1}},
  Observed ForwardDiff.Tag{SciMLBase.UJacobianWrapper{ODEFunction{true, typeof(f!), Matrix{Int64}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED), Nothing, Nothing}, Float64, Vector{Float64}}, ForwardDiff.Dual{ForwardDiff.Tag{DiffEqBase.OrdinaryDiffEqTag, Float64}, Float64, 1}}.
```

taylormcd:
This fix lets ForwardDiff manage its own cache for this fallback case.  Since the cache is only used once anyway, I don't think performance will be adversely affected compared to the previous code.

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -253,7 +253,9 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::Bool, dgrad, dy,
 
                     # This is really slow and allocates, but it's a fallback only for a
                     # rare case so it can be optimized in the future
-                    jacobian!(J, _uf, y, nothing, sensealg, nothing)
+                    _f_cache = DiffEqBase.isinplace(prob) ? deepcopy(y) : nothing
+                    _jac_config = build_jac_config(sensealg, _uf, t * dλ)
+                    jacobian!(J, _uf, y, _f_cache, sensealg, _jac_config)
                 else
                     uf.t = t
                     uf.p = p
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -253,8 +253,7 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::Bool, dgrad, dy,
 
                     # This is really slow and allocates, but it's a fallback only for a
                     # rare case so it can be optimized in the future
-                    _jac_config = build_jac_config(sensealg, uf, t * dλ)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
This needs to pass down an `fdtype` if it's going to get an epsilon.

taylormcd:
I added a commit which closely resemble the original implementation, but fixes the types.  I believe this addresses your comment.

PR #723, opened 2022-08-27
Badges

ArnoStrouwen:
Buildkite badge still needs work but I don't have permissions there.
I can't verify if it links to a static image, and in any case the link to builds is wrong.


ChrisRackauckas:
Interesting, stochastic build failure?

PR #729, opened 2022-09-01
Fix adjoint sensitivity for parameter dependent algebraic constraints

YingboMa:
Fix https://github.com/SciML/SciMLSensitivity.jl/issues/686

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -231,30 +231,6 @@ function vecjacobian(y, λ, p, t, S::TS;
     return _vecjacobian(y, λ, p, t, S, S.sensealg.autojacvec, dgrad, dy, W)
 end
 
-function vecpjacobian!(out, y, λ, p, t, S::TS) where {TS <: SensitivityFunction}
-    return _vecpjacobian!(out, y, λ, p, t, S)
-end
-
-# TODO: actual v'pJ via AD
-function _vecpjacobian!(out, y, λ, p, t, S::TS) where {TS <: SensitivityFunction}
-    @unpack sensealg, f = S
-    @unpack f_cache, pJ, pf, paramjac_config = S.diffcache
-    if DiffEqBase.has_paramjac(f)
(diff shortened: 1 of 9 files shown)


ChrisRackauckas (approved):
The val=true is a little messy, but agreed it's not worth the work to break the interface right now so this is good.

src/derivative_wrappers.jl
```diff
@@ -238,11 +238,11 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::Bool, dgrad, dy,
 
     @unpack J, uf, f_cache, jac_config = S.diffcache
 
-    if J isa DiffCache
+    if J isa DiffCache && dλ !== nothing
```
frankschae, line 241: I removed the additional λ' J_p  definitions
```julia 
function vecpjacobian!(out, y, λ, p, t, S::TS) where {TS<:SensitivityFunction}
    return _vecpjacobian!(out, y, λ, p, t, S, S.sensealg.autojacvec)
end

# TODO: actual v'pJ via AD
function _vecpjacobian!(out, y, λ, p, t, S::TS, isautojacvec::Bool) where {TS<:SensitivityFunction}
    @unpack sensealg, f = S
    @unpack f_cache, pJ, pf, paramjac_config = S.diffcache
    if DiffEqBase.has_paramjac(f)
        # Calculate the parameter Jacobian into pJ
        f.paramjac(pJ, y, p, t)
    else
        pf.t = t
        pf.u = y
        if inplace_sensitivity(S)
            jacobian!(pJ, pf, p, f_cache, sensealg, paramjac_config)
        else
            temp = jacobian(pf, p, sensealg)
            pJ .= temp
        end
    end
    mul!(out', λ', pJ)
end
```
in favor of handling it directly with the established `_vecjacobian!` functions (significantly less code with very little numerical overhead.)

raphaelchinchilla:
@YingboMa, I heard it was a pretty hard issue to fix. Thanks a lot :)

PR #730, opened 2022-09-12
Move LazyBuffer ReverseDiff dispatch to SciMLSensitivity

frankschae:
Moving fixes for LazyBuffer from https://github.com/SciML/PreallocationTools.jl/pull/30/files to this package, see https://github.com/SciML/PreallocationTools.jl/issues/39.


ChrisRackauckas:
Bump the lower bound on PreallocationTools so it won't throw warnings?

frankschae:
don't we first need to fix, merge, and tag https://github.com/SciML/PreallocationTools.jl/pull/40?

PR #731, opened 2022-09-13
Fix for non-diagonal, non-square SDEs

frankschae:
The dimension of the Brownian process `m` remains the same in the adjoint. (Until now, the implementation assumed `d=m` so that index errors were thrown otherwise.)


ChrisRackauckas:
This needs a test

PR #732, opened 2022-09-20
Support `ReverseDiffVJP(true)` for hybrid DEs

frankschae:
This PR adds the kwarg `recompile_tape = true` to `vecjacobian!` to manually enforce a recompilation of the tape in the vjp computation of the event handling where a different/new tape must be generated. The original compiled tape can then still be used for the remaining vjps.

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -206,28 +206,40 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback, DiscreteCallback
 
     cb.save_positions == [1, 0] && error("save_positions=[1,0] is currently not supported.")
 
+    # event times
+    times = if typeof(cb) <: DiscreteCallback
+        cb.affect!.event_times
+    else
+        [cb.affect!.event_times; cb.affect_neg!.event_times]
+    end
+
+    # precompute w and wp to generate a tape
+    if cb isa DiscreteCallback
+        w, wp = setup_w_wp(cb, true, nothing, nothing)
+    elseif cb isa ContinuousCallback
+        # compile a tape for pos and neg
+        w, wp = [setup_w_wp(cb, pos_neg, nothing, nothing) for pos_neg in (true, false)]
+    else
+        # cb isa VectorContinuousCallback
+        # compile a tape for pos/neg and all event indices
+        w, wp = [setup_w_wp(cb, pos_neg, event_idx, nothing) for pos_neg in (true, false)
+                 for event_idx in 1:(cb.len)]
+    end
+
     function affect!(integrator)
         indx, pos_neg = get_indx(cb, integrator.t)
         tprev = get_tprev(cb, indx, pos_neg)
         event_idx = cb isa VectorContinuousCallback ? get_event_idx(cb, indx, pos_neg) :
                     nothing
 
-        w = let tprev = tprev, pos_neg = pos_neg, event_idx = event_idx
-            function (du, u, p, t)
-                _affect! = get_affect!(cb, pos_neg)
-                fakeinteg = FakeIntegrator([x for x in u], [x for x in p], t, tprev)
-                if cb isa VectorContinuousCallback
-                    _affect!(fakeinteg, event_idx)
-                else
-                    _affect!(fakeinteg)
-                end
-                du .= fakeinteg.u
-            end
-        end
+        # how can we make sure tprev didn't change, so precompiled tape is correct?
+        w, wp = setup_w_wp(cb, pos_neg, event_idx, tprev)
+
         S = integrator.f.f # get the sensitivity function
 
         # Create a fake sensitivity function to do the vjps
         @assert sensealg.autojacvec isa ReverseDiffVJP
+        # update diffcache here to use the correct precompiled callback tape?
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

src/derivative_wrappers.jl
```diff
@@ -220,15 +220,15 @@ function jacobianmat!(JM::AbstractMatrix{<:Number}, f, x::AbstractArray{<:Number
 end
 function vecjacobian!(dλ, y, λ, p, t, S::TS;
                       dgrad = nothing, dy = nothing,
-                      W = nothing) where {TS <: SensitivityFunction}
-    _vecjacobian!(dλ, y, λ, p, t, S, S.sensealg.autojacvec, dgrad, dy, W)
+                      W = nothing, kwargs...) where {TS <: SensitivityFunction}
```
ChrisRackauckas, line 223: we should probably be more specific than a full splat

frankschae: How would I do that? By adding the `recompile_tape` kwarg to all `_vecjacobian!` functions even if it's not used except in ReverseDiffVJP?

src/derivative_wrappers.jl
```diff
@@ -431,7 +431,7 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::ReverseDiffVJP, dg
 
     if prob isa Union{SteadyStateProblem, NonlinearProblem} ||
        (eltype(λ) <: eltype(prob.u0) && typeof(t) <: eltype(prob.u0) &&
-        compile_tape(sensealg.autojacvec))
+        compile_tape(sensealg.autojacvec) && !recompile_tape)
```
ChrisRackauckas, line 434: why here?

src/callback_tracking.jl
```diff
@@ -297,13 +298,13 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback, DiscreteCallback
                                                      integrator.sol.prob)
                 #vjp with Jacobin given by dw/dp before event and vector given by grad
                 vecjacobian!(dgrad, integrator.p, grad, y, integrator.t, fakeSp;
-                             dgrad = nothing, dy = nothing)
+                             dgrad = nothing, dy = nothing, recompile_tape = true)
```
ChrisRackauckas, line 301: I still don't quite get why this is necessary. What branch is causing an issue here?

src/derivative_wrappers.jl
frankschae: this is the place where we check if the compiled tape should be used. So I added the additional boolean to avoid using the old tape and build a new one when needed for the callback handling.

src/callback_tracking.jl
frankschae: it's needed because we compute vjps in the correction with respect to different functions (the affect function instead of the ODE function). So the tape precompiled for the ODE function will not work for instance when called from
https://github.com/SciML/SciMLSensitivity.jl/blob/68200c5e4ca60b3cbd2f1fd3ed76c9a380f71942/src/callback_tracking.jl#L299
for the function
https://github.com/SciML/SciMLSensitivity.jl/blob/68200c5e4ca60b3cbd2f1fd3ed76c9a380f71942/src/callback_tracking.jl#L284

ChrisRackauckas: Can we just precompile a few tapes? One for each posneg, for each affect? Usually there's only a handful (vast majority one, almost never more than two) callbacks, so this isn't a huge amount.

frankschae: Sounds like a good idea. It's a little more complicated to implement because I think we have to pull the function construction/precompilation of the tape out of the affect function. I will give it a try.

ChrisRackauckas:
What's the state here?

src/callback_tracking.jl
```diff
@@ -206,27 +206,40 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback, DiscreteCallback
...
+        w, wp = setup_w_wp(cb, true, nothing, nothing)
+    elseif cb isa ContinuousCallback
+        # compile a tape for pos and neg
+        w, wp = [setup_w_wp(cb, pos_neg, nothing, nothing) for pos_neg in (true, false)]
+    else
+        # cb isa VectorContinuousCallback
+        # compile a tape for pos/neg and all event indices
+        w, wp = [setup_w_wp(cb, pos_neg, event_idx, nothing) for pos_neg in (true, false)
+                 for event_idx in 1:(cb.len)]
+    end
```
frankschae, line 227: Here's the part where I am still unsure how to do it in a good way. Should we try to re-use the tools from `adjointdiffcache` to build the tape for the callbacks, or is it better to generate only new ReverseDiff tapes here? (I would hope that if I do the former correctly that we can also support Enzyme)

ChrisRackauckas: Maybe the first? Though that might be trying too hard.

ChrisRackauckas:
Whatever happened with this?

frankschae:
Got pushed back because of conferences and travel. I hope to look into this again during the Christmas break.

frankschae:
rebased

ChrisRackauckas:
Awesome! This should be a major performance win for a lot of people.

PR #747, opened 2022-10-27
Filtering of time steps for `ForwardDiffSensitivity` with Callbacks

frankschae:
See the discourse thread:
https://discourse.julialang.org/t/diffeq-continuouscallback-not-callable-errors/88198

and previous PRs
https://github.com/SciML/SciMLSensitivity.jl/pull/690
https://github.com/SciML/SciMLSensitivity.jl/pull/436

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -629,6 +629,9 @@ function DiffEqBase._concrete_solve_adjoint(prob::Union{SciMLBase.AbstractODEPro
     sol = solve(remake(prob, p = p, u0 = u0), alg, args...; saveat = _saveat, kwargs...)
 
     # saveat values
+    # need all values here. Not only unique ones.
+    # if a callback is saving two times in primal solution, we also need to get it at least
+    # two times in the solution using dual numbers.
     ts = sol.t
 
     function forward_sensitivity_backpass(Δ)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Can you add nefarious MWEs to the tests?

frankschae:
I added one more (I'll think about a few more). In the newly added example, we store both values of a terminating callback in the primal but throw one of them away afterward in the loss function. (This didn't work before with InterpolatingAdjoint because we would miss the end time in the automated selection of the checkpoints.) Using dual numbers, we do not hit this time point. So both have to be kept and used as saveat, and the superficial points at different times have to be filtered. 

```julia
cb_t = DiscreteCallback(condition, affect!, save_positions = (true, true)
function loss(u0, p; sensealg = nothing, save_start = false, save_everystep = false)
            _prob = remake(prob, u0 = u0, p = p)
            # saving arguments can have a huge influence here
            sol = solve(_prob, Tsit5(), reltol = tol, abstol = tol, sensealg = sensealg,
                        callback = cb_t,
                        save_start = save_start, save_everystep = save_everystep)
            res = sol.u[end]
            g(res, p, nothing)
end
```
3 saved time points occur when saving callbacks to points that are also saved otherwise. This happens in the HybridNODE examples
https://github.com/SciML/SciMLSensitivity.jl/blob/master/test/HybridNODE.jl#L23
For these cases, the second step is necessary.

ChrisRackauckas:
I think this looks good. Merge when you think it's done.

PR #762, opened 2023-01-07
Support Enzyme and ReverseDiff with compiled tape for adjoint sensitivity methods with callbacks

frankschae:
Follow up for https://github.com/SciML/SciMLSensitivity.jl/pull/732 and try to solve https://github.com/SciML/SciMLSensitivity.jl/issues/630.

This PR slightly refactors the `adjointdiffcache` function, so that `pf` and `paramjac_config` for `ReverseDiffVJP` and `EnzymeVJP` can be recreated for the callback affect functions (`w` and `wp`) (i.e. the functions that create `pf` and `paramjac_config` for the `DEfunction` `f` are re-used in the callback handling for `w` and `wp`). 

The diffcaches are stored in a dictionary and then looked up in the `affect!` function
```julia
diffcaches = cb_diffcaches[pos_neg, event_idx]
``` 

No more need for `kwarg` splatting and the recompile kwarg!


src/callback_tracking.jl
```diff
@@ -205,6 +205,7 @@ function _setup_reverse_callbacks(cb::Union{ContinuousCallback, DiscreteCallback
     # if save_positions = [1,0] the gradient contribution is added before, and in principle we would need to correct the adjoint state again. Thefore,
 
     cb.save_positions == [1, 0] && error("save_positions=[1,0] is currently not supported.")
+    @assert sensealg.autojacvec isa Union{ReverseDiffVJP, EnzymeVJP}
```
ChrisRackauckas, line 208: Instead of an assertion, it would be good to throw an informative error that only those two are allowed.

ChrisRackauckas:
This looks much cleaner! I like the direction this went.

frankschae:
Do you have an idea regarding the Enzyme segfault errors? Is this just a strange version thing? 

For example, the one thrown in https://github.com/SciML/SciMLSensitivity.jl/blob/master/test/complex_adjoints.jl#L59-L72
can be solved by adding `nothing` to the ODE function
```julia
function fiip(du, u, p, t)
    du[1] = dx = p[1] * u[1] - p[2] * u[1] * u[2]
    du[2] = dy = -p[3] * u[2] + p[4] * u[1] * u[2]
    nothing
end
```
I actually get this error locally also on the main branch.

My manifest shows:
```
[[deps.Enzyme]]
deps = ["CEnum", "EnzymeCore", "Enzyme_jll", "GPUCompiler", "LLVM", "Libdl", "LinearAlgebra", "ObjectFile", "Printf", "Random"]
git-tree-sha1 = "77c9ef49f69af89386ddba92760293bf76f78305"
uuid = "7da242da-08ed-463a-9acd-ee780be4f1d9"
version = "0.10.13"

[[deps.EnzymeCore]]
deps = ["Adapt"]
git-tree-sha1 = "238032b8e2a02e06bc8e257ff9484a96db8fea1b"
uuid = "f151be2c-9106-41f4-ab19-57ee4f262869"
version = "0.1.0"

[[deps.Enzyme_jll]]
deps = ["Artifacts", "JLLWrappers", "LazyArtifacts", "Libdl", "Pkg", "TOML"]
git-tree-sha1 = "bb013ec867173e3a710005374455818eb04afb81"
uuid = "7cc45869-7501-5eee-bdea-0790c847d4ef"
version = "0.0.47+0"
```

frankschae:
Interesting - the failing SDE test is for an in-place `SDEProblem` test that previously worked with `ZygoteVJP`. 
```julia
res_sde_u0a, res_sde_pa = adjoint_sensitivities(sol, EulerHeun(), t=Array(t),
    dgdu_discrete=dg!,
    dt=dtnd, adaptive=false,
    sensealg=BacksolveAdjoint(autojacvec=ZygoteVJP()))
```

ChrisRackauckas:
The SDE failure seems to be on master.Maybe it's from a Zygote change?

ChrisRackauckas:
It's just one test right? Let's mark it as `@test_broken` and get an MWE to Zygote.jl. Sounds like it's related to support of Buffers which isn't the most widely used feature.

ChrisRackauckas:
https://github.com/SciML/SciMLSensitivity.jl/pull/766 is the proof of the Zygote regression.

frankschae:
I think it's one for `BacksolveAdjoint` and one for `InterpolatingAdjoint` (but yes, only for in-place functions using `ZygoteBuffer`).

ChrisRackauckas:
I marked them both as `@test_broken`. Can you try and make the MWE? Then I think we're done with this for now.

test/sde_nondiag_stratonovich.jl
```diff
@@ -154,24 +154,26 @@ end
...
-    @test isapprox(res_sde_u0a, res_sde_u0, rtol = 1e-6)
-    @test isapprox(res_sde_pa, res_sde_p, rtol = 1e-6)
+    
+    @test_broken begin
+        res_sde_u0a, res_sde_pa = adjoint_sensitivities(sol, EulerHeun(), t = Array(t),
+                                                        dgdu_discrete = dg!,
+                                                        dt = dtnd, adaptive = false,
+                                                        sensealg = BacksolveAdjoint(autojacvec = ZygoteVJP()))
+
+        @test all(isapprox(res_sde_u0a, res_sde_u0, rtol = 1e-6),isapprox(res_sde_pa, res_sde_p, rtol = 1e-6))
```
ChrisRackauckas, line 164:
```suggestion
        @test all((isapprox(res_sde_u0a, res_sde_u0, rtol = 1e-6),isapprox(res_sde_pa, res_sde_p, rtol = 1e-6)))
```

frankschae:
The MWE is here: https://github.com/FluxML/Zygote.jl/issues/1352

ChrisRackauckas:
The docs failure looks spurious, but we should follow up and make sure it actually does build and is just a neural SDE that initialized bad.

PR #769, opened 2023-01-17
`solve_up` implementation for array of `ReverseDiff.TrackedReal`

torfjelde:
Makes it possible to have initial state `u0`and/or parameters `p` be `AbstractArray{<:ReverseDiff.TrackedReal}`, which then calls the `ReverseDiff.TrackedArray` version.

### DIFF
--- a/src/reversediff.jl
+++ b/src/reversediff.jl
@@ -57,6 +57,7 @@ end
     abs(u)
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

torfjelde:
I realized I forgot to bump the version. Will you do that @ChrisRackauckas or should I make another PR?

PR #770, opened 2023-01-17
Test SimpleNonlinearSolve's compatibility with adjoints

ChrisRackauckas:


ChrisRackauckas:
@avik-pal I think with this, there's a Broyden, Klement, etc. I think we can drop a lot of stuff from DeepEquilibriumNetworks.jl

avik-pal:
Awesome, I will take a closer look in early february

PR #780, opened 2023-02-06
Remove AD weak deps

ChrisRackauckas:
Requires:

* https://github.com/SciML/PreallocationTools.jl/pull/58
* https://github.com/SciML/RecursiveArrayTools.jl/pull/247
* https://github.com/SciML/DiffEqBase.jl/pull/871

Moving all of those to their rightful places instead of pirating now that a weak dependency system exists.


src/SciMLSensitivity.jl
```diff
@@ -32,6 +32,12 @@ abstract type TransformedFunction end
 
 import SciMLBase: unwrapped_f
 
+abstract type AbstractSensitivityAlgorithm{CS, AD, FDT} <: AbstractSciMLAlgorithm end
+
```
ChrisRackauckas, line 36:
```suggestion
```

PR #786, opened 2023-02-24
Isolate broken tests

ChrisRackauckas:


ChrisRackauckas:
@frankschae can you look at the terminate callback ones? I am not entirely sure how those were supposed to be working. They flip a problem from (0, Inf) to (Inf, 0) but then fail because they start at Inf. Is there supposed to be a mechanism that flips it to start at the terminating point? How is that terminating point saved?

frankschae:
Yup, exactly. It's in the construction of the `ODEAdjointProblem`
https://github.com/SciML/SciMLSensitivity.jl/blob/master/src/backsolve_adjoint.jl#L148-L154

ChrisRackauckas:
Awesome, and that explains the broken tests.

test/steady_state.jl
```diff
@@ -485,7 +471,7 @@ end
                                     u0, p)
         @test du0≈Zdu0 atol=1e-4
         @test dp≈Zdp atol=1e-4
-        Zdu0, Zdp = Zygote.gradient((u0, p) -> loss(u0, p, sensealg = BacksolveAdjoint()),
+        Zdu0, Zdp = Zygote.gradient((u0, p) -> loss(u0, p, sensealg = BacksolveAdjoint(autojacvec = EnzymeVJP())),
```
ChrisRackauckas, line 474:
```suggestion
        Zdu0, Zdp = Zygote.gradient((u0, p) -> loss(u0, p, sensealg = BacksolveAdjoint()),
```

PR #791, opened 2023-03-02
Fixes to SteadyState / Nonlinear Adjoint

avik-pal:


src/steadystate_adjoint.jl
```diff
@@ -188,3 +148,37 @@ end
...
+                    y, reshape(u, size(y)))
+        return du
+    end
+    return FunctionOperator(f_operator!, vec(y); p)
+end
+
+# LinearSolve tries to deepcopy the operator
+# Having a Zygote Pullback causes `deepcopy` to error with
+# `deepcopy of Modules not supported`.
+# This is a hacky workaround for that
```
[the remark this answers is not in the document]
avik-pal: done

src/steadystate_adjoint.jl
```diff
@@ -126,42 +103,27 @@ end
...
+        operator = FixedVecJacOperator(f, y, p, Val(DiffEqBase.isinplace(sol.prob)))
+        linear_problem = LinearProblem(operator, vec(dgdu_val); u0 = vec(λ))
     else
-        linear_problem = LinearProblem(diffcache.J', vec(dgdu_val'), u0 = vec(λ))
+        linear_problem = LinearProblem(diffcache.J', vec(dgdu_val'); u0 = vec(λ))
     end
 
-    solve(linear_problem, linsolve) # u is vec(λ)
+    # Zygote pullback function won't work with deepcopy
+    solve(linear_problem, linsolve; alias_A = !needs_jac) # u is vec(λ)
```
ChrisRackauckas, line 123: why not just true?

avik-pal: I was not sure why the default is `false`. So wanted to keep it that for cases where the default works

ChrisRackauckas: It's a safety measure if the `A` is reused, but it's not here.

ChrisRackauckas (approved):

PR #801, opened 2023-03-17
Fix for SciMLSensitivity.jl issue #564 and #809

ThummeTo:
Please see the discussion in [this PR](https://github.com/SciML/SciMLSensitivity.jl/pull/801) concerning [this issue](https://github.com/SciML/SciMLSensitivity.jl/issues/809) for further information.

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -60,8 +60,6 @@ struct TrackedAffect{T, T2, T3, T4, T5, T6}
     affect!::T4
     correction::T5
     event_idx::Vector{T6}
-    func_start::Bool
-    funciter::Int
 end
 
 TrackedAffect(t::Number, u, p, affect!::Nothing, correction) = nothing
@@ -69,9 +67,7 @@ function TrackedAffect(t::Number, u, p, affect!, correction)
     TrackedAffect(Vector{typeof(t)}(undef, 0), Vector{typeof(t)}(undef, 0),
                   Vector{typeof(u)}(undef, 0), Vector{typeof(p)}(undef, 0), affect!,
                   correction,
-                  Vector{Int}(undef, 0),
-                  true,
-                  0)
+                  Vector{Int}(undef, 0))
 end
 
 function (f::TrackedAffect)(integrator, event_idx = nothing)
@@ -105,12 +101,18 @@ function (f::TrackedAffect)(integrator, event_idx = nothing)
     end
 end
 
+# extend the DiffEqCallbacks.functioncalling_affect_initialize to fit TrackedAffect
+import DiffEqCallbacks: functioncalling_affect_initialize
+function functioncalling_affect_initialize(affect::TrackedAffect, u, t, integrator)
+    functioncalling_affect_initialize(affect.affect!, u, t, integrator)
+end
+
 function _track_callback(cb::DiscreteCallback, t, u, p, sensealg)
     correction = ImplicitCorrection(cb, t, u, p, sensealg)
     DiscreteCallback(cb.condition,
                      TrackedAffect(t, u, p, cb.affect!, correction),
-                     cb.initialize,
-                     cb.finalize,
+                     cb.initialize, 
+                     cb.finalize, 
                      cb.save_positions)
 end
 
@@ -119,8 +121,8 @@ function _track_callback(cb::ContinuousCallback, t, u, p, sensealg)
     ContinuousCallback(cb.condition,
                        TrackedAffect(t, u, p, cb.affect!, correction),
                        TrackedAffect(t, u, p, cb.affect_neg!, correction),
-                       cb.initialize,
-                       cb.finalize,
+                       cb.initialize, 
+                       cb.finalize, 
                        cb.idxs,
                        cb.rootfind, cb.interp_points,
                        cb.save_positions,
@@ -132,7 +134,10 @@ function _track_callback(cb::VectorContinuousCallback, t, u, p, sensealg)
     VectorContinuousCallback(cb.condition,
                              TrackedAffect(t, u, p, cb.affect!, correction),
                              TrackedAffect(t, u, p, cb.affect_neg!, correction),
-                             cb.len, cb.initialize, cb.finalize, cb.idxs,
+                             cb.len,
+                             cb.initialize, 
+                             cb.finalize, 
+                             cb.idxs,
                              cb.rootfind, cb.interp_points,
                              collect(cb.save_positions),
                              cb.dtrelax, cb.abstol, cb.reltol, cb.repeat_nudge)
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -111,8 +111,8 @@ function _track_callback(cb::DiscreteCallback, t, u, p, sensealg)
     correction = ImplicitCorrection(cb, t, u, p, sensealg)
     DiscreteCallback(cb.condition,
                      TrackedAffect(t, u, p, cb.affect!, correction),
-                     cb.initialize, 
-                     cb.finalize, 
+                     cb.initialize,
+                     cb.finalize,
                      cb.save_positions)
 end
 
@@ -121,8 +121,8 @@ function _track_callback(cb::ContinuousCallback, t, u, p, sensealg)
     ContinuousCallback(cb.condition,
                        TrackedAffect(t, u, p, cb.affect!, correction),
                        TrackedAffect(t, u, p, cb.affect_neg!, correction),
-                       cb.initialize, 
-                       cb.finalize, 
+                       cb.initialize,
+                       cb.finalize,
                        cb.idxs,
                        cb.rootfind, cb.interp_points,
                        cb.save_positions,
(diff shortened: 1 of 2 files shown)


src/callback_tracking.jl
```diff
@@ -101,6 +101,12 @@ function (f::TrackedAffect)(integrator, event_idx = nothing)
     end
 end
 
+# extend the DiffEqCallbacks.functioncalling_affect_initialize to fit TrackedAffect
+import DiffEqCallbacks: functioncalling_affect_initialize
+function functioncalling_affect_initialize(affect::TrackedAffect, u, t, integrator)
+    functioncalling_affect_initialize(affect.affect!, u, t, integrator)
+end
```
ChrisRackauckas, line 108: Why would this be specialized to just a single initialize? Why not directly handle initialize!?

ThummeTo: I see, using a user-defined custom initialization function will suffer from the same issue. I will change both PRs, so that every `initialize!` (and also `finalize` if I am already on it?) call is filtered to `FunctionCallingAffect` instead of `TrackedAffect`

ThummeTo: What would be the right location to accomplish that?
One way would be to add a dispatch [in the callbacks' constructors](https://github.com/SciML/SciMLBase.jl/blob/587cfcc9cf4277b35d3a0a2b08a21d5c9467bf8f/src/callbacks.jl#L94) and substitude the given function `initialize` (and `finalize` analogous) with `(c, u, t, integrator) -> initialize(c.affect!, u, t, integrator)`  if the constructor is called with `F2` = `typeof(affect!)`  = `TrackedAffect`?

These new constructor overloads could further be deployed in SciMLSensitivity next to the definition of the `TrackedAffect`.

ThummeTo: Hey @ChrisRackauckas,

I took a few minutes to think about this, but adding an additional dispatch for `initialize!` ([Code](https://github.com/SciML/DiffEqBase.jl/blob/e8c0a33c5a6be7d7bbbca190e7e0a14d5a5f7f1d/src/callbacks.jl#L16)) may not be the right way. The function interface of the callback initialization function is `(c::DECallback, u, t, integrator)`, so another dispatch that invokes the function on `c.affect!` would violate this interface. Initialization routines will be able to use `c.affect!` as to expect, but other attributes of `FunctionCallingAffect` are not part of the `TrackedAffect`, like for example `funciter`.

Curren timplementation, see [Code](https://github.com/SciML/DiffEqBase.jl/blob/e8c0a33c5a6be7d7bbbca190e7e0a14d5a5f7f1d/src/callbacks.jl#L16):
```
function initialize!(u, t, integrator::DEIntegrator, any_modified::Bool,
                     c::DECallback)
    c.initialize(c, u, t, integrator)
    any_modified || integrator.u_modified
end
```

As far as I understand, this would be the method to add (but it's bad!):
(see [Code](https://github.com/SciML/SciMLBase.jl/blob/f45dd60b50a9c207e9631356e887522d210a0d73/src/callbacks.jl#L293) for the DiscreteCallback implementation)
```
# new dispatch for discrete callbacks with `typeof(affect!)==TrackedAffect`
function initialize!(u, t, integrator::DEIntegrator, any_modified::Bool,
                     c::DiscreteCallback{Any, TrackedAffect, Any, Any}) 
    c.initialize(c.affect!, u, t, integrator) # we need to dive on layer into the TrackedAffect
    any_modified || integrator.u_modified
end
```

The problem here is, that the initialization routine is not performed on the callback itself, but on the `TrackedAffect`. So this will not work.

Because the core problem is that a TrackedAffect doesn't behave like a FunctionCallingAffect (because of the different fields), my suggestion would be to mimic the existence of the fields of the  FunctionCallingAffect inside of the TrackedAffect struct - this could easily be implemented overloading the `hasproperty`, `setproperty!` and `getproperty` functions. Another advantage of this is, that all changes would be inside of DiffEqCallbacks and other libraries are not affected.

What do you think?
I can rapidely deploy code if this is the way to go.

Best regards, 
Tobi

ChrisRackauckas: I think it may need to be a separate TrackedAffect. Thoughts @frankschae?

frankschae: Wouldn't it work if we extend the signature for `initialize!` to depend on both the callback and the `TrackedAffect`? (Not sure we really want to do it that way.... but it seems much better to handle different custom initializations as well.)

src/callback_tracking.jl
```diff
@@ -101,6 +101,12 @@ function (f::TrackedAffect)(integrator, event_idx = nothing)
...
+
 function _track_callback(cb::DiscreteCallback, t, u, p, sensealg)
     correction = ImplicitCorrection(cb, t, u, p, sensealg)
     DiscreteCallback(cb.condition,
```
frankschae, line 112: Can you maybe add an MWE as a test set for this and (if we manage to do it) a custom initialization?

frankschae: adding the MWE to https://github.com/SciML/SciMLSensitivity.jl/tree/master/test/callbacks, so it gets tested and we don't miss updating it if there are changes e.g. to `TrackedAffect!` in the future.

src/callback_tracking.jl
frankschae: yeah, this seems like a more general way to handle it to me.

src/callback_tracking.jl
ThummeTo: sure, no problem!

src/callback_tracking.jl
ThummeTo: I will make a quick suggestion, with `initialize!` depending on the callback and the (correct) `affect!`.

ThummeTo:
Dear @ChrisRackauckas and @frankschae,
as discussed, I made some minor modifications:

1. In DiffEqBase, the function signatures for the functions `initialize` and `finalize` on callbacks changed from `(cb, u, t, integrator)` to `(cb, affect!, u, t, integrator)` (theses functions are called in `initialize!` and `finalize!`), see this PR [WIP]. 
2. In SciMLSensitivity, the functions `initialize!` and `finalize!` are extended to handle generic callbacks using `TrackedAffect` correctly, see [this PR](https://github.com/SciML/SciMLSensitivity.jl/pull/801). Further, the existing test for Callbacks + ReverseDiff was extended (as suggested by @frankschae) by a FunctionCallingCallback, that suffers from this issue.
3. In DiffEqCallbacks, the default initilization function was adjusted to fit the new interface (with the additional `affect!` argument), for FunctionCallingCallbacks, PresetTimeCallbacks and PeriodicCallbacks see [this PR](https://github.com/SciML/DiffEqCallbacks.jl/pull/146). 
4. In SciMLBase, the functions `INITIALIZE_DEFAULT` and `FINALIZE_DEFAULT` where adjusted for the new signature including the `affect!`, see [this PR](https://github.com/SciML/SciMLBase.jl/pull/426).

Not really happy with the fact that these changes affect so many libraries, on the other hand I am with @frankschae at the point that this might be a clean way to handle this problem.

Tests for SciMLSensitivity running locally if all PRs are checked in.

Best regards!

ThummeTo:
All 4 PRs are ready to review from my side.

ChrisRackauckas:
> In DiffEqBase, the function signatures for the functions initialize and finalize on callbacks changed from (cb, u, t, integrator) to (cb, affect!, u, t, integrator) (theses functions are called in initialize! and finalize!), see https://github.com/SciML/DiffEqBase.jl/pull/899.

The function is already passed the callback, so `c.affect!` is in the first argument. How does this make a difference? It's a breaking change that's rather difficult to do so it should only be done if necessary, and it doesn't look necessary as the thing that's passed is already passed.

ThummeTo:
> > In DiffEqBase, the function signatures for the functions initialize and finalize on callbacks changed from (cb, u, t, integrator) to (cb, affect!, u, t, integrator) (theses functions are called in initialize! and finalize!), see [SciML/DiffEqBase.jl#899](https://github.com/SciML/DiffEqBase.jl/pull/899).
> 
> The function is already passed the callback, so `c.affect!` is in the first argument. How does this make a difference? It's a breaking change that's rather difficult to do so it should only be done if necessary, and it doesn't look necessary as the thing that's passed is already passed.

It this function is called on a "regular" Callback (e.g. during forward solving an ODE) you are right. But if sensitivities are computed, the field `c.affect!` becomes `c.affect.affect!` (because `c.affect` is now of type `TrackedAffect` which holds its own field for the original `affect!` function).

ChrisRackauckas:
Can't you just forward `c.affect` calls to c.affect.affect! ?

ThummeTo:
Like overwriting `setproperty!`,  `getproperty`?

ChrisRackauckas:
Yes, that would be safer. Because the other issue with this form is that if people update and still use `c.affect!` in `initialize!` and you passed in a TrackedAffect for the `affect!` part, then the tracked part will be missed and you'll get incorrect gradients. So it would rely heavily on all code updating in a specific way that's silent about correctness until looking at the gradients, which isn't a great property to have.

ThummeTo:
Ok, I see. This was basically my first suggestion (see discussion above) :-)

So the good thing is, the 4 PRs reduce to just one (in DiffEqCallbacks) and this is not even a breaking change but just a fix.

So I would overwrite `hasproperty`, `getproperty` and `setproperty!` for the TrackedAffect in DiffEqCallbacks so that it behaves like any other affect-type like e.g. FunctionCallingAffect? Basically it feels like TrackedAffect "inherits" (like in object orientated inheritance) the properties from the "parent" affect and feels just like the parent struct (with some tracking extras) from a user perspective?

I can go for it if you give me the OK ;-)

ThummeTo:
[This PR](https://github.com/SciML/SciMLSensitivity.jl/pull/810) would be the other approach that overwrites property getter/setters to mimic existance of fields from the "parent" struct inside of `TrackedAffect`. This would only be a minor change/soft update.

PR #808, opened 2023-03-22
Use SparseDiffTools v2 for steadystateadjoint

vpuri3:
part of: https://github.com/SciML/SciMLOperators.jl/issues/142


vpuri3:
TODO- update OrdinaryDiffEq to use sparsedifftools v2

src/steadystate_adjoint.jl
```diff
@@ -98,8 +100,7 @@ end
     end
 
     if !needs_jac
-        # TODO: FixedVecJacOperator should respect the `autojacvec` of the algorithm
-        operator = FixedVecJacOperator(f, y, p, Val(DiffEqBase.isinplace(sol.prob)))
+        operator = VecJac(f, y, p; autodiff = get_autodiff_from_vjp(vjp))
```
avik-pal, line 103: This is not really equivalent. IIRC `VecJac` recomputes the pullback everytime a call to `mul!` is made. In this case, we have a fixed input, only the seeding changes so we compute the pullback once and just reevaluate it multiple times.

vpuri3:
We should add a test case to ensure that `Zygote` is able to propagate gradients past `solve(::LinearProblem)`

vpuri3:
with this branch, your example is erroring on some `try/catch` block which is incompatible with zygote.
```julia
ERROR: LoadError: Compiling Tuple{LinearSolve.var"##solve#32", Base.Pairs{Symbol, Union
{}, Tuple{}, NamedTuple{(), Tuple{}}}, typeof(solve), LinearSolve.LinearCache{Base.Resh
apedArray{Float32, 2, SubArray{Float32, 1, Vector{Float32}, Tuple{UnitRange{Int64}}, tr
ue}, Tuple{}}, SubArray{Float32, 1, Vector{Float32}, Tuple{UnitRange{Int64}}, true}, Ve
ctor{Float32}, SciMLBase.NullParameters, KrylovJL{typeof(Krylov.gmres!), Int64, Tuple{}
, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}, Krylov.GmresSolver{Fl
oat32, Float32, Vector{Float32}}, SciMLOperators.IdentityOperator, SciMLOperators.Ident
ityOperator, Float32, true, LinearSolve.OperatorCondition.IllConditioned}, KrylovJL{typ
eof(Krylov.gmres!), Int64, Tuple{}, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(),
 Tuple{}}}}}: try/catch is not supported.                                              
Refer to the Zygote documentation for fixes.                                           
https://fluxml.ai/Zygote.jl/latest/limitations                                         ```

avik-pal:
Yes, ik I gave that as a testcase that doesn't work. The linear problem doesn't have tests because the adjoints aren't implemented.

Also, taking a look at how linearsolve works, it might a bit refactoring before we can include adjoints since it dispatches on solve instead of __solve

ChrisRackauckas:
We can make a higher level in SciMLBase and dispatch it on __solve.

avik-pal:
We can get this PR merged, right? The linear solve issue is entirely tangential and needs to be handled downstream first.

ChrisRackauckas:
someone needs to resolve the merge.

src/steadystate_adjoint.jl
avik-pal: Yes, see https://github.com/JuliaDiff/SparseDiffTools.jl/pull/245

ChrisRackauckas:
No Enzyme or ReverseDiff though?

PR #816, opened 2023-04-13
Support Enzyme v0.11

ChrisRackauckas:


src/concrete_solve.jl
```diff
@@ -8,11 +8,13 @@ function inplace_vjp(prob, u0, p, verbose)
...
     ez = try
         f = unwrapped_f(prob.f)
-        Enzyme.autodiff(Enzyme.Duplicated(du, du),
-                        copy(u0), copy(p), prob.tspan[1]) do out, u, _p, t
+
+        function adfunc(out, u, _p, t)
             f(out, u, _p, t)
             nothing
         end
+        Enzyme.autodiff(Enzyme.Reverse, adfunc, Enzyme.Duplicated(du, du),
```
wsmoses, line 16: @ChrisRackauckas you don't need to use adfunc here, you can just use `f` as the function to differentiate.

ChrisRackauckas: Explicitly forcing the return to be `nothing` won't help performance?

wsmoses: If you choose to remove the closure, you would likely want to add an extra const to specify the activity of the return.

Though I suppose forcing it to be nothing could improve performance regardles.

PR #827, opened 2023-05-23
Fix J and pJ construction for forward mode

avik-pal:
I can't seem to figure out a MWE for the problem. But a longer form trying to use the updated HNN in DiffEqFlux https://github.com/SciML/DiffEqFlux.jl/blob/a1b5846bc48f1f50b745dc7c2e5a5a5298abb96b/src/hnn.jl. The Jacobian and Parameter Jacobians were not being constructed despite both ForwardDiff and FiniteDiff needing them (FiniteDiff case was fine but this should fix the forwardDiff case)

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -101,7 +101,7 @@ function adjointdiffcache(g::G, sensealg, discrete, sol, dgdu::DG1, dgdp::DG2, f
         algevar_idxs = 1:0
     end
 
-    if !needs_jac && !(issemiexplicitdae || !isautojacvec)
+    if !needs_jac && !issemiexplicitdae && !(autojacvec isa Bool)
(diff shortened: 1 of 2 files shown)


src/adjoint_common.jl
```diff
@@ -101,7 +101,7 @@ function adjointdiffcache(g::G, sensealg, discrete, sol, dgdu::DG1, dgdp::DG2, f
         algevar_idxs = 1:0
     end
 
-    if !needs_jac && !(issemiexplicitdae || !isautojacvec)
+    if !needs_jac && !issemiexplicitdae && !(autojacvec isa Bool)
```
ChrisRackauckas, line 104: I think this needs to check if AutoFiniteDiff and AutoForwardDiff now?

PR #829, opened 2023-05-27
Test SparseDiffTools v2 with new OrdinaryDiffEq and StochasticDiffEq

ChrisRackauckas:


vpuri3:
@ChrisRackauckas i don;t think the `cudnn` precompile failure be related to this?

ChrisRackauckas:
It is not.

PR #830, opened 2023-05-30
Iterating over TrackedArray changes structure

avik-pal:
For details see https://discourse.julialang.org/t/fixed-frequency-control-in-diffeq/99446

### DIFF
--- a/src/callback_tracking.jl
+++ b/src/callback_tracking.jl
@@ -414,7 +414,7 @@ function setup_w_wp(cb::VectorContinuousCallback,
 end
 
 function get_FakeIntegrator(autojacvec::ReverseDiffVJP, u, p, t, tprev)
-    FakeIntegrator([x for x in u], [x for x in p], t, tprev)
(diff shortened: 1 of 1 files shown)


avik-pal:
This change breaks callbacks which modify the parameters. https://github.com/SciML/SciMLSensitivity.jl/blob/834a4c70c45c6525d66d6fa7ee7c29837435e0f9/test/callbacks/discrete_callbacks.jl#L203

Not really sure how to handle these cases.

avik-pal:
One solution might be reworking how CA.jl handles ReverseDiff https://github.com/jonniedie/ComponentArrays.jl/blob/master/ext/ComponentArraysReverseDiffExt.jl which honestly is not the correct way of handling tracked arrays. Instead, something like https://github.com/LuxDL/Lux.jl/blob/main/ext/LuxComponentArraysTrackerExt.jl should be done.

ChrisRackauckas:
I agree it probably needs to be handled in CA.jl and generate TrackedArrays

PR #847, opened 2023-07-13
Test Enzyme 0.11.0

ChrisRackauckas:


ChrisRackauckas:
@wsmoses it looks like we get 0 gradient issues with the latest Enzyme.

wsmoses:
Got a test/reproducer?

Also are you using the 0.11.5 that was released a few hours ago, or a different one?

Project.toml
```diff
@@ -54,7 +54,7 @@ DiffEqNoiseProcess = "4.1.4, 5.0"
 DiffRules = "1"
 Distributions = "0.24, 0.25"
 EllipsisNotation = "1"
-Enzyme = "0.11.2"
+Enzyme = "=0.11.0"
```
ChrisRackauckas, line 57:
```suggestion
Enzyme = "0.11.5"
```

ChrisRackauckas:
It was 0.11.4. Was that known to have some issues? https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/5535634554/jobs/10102198548?pr=1984#step:6:628

Pretty much every test failed with it. Running 0.11.5 now.

wsmoses:
No correctness ones that I'm aware of, so having a test would be useful for sure.

wsmoses:
I have confirmed at least one of your issues is bug in the KernelAbstractions extension package (see https://github.com/JuliaGPU/KernelAbstractions.jl/pull/412), hence any Enzyme version bumps will not fix the issue, since the issue is in another package.

In any case it requires the fix to KA to land and be bumped, at least for your test to pass.

ChrisRackauckas:
It looks like that was it. I still bumped the patch to keep life simple. Let's see what people throw back at me now 😅 . I think we're getting really close though, the errors are getting more obscure, really nice work @wsmoses!

PR #853, opened 2023-07-21
Fix cache type for tracked VectorContinuousCallback

frankschae:
Should fix https://github.com/SciML/SciMLSensitivity.jl/issues/852


ChrisRackauckas:
Can you add a test?

PR #856, opened 2023-07-31
Test master and see if docs build

ChrisRackauckas:


frankschae:
SDE Control docs failed because of
https://github.com/domluna/JuliaFormatter.jl/issues/745

PR #861, opened 2023-08-08
new gauss adjoint implemented

acoh64:
I created the new Gauss adjoint method in `src/gauss_adjoint.jl,` with a test in `test/gauss_adjoint.jl`. I do not think it is optimal yet but let me know how I can improve it!


ChrisRackauckas:
There's a big file `adjoint.jl` in the test cases that loops through all combinations of adjoints, can you add it to there as well?

src/gauss_adjoint.jl
```diff
@@ -0,0 +1,431 @@
...
+        end
+        tmp = back(λ)
+        out[:] .= vec(tmp[1])
+    elseif sensealg.autojacvec isa EnzymeVJP
+        tmp3, tmp4 = paramjac_config
+        tmp4 .= λ
+        out .= 0
+        Enzyme.autodiff(Enzyme.Reverse, pf, Enzyme.Duplicated(tmp3, tmp4),
+                        y, Enzyme.Duplicated(p, out), t)
+    end
```
ChrisRackauckas, line 289: we should probably `else error "not supported"`

ChrisRackauckas:
Rebase

src/gauss_adjoint.jl
acoh64: added

src/sensitivity_algorithms.jl
```diff
@@ -483,6 +483,105 @@ function setvjp(sensealg::QuadratureAdjoint{CS, AD, FDT, Nothing}, vjp) where {C
...
+
+Kim, S., Ji, W., Deng, S., Ma, Y., & Rackauckas, C. (2021). Stiff neural ordinary
+differential equations. Chaos: An Interdisciplinary Journal of Nonlinear Science, 31(9), 093122.
+"""
+struct GaussAdjoint{CS, AD, FDT, VJP} <:
+       AbstractAdjointSensitivityAlgorithm{CS, AD, FDT}
+    autojacvec::VJP
+    abstol::Float64
+    reltol::Float64
+end
```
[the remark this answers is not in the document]
acoh64: removed these tolerances

acoh64:
> 

I added tests here, currently GaussAdjoint passes for all tests except DAEs

ChrisRackauckas:
You could mark those as broken for now. DAEs and checkpointing are the two major missing features I see.

acoh64:
Just marked DAE tests as broken.

ChrisRackauckas:
Rebase

test/gauss_adjoint.jl
frankschae, line 1: I guess we can remove this since all tests are in tests/adjoint.jl?

test/adjoint.jl
```diff
@@ -761,6 +830,13 @@ _, res = adjoint_sensitivities(sol_mm, alg, t = ts, dgdu_discrete = dg,
 reference_sol = ForwardDiff.gradient(p -> G(p, prob_mm, ts, sum), vec(p))
 @test res'≈reference_sol rtol=1e-11
 
+# TEST FAILS HERE. WORKS FOR LOWER TOLERANCE
+_, res_gauss = adjoint_sensitivities(sol_mm, alg, t = ts, dgdu_discrete = dg,
+    abstol = 1e-14, reltol = 1e-14,
+    sensealg = GaussAdjoint())
+@test_broken res_gauss≈res rtol=1e-11
+@test res_gauss≈res rtol=1e-3
```
frankschae, line 838: do you understand already why it doesn't work here? Saving/callback issues?

src/sensitivity_algorithms.jl
```diff
@@ -483,6 +483,99 @@ function setvjp(sensealg::QuadratureAdjoint{CS, AD, FDT, Nothing}, vjp) where {C
...
+        This is the default and recommended for most users.
+      + `false`: the Jacobian is constructed via FiniteDiff.jl
+      + `true`: the Jacobian is constructed via ForwardDiff.jl
+      + `TrackerVJP`: Uses Tracker.jl for the vjp.
+      + `ZygoteVJP`: Uses Zygote.jl for the vjp.
+      + `EnzymeVJP`: Uses Enzyme.jl for the vjp.
+      + `ReverseDiffVJP(compile=false)`: Uses ReverseDiff.jl for the vjp. `compile`
+        is a boolean for whether to precompile the tape, which should only be done
+        if there are no branches (`if` or `while` statements) in the `f` function.
+  - `abstol`: absolute tolerance for the quadrature calculation
```
frankschae, line 534: these two lines can also be removed from the description

test/gauss_adjoint.jl
acoh64: Good point, I just removed this.

test/adjoint.jl
acoh64: I think it is a problem with the callback. I have noticed that it is not actually being called at every step of the integration. I am not entirely sure what is happening but I am looking into it now

acoh64: This was a separate issue with the IntegratingCallback, causing it to store double the amount of memory necessary. I have fixed this. I am still working on this issue

test/adjoint.jl
```diff
@@ -939,10 +1042,19 @@ _, res = adjoint_sensitivities(sol_singular_mm, alg, t = ts,
...
+    InterpolatingAdjoint(),
+    BacksolveAdjoint(),
+    GaussAdjoint(),
+]
     _, res = adjoint_sensitivities(sol_singular_mm, alg, t = ts,
         dgdu_discrete = dg_singular, abstol = 1e-14,
         reltol = 1e-14, sensealg = salg,
         maxiters = Int(1e6))
-    @test res'≈reference_sol rtol=1e-7
+    if salg == GaussAdjoint()
```
YingboMa, line 1055:
```suggestion
    if salg isa GaussAdjoint
```

test/adjoint.jl
```diff
@@ -939,10 +1042,19 @@ _, res = adjoint_sensitivities(sol_singular_mm, alg, t = ts,
...
+        @test_broken res'≈reference_sol rtol=1e-7
```
YingboMa, line 1056: Do we know why it's broken?

acoh64: I am not sure yet, trying to figure it out now

acoh64: Fixed the issue, GaussAdjoint passes all tests now

acoh64: To fix it, I had to add the reverse loss callback part in. I am not entirely sure what this part does. Do you know?

acoh64: Because maybe there is a smarter, more efficient way to deal with this in the IntegratingCallback

YingboMa: Yes, that's the first term in https://github.com/SciML/DiffEqDevMaterials/blob/master/dae_adjoint/output/main.pdf (32)

acoh64: The code that computes this doesn't use $\lambda$ at all though

acoh64: I think I see what is happening now, and I don't think there is a more efficient way to compute this since we do not need $\lambda$ for the calculation

src/sensitivity_algorithms.jl
acoh64: fixed

src/gauss_adjoint.jl
```diff
@@ -0,0 +1,431 @@
...
+        dgdu_continuous, dgdp_continuous, g, Val(true);
+        callback)
+    adj_sol = solve(adj_prob, alg; abstol = abstol, reltol = reltol,
+        save_everystep = true, save_start = true, kwargs...)
+    return adj_sol[end], compute_dGdp(integrand_values)'
+end
+
+function compute_dGdp(integrand::IntegrandValues)
+    temp = zeros(length(integrand.integrand), length(integrand.integrand[1]))
+    for i in 1:length(integrand.integrand)
```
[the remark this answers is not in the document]
acoh64: just fixed this and optimized this function (less memory)

test/adjoint.jl
```diff
@@ -245,7 +263,14 @@ _, easy_res11 = adjoint_sensitivities(soloop_nodense, Tsit5(), t = t, dgdu_discr
...
 #                                     ) isa Tuple
-
+_, easy_res12 = adjoint_sensitivities(soloop, Tsit5(), t = t, dgdu_discrete = dg,
+    abstol = 1e-14,
+    reltol = 1e-14,
+    sensealg = GaussAdjoint())
+_, easy_res122 = adjoint_sensitivities(soloop, Tsit5(), t = t, dgdu_discrete = dg,
+    abstol = 1e-14,
+    reltol = 1e-14,
+    sensealg = GaussAdjoint(autojacvec = ReverseDiffVJP(true)))
```
ChrisRackauckas, line 273: this is checking the oop

ChrisRackauckas (approved):

ChrisRackauckas:
I think this looks good to go. Issues to open after merging:

1. This is missing tests for callback support. It may just work, may need a tweak, but its untested as @frankschae points out so that would be a good follow up.
2. Performance optimizations. I don't think we've seen a flamegraph yet, and I'm sure we're missing some optimizations like https://github.com/SciML/DiffEqCallbacks.jl/issues/152 which are causing some unnecessary allocations.
3. Compatibility with Functors and non-vector parameters.

PR #862, opened 2023-08-09
Make has_original_jac check original_jac

oxinabox:
It seems to me that this makes more sense as a definition. And that doing it this way, makes my code work.
(though I haven't checked correctness)

Whereas without this, the branch is hit that tries to compute the original jacobian via finite differencing (with `autodiff=false, autojacvec=false, autojacmat=false`)

### DIFF
--- a/test/forward.jl
+++ b/test/forward.jl
@@ -232,7 +232,7 @@ p = [1.5, 1.0, 3.0, 1.0]
 jac_call_count = 0
 function jac_with_count(J, u, p, t)
     global jac_call_count
-    jac_call_count+=1
+    jac_call_count += 1
     (x, y, a, b, c) = (u[1], u[2], p[1], p[2], p[3])
     J[1, 1] = a + y * b * -1
     J[2, 1] = t * y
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Can you add a test for this?

test/forward.jl
```diff
@@ -227,3 +227,24 @@ p = [1.5, 1.0, 3.0, 1.0]
...
+    J[2, 2] = t * c * -1 + t * x
+end
+
+f = ODEFunction(fb, jac = jac_with_count, paramjac = paramjac)
+p = [1.5, 1.0, 3.0]
+absolutely_no_ad_sensealg = ForwardSensitivity(autodiff=false, autojacvec=false, autojacmat=false)
+prob = ODEForwardSensitivityProblem(f, [1.0; 1.0], (0.0, 10.0), p, absolutely_no_ad_sensealg)
+@test SciMLSensitivity.has_original_jac(prob.f)
+@assert jac_call_count == 0
+solve(prob, Tsit5(), abstol = 1e-14, reltol = 1e-14)
```
oxinabox, line 249: This actually errors, without this change.
Which I think might be an unrelated bug. Not sure

ChrisRackauckas (approved):

oxinabox:
tests failing seem 
1. unrelated
2. the same ones as in https://github.com/SciML/SciMLSensitivity.jl/pull/857

oxinabox:
bump

test/forward.jl
```diff
@@ -242,9 +242,15 @@ end
...
-absolutely_no_ad_sensealg = ForwardSensitivity(autodiff=false, autojacvec=false, autojacmat=false)
-prob = ODEForwardSensitivityProblem(f, [1.0; 1.0], (0.0, 10.0), p, absolutely_no_ad_sensealg)
+absolutely_no_ad_sensealg = ForwardSensitivity(autodiff = false,
+    autojacvec = false,
+    autojacmat = false)
+prob = ODEForwardSensitivityProblem(f,
+    [1.0; 1.0],
+    (0.0, 10.0),
+    p,
+    absolutely_no_ad_sensealg)
```
ChrisRackauckas, line 252: Shit I didn't check this. This looks to be the wrong format.

oxinabox: I just used julia formatter with the SciMLStyle

PR #864, opened 2023-08-12
Fix continuous adjoints for ComponentArrays

glatteis:
Attempts to fix https://github.com/SciML/SciMLSensitivity.jl/issues/854

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -915,10 +915,10 @@ function _jacNoise!(λ, y, p, t, S::TS, isnoise::ZygoteVJP, dgrad, dλ,
             if dgrad !== nothing
                 tmp2 = last.(out)
                 if !(eltype(tmp2) isa Nothing)
-                    !isempty(dgrad) && (dgrad .= stack(collect.(tmp2)))
+                    !isempty(dgrad) && (dgrad .= reduce(hcat, tmp2))
                 end
             end
-            dλ !== nothing && (dλ .= stack(collect.(first.(out))))
+            dλ !== nothing && (dλ .= reduce(hcat, first.(out)))
             dy !== nothing && (dy .= _dy)
         end
     else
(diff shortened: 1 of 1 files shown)


glatteis:
the `collect` broadcast does seem to work but it's scalar indexing when running it on a gpu, so I have to find another way to do this

glatteis:
Seems like `Diagonal` does not work with CUDA

src/derivative_wrappers.jl
```diff
@@ -911,14 +911,16 @@ function _jacNoise!(λ, y, p, t, S::TS, isnoise::ZygoteVJP, dgrad, dλ,
                     f(u, p, t)
                 end
             end
-            out = [back(x) for x in eachcol(Diagonal(λ))]
+            len = length(λ)
+            diag = Diagonal(λ) + 0 * reshape(repeat(y, len), len, len) # workaround: create dense diagonal matrix without scalar indexing
```
ChrisRackauckas, line 915: what is this doing? It's just a conversion, so just use `adapt`?

glatteis: It's a bad hack, `adapt` seems better but I realised this entire venture is not so meaningful because there is an assumption I missed (see the newest reply in https://github.com/SciML/SciMLSensitivity.jl/issues/854) :)

glatteis: Oh, the reason is that I can't `adapt` to something that I don't know, I think? Because CUDA is not a dependency. Maybe can I adapt to `typeof(` some existing value, but I couldn't get it to work this morning

ChrisRackauckas: Use `adapt(SciMLBase.parameterless_type(λ), Diagonal(λ))`

ChrisRackauckas:
@avik-pal do you know what this is for?

glatteis:
> @avik-pal do you know what this is for?

idea was to fix https://github.com/SciML/SciMLSensitivity.jl/issues/854, but that's more complicated

PR #877, opened 2023-08-21
Better Steady State Adjoint

avik-pal:
- [x] External Heuristics Control
  - [x] NoHeuristic --> Follow exactly what user wants
  - [x] SmallJACHeuristic --> If Jacobian is small then construct it ignoring user requirements
- [x] concrete_jac -- Force Jacobian Computation
- [x] Assume Uniform Square Block Diagonal (is there a mathematical name for these matrices where all the blocks are of uniform size?) -- Most Machine Learning Applications. But we can't default to it since the user might use `BatchNorm` which messes up the sparsity pattern assumption.
  - [x] If concrete jac -- use sparsity pattern to collapse Jacobian computation of an entire batch to a single batch (should scale really well)
  - [x] Use a specialized GMRES solver for this part.
- [x] Supports arbitrary parameter structures


avik-pal:
Needs https://github.com/SciML/LinearSolve.jl/pull/366 and a tag of SparseDiffTools.jl

ChrisRackauckas:
was this supposed to pass tests?

avik-pal:
> was this supposed to pass tests?

Yup :sweat:. If it doesn't pass now I will have to debug furthur

ChrisRackauckas:
It looks like this causes a regression in the vjps? See the Core 3 test failures.

avik-pal:
Closing this but not deleting the branch.

PR #881, opened 2023-08-24
GaussAdjoint now uses in-place integrand function

acoh64:
I tested this and it does reduce allocations

### DIFF
--- a/src/gauss_adjoint.jl
+++ b/src/gauss_adjoint.jl
@@ -152,7 +152,7 @@ end
     if RetCB
         return ODEProblem(odefun, z0, tspan, p), cb, rcb
     else
-        return ODEProblem(odefun, z0, tspan, p, callback = cb)
+        return ODEProblem(odefun, z0, tspan, p, callback = cb), cb, rcb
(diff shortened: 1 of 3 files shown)


src/gauss_adjoint.jl
```diff
@@ -326,7 +326,7 @@ function _adjoint_sensitivities(sol, sensealg::GaussAdjoint, alg; t = nothing,
     kwargs...)
     integrand = GaussIntegrand(sol, sensealg, dgdp_continuous)
     integrand_values = IntegrandValues(Vector{Float64})
-    cb = IntegratingCallback((u, t, integrator) -> integrand(t, u)[:], integrand_values)
+    cb = IntegratingCallback((out, u, t, integrator) -> integrand(out, t, u)[:], integrand_values, similar(sol.prob.p))
```
ChrisRackauckas, line 329:
```suggestion
    cb = IntegratingCallback((out, u, t, integrator) -> vec(integrand(out, t, u)), integrand_values, similar(sol.prob.p))
```

PR #882, opened 2023-08-25
mark a bugfix release

oxinabox:
I would like a release tagged here.
This has the bugfix #862  but doesn't have the new GaussAdjoint feature.
For some reason newer versions break for me, and I do need to debug it,
but right now I just want to get this release done and come back to that other stuff later.
So a bugfix release would be ideal.

I thought I could just use Manifests to pin my stuff to this, but it makes downstream testing hard.
So better if I can have a release and just pin it with a `~` selector in my project

**Don't Merge this PR to master**
merge it *somewhere else* then tag a release off of it.
(I can't create a branch on this repo)


ChrisRackauckas:
There have already been two releases since this PR?

PR #884, opened 2023-08-26
added checkpointing to gauss adjoint

acoh64:
See: https://github.com/SciML/SciMLSensitivity.jl/issues/869. I think this needs some more testing first. It only seems to be working at lower tolerances (1e-4), so I need to figure out why


ChrisRackauckas:
https://github.com/SciML/SciMLSensitivity.jl/actions/runs/5982716510/job/16232100431?pr=884#step:6:872

ChrisRackauckas:
@avik-pal would you know where that is from?

avik-pal:
Not really. Core2 tests are passing in other CI https://github.com/SciML/SciMLSensitivity.jl/actions/runs/6001815834/job/16276828862?pr=885. Also I can't find any `.t` in code, and the stacktrace seems incomplete?

ChrisRackauckas:
>  It only seems to be working at lower tolerances (1e-4), so I need to figure out why

Do the other ones do this too? We only have low tolerance tests.

src/sensitivity_interface.jl
acoh64, line 1: I think the issue is here. I removed checkpoints from kwargs because of an error that `:checkpoints` is an unrecognized kwarg for `solve`. However, I just realized that I don't think the correct checkpoints are being used now since `checkpoints` is not being passed to anything

acoh64:
The other ones checkpoint tests (for interpolating adjoint) work at a tolerance of 1e-9.

acoh64:
I fixed errors in the adjoint state that came from not passing the proper tolerances into the checkpointing solves. However, the gradient calculation still only works at tolerances of 1e-4

ChrisRackauckas:
Is the solve dense and not using `saveat`?

acoh64:
The checkpoint solves are dense but the adjoint solve is not

acoh64:
Which solve were you referring to? When I use the dense forward solve with checkpointing, then GaussAdjoint works to 1e-9 tolerance, but this feels like cheating.

The issue is in the forward solves, since the adjoint solution matches perfectly and the callback times line up

acoh64:
@ChrisRackauckas I figured out the problem: the integrand function is made ahead of time with the solution to the nondense solve. Therefore, it is using the parameter jacobians from the nondense solve in the integral calculation, causing it to be less accurate. I think the best solution is to make a callback which takes as an argument the integrator and a `sol`, however, I don't think this can be done with the current callback interface. Is there a way around this?

ChrisRackauckas:
Add the GaussIntegrand to the ODEGaussAdjointSensitivityFunction and make it a mutable struct so you can mutate `integrand.sol` when a new `cpsol` is taken.

acoh64:
Thanks, checkpointing now works for GaussAdjoint!

test/sde_scalar_stratonovich.jl
```diff
@@ -82,7 +85,21 @@ p2 = [1.01, 0.87]
...
+    
+    res_sde_u02, res_sde_p2 = adjoint_sensitivities(sol, EulerHeun(), t = Array(t),
+        dgdu_discrete = dg!,
+        abstol = 1e-14,
+        reltol = 1e-14,
+        adaptive = false,
+        sensealg = GaussAdjoint(autojacvec = false))
+
+    @testbroken isapprox(res_sde_u0, res_sde_u02, rtol = 1e-4)
+    @testbroken isapprox(res_sde_p, res_sde_p2, rtol = 1e-4)
```
ChrisRackauckas, line 97:
```suggestion
    @test_broken isapprox(res_sde_u0, res_sde_u02, rtol = 1e-4)
    @test_broken isapprox(res_sde_p, res_sde_p2, rtol = 1e-4)
```

ChrisRackauckas:
Looks like there's still a few test failures?

acoh64:
Checkpointing tests pass but I am still working on the SDE tests

frankschae (commented):
[@glatteis](https://github.com/glatteis) @acoh64's `GaussAdjoint` will probably also solve the efficiency problem you spotted. I think we can add in the callback
```julia
dy2, back = Zygote.pullback(y, p) do u, p
    g(u, p, t) .* dW 
end
out2 = back(λ)
```

test/adjoint.jl
```diff
@@ -2,6 +2,86 @@ using SciMLSensitivity, OrdinaryDiffEq, RecursiveArrayTools, DiffEqBase,
     ForwardDiff, Calculus, QuadGK, LinearAlgebra, Zygote
 using Test
 
+
+
+
+using Lux, Optimization, Plots, Random
```
ChrisRackauckas, line 8: This is not in the right place?

test/adjoint.jl
```diff
@@ -2,6 +2,86 @@ using SciMLSensitivity, OrdinaryDiffEq, RecursiveArrayTools, DiffEqBase,
...
+end
+
+adtype = Optimization.AutoZygote()
+optf = Optimization.OptimizationFunction((x, p) -> loss(x), adtype)
+optprob = Optimization.OptimizationProblem(optf, p_model)
+
+tmp1 = Zygote.gradient(loss,ComponentArray(p_model))
+tmp2 = Zygote.gradient(loss,p_model)
+
+res0 = Optimization.solve(optprob, PolyOpt(), maxiters = 100)
```
ChrisRackauckas, line 60: Test just the adjoint

acoh64:
@ChrisRackauckas I removed the SDE stuff so this should be ready to go with checkpointing and nonvector parameter implementations

ChrisRackauckas:
It looks like there's two test failures still seen, one in the docs and one in core 3.

PR #885, opened 2023-08-27
Fix SteadyStateAdjoint Breakage

avik-pal:
Not sure how the DEQ downstream tests did not catch this. Needs https://github.com/JuliaDiff/SparseDiffTools.jl/pull/256 as well.

I have added a direct test here as well to ensure this doesn't happen again.


avik-pal:
@ChrisRackauckas test fails seem same as master. Is this good to go?

PR #890, opened 2023-09-01
Delegate observed for ForwardSensitivityFunction

### LINKED ISSUE #888: sol[var] etc observations do not work on ODEForwardSensitivityProblem
`SciMLBase.getobserved(prob)` where `prob` was constructed via `ODEForwardSensitivityProblem`
always returns `DEFAULT_OBSERVED` -- which errors if anyone tries to read anything from it.
Even if the original problem did have observations.

My hack around this is:
```julia
function Base.getproperty(f::ODEForwardSensitivityFunction, name::Symbol)
    name == :observed && return f.f.observed
    return getfield(f, name)
end
Base.hasproperty(f::ODEForwardSensitivityFunction, name::Symbol) = (name == :observed) || hasfield(f, name)
SciMLBase.has_observed(f::ODEForwardSensitivityFunction) = SciMLBase.has_observed(f.f)
```

but possibly we should instead be passing it onwards when we make the `ODEProblem` that `ODEForwardSensitivityProblem`,
rather than relying on `getproperty` overloads.

I *think* this is correct as the equations and variables in the expanded system of equations are all placed *after* the original ones. Right?

(This is a special case of #886, but i wanted to open a seperate issue because the discussion might be nontrivial)

oxinabox:
closes #888 
idk if this is the best way but it works.

Not to sure how to test it better than this as IDK how to create a simple ODEProblem that has an observed.

### DIFF
--- a/src/forward_sensitivity.jl
+++ b/src/forward_sensitivity.jl
@@ -86,6 +86,13 @@ function ODEForwardSensitivityFunction(f, analytic, tgrad, original_jac, jac, ja
     return sensefun
 end
 
+function Base.getproperty(f::ODEForwardSensitivityFunction, name::Symbol)
+    name == :observed && return f.f.observed
+    return getfield(f, name)
+end
+Base.propertynames(f::ODEForwardSensitivityFunction) = (:observed, fieldnames(typeof(f)))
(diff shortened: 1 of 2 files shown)


oxinabox:
Bump

ChrisRackauckas:
It's blocked by the broken tests, but @frankschae is looking into that.

Issue #854, opened 2023-07-28
Scalar indexing in pullback for diagonal noise


glatteis:
Hi,
this line is scalar indexing in a pullback:
https://github.com/SciML/SciMLSensitivity.jl/blob/1997fb1a2c288f3da37f61c7b0894eb4e42c5cd6/src/derivative_wrappers.jl#L899

glatteis:
It seems like I can't run my model on the GPU at all :(

MWE:
```
using Lux, Zygote, DifferentialEquations, ComponentArrays, Random, CUDA, SciMLSensitivity

rng = Xoshiro()

drift_net = Dense(2 => 2)
diffusion_net = Dense(2 => 2)

ps_drift_, st_drift = Lux.setup(rng, drift_net)

ps_diffusion_, st_diffusion = Lux.setup(rng, diffusion_net)

ps_ = ComponentArray((ps_drift=ps_drift_,ps_diffusion=ps_diffusion_)) |> Lux.gpu

function drift(u, ps, t)
    drift_net(u, ps.ps_drift, st_drift)[1]
end
function diffusion(u, ps, t)
    diffusion_net(u, ps.ps_diffusion, st_diffusion)[1]
end

u0 = [1f0, 1f0] |> Lux.gpu

tspan = (0f0, 1f0)
datasize = 10

solver = EulerHeun()
sensealg = InterpolatingAdjoint(autojacvec=ZygoteVJP())

function loss(ps)
    problem = SDEProblem(drift, diffusion, u0, tspan, ps)
    solution = solve(problem, solver; sensealg=sensealg, saveat=collect(range(tspan[1], tspan[end], datasize)), dt=(tspan[end] / datasize))
    return sum(vec(solution |> Lux.gpu))
end

println(loss(ps_))
println(Zygote.gradient(ps -> loss(ps), ps_))
```

On CPU:
```
┌ Info: The GPU function is being called but the GPU is not accessible.
│ Defaulting back to the CPU. (No action is required if you want
└ to run on the CPU).
-0.445899
((ps_drift = (weight = Float32[11.908233 11.131403; 1.8911543 1.8107269], bias = Float32[10.83848; 1.7313616;;]), ps_diffusion = (weight = Float32[-22.582191 -22.426481; -0.6261419 -1.0105969], bias = Float32[-21.103664; -0.61384314;;])),)
```

On GPU:
```
41.25828
ERROR: LoadError: Scalar indexing is disallowed.
[...]

PR #916, opened 2023-10-17
Add disclaimer that optimize-than-discretize adjoints for SDE run in O(n^2)

glatteis:
...until issue #854 is solved. Otherwise, people would maybe waste time trying to use this method even though its implementation is currently not efficient.


ChrisRackauckas:
@frankschae were you going to look into it?

frankschae:
Yes, I could do that. We should maybe discuss once, before I do it, what other changes might be useful for StochasticDiffEq (e.g. diagonal noise as vector -> diagonal matrix? or https://github.com/SciML/StochasticDiffEq.jl/issues/472?).

PR #921, opened 2023-10-24
Remove EnzymeVJP from defaults list when alg forwarddiffs

ChrisRackauckas:
Works around https://github.com/SciML/SciMLSensitivity.jl/issues/920


wsmoses:
Per https://github.com/SciML/SciMLSensitivity.jl/issues/920#issuecomment-1778630875 I do not think this is an Enzyme bug

wsmoses (changes requested):
Triggering issue does not appear to be an Enzyme bug, but caller error (see: https://github.com/SciML/SciMLSensitivity.jl/issues/920#issuecomment-1778630875)

PR #961, opened 2023-12-27
Make ReverseDiffAdjoint return ODE solutions

ChrisRackauckas:
Fixes https://github.com/SciML/SciMLSensitivity.jl/issues/894

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,7 @@ jobs:
           - SDE2
           - SDE3
         version:
-          - '1'
+          - '1.9'
     steps:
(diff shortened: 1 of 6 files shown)


avik-pal:
Oh this is really nice. No longer will weird dispatches be needed https://github.com/LuxDL/Lux.jl/pull/431

ChrisRackauckas:
Yeah, this is a consequence of RAT v3 making AbstractVectorOfArray no longer an AbstractArray. ReverseDiff doesn't require the auto-conversions anymore (which is a weird thing that it did in the first place, but 🤷 at least it's not an issue anymore).

PR #963, opened 2023-12-28
Improve Robustness to RATv3

ChrisRackauckas:
Since AbstractVectorOfArray should be handled properly given the deprecation warning


src/concrete_solve.jl
```diff
@@ -437,8 +438,8 @@ function DiffEqBase._concrete_solve_adjoint(prob::Union{SciMLBase.AbstractODEPro
             else
                 !Base.isconcretetype(eltype(Δ)) &&
                     (Δ[i] isa NoTangent || eltype(Δ) <: NoTangent) && return
-                if Δ isa AbstractArray{<:AbstractArray} || Δ isa DESolution
-                    x = Δ[i]
+                if Δ isa AbstractArray{<:AbstractArray} || Δ isa AbstractVectorOfArray
+                    x = Δ isa AbstractVectorOfArray ? Δ.u[i] : Δ[i]
                     if _save_idxs isa Number
                         _out[_save_idxs] = @view(x[_save_idxs])
```
ChrisRackauckas, line 444:
```suggestion
                        _out[_save_idxs] = x[_save_idxs]
```

src/concrete_solve.jl
```diff
@@ -411,16 +411,17 @@ function DiffEqBase._concrete_solve_adjoint(prob::Union{SciMLBase.AbstractODEPro
...
                 eltype(Δ) <: NoTangent && return
-                if Δ isa AbstractArray{<:AbstractArray} && length(Δ) == 1 && i == 1
+                if (Δ isa AbstractArray{<:AbstractArray} || Δ isa AbstractVectorOfArray) && length(Δ) == 1 && i == 1
                     # user did sol[end] on only_end
+                    x = Δ isa AbstractVectorOfArray ? Δ.u[1] : Δ[1]
                     if _save_idxs isa Number
-                        x = vec(Δ[1])
-                        _out[_save_idxs] .= adapt(outtype, @view(x[_save_idxs]))
+                        vx = vec(x)
+                        _out[_save_idxs] .= adapt(outtype, @view(vx[_save_idxs]))
```
ChrisRackauckas, line 419:
```suggestion
                        _out[_save_idxs] .= vx[_save_idxs]
```

AayushSabharwal:
So the issue with this is the following (so far as I can determine):
In [this line of Zygote](https://github.com/FluxML/Zygote.jl/blob/223f85620749bb96e68e57b13f1161034c5b528c/src/lib/broadcast.jl#L218), it maps over a `VectorOfArray` `y∂b`. The code expects that iterating over it will yield items of type `eltype(y∂b)`, but VoA doesn't work like that and yields the subarrays instead, which breaks the code

AayushSabharwal:
The "fix" I have is also not great. There's the underlying issue that sometimes Zygote turns `VectorOfArray`s into `Ref`s containing a `NamedTuple`. There's a hack around this in RAT's ZygoteExt, but it crops up here too

ChrisRackauckas:
Oh that means that the constructor of a VectorOfArray is missing a chain rule.

AayushSabharwal:
Could it also be that the `Array(::VectorOfArray)` functions are missing an adjoint? Adding one for that fixed it for me

ChrisRackauckas:
Ahh that could do it.

AayushSabharwal:
I think we also need a method for this:
```
ERROR: MethodError: no method matching (::ChainRulesCore.ProjectTo{VectorOfArray, NamedTuple{(), Tuple{}}})(::Vector{Float64})
```

AayushSabharwal:
I don't fully understand what `Zygote.unbroadcast` is supposed to do. I hacked together something that works specifically for `VectorOfArray`, but it doesn't work for `ODESolution`

PR #966, opened 2023-12-28
test v1.9 with RAT v3

ChrisRackauckas:


ChrisRackauckas:
Everything passes with RATv2, so that's the culprit 😅

PR #971, opened 2024-01-10
Try fixing the 2nd callback test set

ChrisRackauckas:


ChrisRackauckas:
Looks like that was it, thanks.

ChrisRackauckas:
Note @acoh64 that we should probably do this with GaussAdjoint too since that is checkpointing.

acoh64:
It looks like the change was in the one line of `callback_tracking.jl`. I can add a check for `GaussAdjoint` too. Were there any changes to how the checkpointing is done?

ChrisRackauckas:
Nope, no changes to checkpointing.

PR #980, opened 2024-01-18
ci: update TagBot workflow

thazhemadam:
* Update TagBot workflow to expect the `lookback` input for workflow dispatches.
* Explicitly specify permissions required for the TagBot workflow.
* Also, cleanup and remove extraneous `.vscode`.


.github/workflows/TagBot.yml
```diff
@@ -4,6 +4,22 @@ on:
...
+  contents: write
+  deployments: read
+  issues: read
+  discussions: read
+  packages: read
+  pages: read
+  pull-requests: read
+  repository-projects: read
+  security-events: read
+  statuses: read
```
ChrisRackauckas, line 22: Why does this one need to be different from the other repos?

thazhemadam: This isn't different, it's just explicitly mentioned while the TagBot workflows in other repos are likely not up-to-date.
This is the recommended list of permissions, as per [TagBot's README.md](https://github.com/thazhemadam/TagBot/blob/16a1824512433d6aa510b6f43a9f00877510c6b9/README.md#setup).

PR #982, opened 2024-01-21
CI Trigger

ArnoStrouwen:
This hangs indefinitely for me locally on windows.
Let us see if CI has the same issue.

```jl
julia> sensealg = SciMLSensitivity.setvjp(sensealg,
           SciMLSensitivity.inplace_vjp(fwd_sol.prob, fwd_sol.prob.u0, fwd_sol.prob.p, true))
┌ Warning: Using fallback BLAS replacements, performance may be degraded
└ @ Enzyme.Compiler C:\Users\arno\.julia\packages\GPUCompiler\U36Ed\src\utils.jl:59
warning: didn't implement memmove, using memcpy as fallback which can result in errors
┌ Warning: Using fallback BLAS replacements, performance may be degraded
└ @ Enzyme.Compiler C:\Users\arno\.julia\packages\GPUCompiler\U36Ed\src\utils.jl:59
warning: didn't implement memmove, using memcpy as fallback which can result in errors
```


ArnoStrouwen:
@ChrisRackauckas is this one of the known issues already, or a new one?
https://github.com/SciML/SciMLSensitivity.jl/actions/runs/7598726366/job/20694999313?pr=982#step:6:631

ChrisRackauckas:
The warning isn't unexpected, but I am surprised it just stalled the tests for 6 hours?

ArnoStrouwen:
Yes, it stalls indefinitely on that line, but only on Windows.

PR #986, opened 2024-01-22
wrap safetestsets into testset per GROUP

ArnoStrouwen:
I think this extra wrapping in testsets is required to make sure all the safetestsets get run per group, instead of stopping at the first failing safetestset.

Specifically, steady state failing caused the next two not to be ran:
```jl
@time @safetestset "Steady State Adjoint" include("steady_state.jl")
@time @safetestset "Concrete Solve Derivatives of Second Order ODEs" include("second_order_odes.jl")
@time @safetestset "Parameter Compatibility Errors" include("parameter_compatibility_errors.jl")
```
While in this PR they should run.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -9,115 +9,141 @@ function activate_gpu_env()
     Pkg.instantiate()
 end
 
-@time begin
+@time @testset "SciMLSensitivity" begin
     if GROUP == "All" || GROUP == "Core1" || GROUP == "Downstream"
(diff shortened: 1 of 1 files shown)


ArnoStrouwen:
It now runs all the tests correctly:
https://github.com/SciML/SciMLSensitivity.jl/actions/runs/7611215862/job/20726219373#step:6:969

ArnoStrouwen:
This is ready for review.

PR #1008, opened 2024-02-08
Fix tolerance definitions in adjoint_oop

ChrisRackauckas:


test/adjoint_oop.jl
```diff
@@ -183,8 +183,7 @@ f_dp = ForwardDiff.gradient(G_p, p)
 
 du0, dp = Zygote.gradient((u0, p) -> sum(concrete_solve(prob, Tsit5(), u0, p,
         abstol = 1e-10, reltol = 1e-10, saveat = tsteps,
```
ChrisRackauckas, line 185:
```suggestion
        abstol = 1e-14, reltol = 1e-14, saveat = tsteps,
```

PR #1010, opened 2024-02-08
Use the SciMLStructures Tunable interface

ChrisRackauckas:
This uses the SciMLStructures Tunable interface https://github.com/SciML/SciMLStructures.jl in order to allow more generalized definitions of `p`.

- [ ] Ensure Lux.jl is well supported (componentarrays extension in SciMLStructures
- [ ] Add a test for a custom SciMLStructure

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -6,14 +6,14 @@
 const have_not_warned_vjp = Ref(true)
 const STACKTRACE_WITH_VJPWARN = Ref(false)
 
-function inplace_vjp(prob, u0, p, verbose)
+function inplace_vjp(prob, u0, p, verbose, repack)
     du = zero(u0)
 
     ez = try
         f = unwrapped_f(prob.f)
 
         function adfunc(out, u, _p, t)
-            f(out, u, _p, t)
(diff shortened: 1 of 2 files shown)


ArnoStrouwen:
You need to actually pass `repack`?
https://github.com/SciML/SciMLSensitivity.jl/blob/dcf961ad5739be4167877a25bc6ec6fb36198cb7/src/concrete_solve.jl

ChrisRackauckas:
Yeah this isn't quite done. We should do a branch of, if it's a mutable structure then we just use `replace!(::AbstractPortion, p::T1, new_values)`, and only `repack` if it's immutable.

ArnoStrouwen:
Something isn't completely clicking with me, how this change can be some simple.
Where is the logic living that when in the adjoint equations `df/dp` and `dg/dp`  occurs, we should only take the tunable parameters?

ChrisRackauckas:
I should've marked WIP. Plan to finish it during today.

xtalax:
Can we expect this to be released by Monday?

ChrisRackauckas:
yes

PR #1020, opened 2024-02-24
reapply formatter

ArnoStrouwen:
This is a large one, let CI run to completion in case I overlooked the formatter eating a line.


ArnoStrouwen:
@ChrisRackauckas merge this in if tests pass. First try failed because of failure to download Julia.

PR #1037, opened 2024-03-28
Fixes needed for Manifold Projection Semi-Explicit DAEs

avik-pal:
I am using this to accumulate all the fixes needed for the project, so let's hold off on merging.

- Fixes SteadyState Adjoint for OOP jacobian


ChrisRackauckas:
I assume everything just uses latest these days?

PR #1038, opened 2024-03-29
fix sign mistake in direct sensitivity tutorial

ArnoStrouwen:
@SebastianM-C noticed a sign difference between direct sensitivity method, and forwarddiff on:
https://docs.sciml.ai/SciMLSensitivity/stable/tutorials/direct_sensitivity/.
I think a minus sign when applying the chain rule was missed.


SebastianM-C:
It might be good to have doctests when asserting something in the text, so that these kind of things fail the docs build, but that's a more general thing.

PR #1060, opened 2024-05-29
Fix GaussAdjoint with callbacks

ChrisRackauckas:


Vaibhavdixit02:
Ah, this was what https://github.com/SciML/SciMLSensitivity.jl/pull/1034 was trying to address lol

ChrisRackauckas:
That would've been good to know 😅

src/concrete_solve.jl
```diff
@@ -245,7 +245,8 @@ function DiffEqBase._concrete_solve_adjoint(
         has_cb = false
     end
     default_sensealg = automatic_sensealg_choice(prob, u0, p, verbose)
-    if has_cb && default_sensealg isa AbstractAdjointSensitivityAlgorithm
+    if has_cb && default_sensealg isa AbstractAdjointSensitivityAlgorithm &&
+       !isinplace(sol.prob)
```
ChrisRackauckas, line 249:
```suggestion
    if has_cb && default_sensealg isa AbstractAdjointSensitivityAlgorithm
```

ChrisRackauckas:
Bump?

frankschae:
hmm ... it's probably close but it currently fails for the callback where the correction should be 0:

```julia
g(sol) = sum(sol)
function dg!(out, u, p, t, i)
    (out .= 1)
end
@testset "callbacks with no effect" begin
    condition(u, t, integrator) = t == 5
    affect!(integrator) = integrator.u[1] += 0.0
    cb = DiscreteCallback(condition, affect!, save_positions = (false, false))
    tstops = [5.0]
    test_discrete_callback(cb, tstops, g, dg!)
end
```

I think it's the line:

```julia
if sensealg isa GaussAdjoint
            @assert integrator.f.f isa ODEGaussAdjointSensitivityFunction
            @show integrator.f.f.integrating_cb.affect!.integrand_values.integrand dgrad
            integrator.f.f.integrating_cb.affect!.integrand_values.integrand .= dgrad
```
in callback_tracking.jl, whch for the example above prints:

```julia
integrator.f.f.integrating_cb.affect!.integrand_values.integrand = [4.318834073956617, -23.50595577533703, 3.507883353046969, -26.70725798446173]
dgrad = [-0.0, -0.0, -0.0, -0.0]
```

ChrisRackauckas:
Is that with the VJPs as Enzyme or ReverseDiff? IIUC it always defaults to ReverseDiff right now?

ChrisRackauckas:
The incorrect values I think stem from missing `make_zero!`s, but it's currently waiting on an Enzyme tag from @wsmoses before https://github.com/SciML/SciMLSensitivity.jl/pull/1067 finishes and then this can get retested.

ChrisRackauckas:
@jClugstor did your PR look at GaussAdjoint?

jClugstor:
No, but there is a '@test_broken' for 'GaussAdjoint' with a callback in that PR, so if callbacks get fixed that will need to change.

src/gauss_adjoint.jl
```diff
@@ -577,7 +580,8 @@ function _adjoint_sensitivities(sol, sensealg::GaussAdjoint, alg; t = nothing,
 
     if sol.prob isa ODEProblem
         adj_prob, cb2, rcb = ODEAdjointProblem(
-            sol, sensealg, alg, integrand, t, dgdu_discrete,
+            sol, sensealg, alg, integrand, integrating_cb,
```
ChrisRackauckas, line 583:
```suggestion
            sol, sensealg, alg, integrand, cb,
```

PR #1061, opened 2024-05-31
Vitepress based documentation

lazarusA:
let's see.


avik-pal:
We need a proper support for MultiDocumenter before SciML can adopt vitepress

avik-pal:
Also buildkite deployment detection fails for reasons unknown, that's why we had added passing in the decision variable to Vitepress. See how Lux does it https://github.com/LuxDL/Lux.jl/blob/60c595ec9490dbdcce70c21ec4d577ef1d4f9165/docs/make.jl#L87-L88

lazarusA:
> We need a proper support for MultiDocumenter before SciML can adopt vitepress

Definitely, (or similar?). I just wanna to see if something could be broken. This is a huge use case. I will take a look at the deploy issue again. Thanks.

docs/make.jl
```diff
@@ -9,14 +10,24 @@ using Plots
...
     modules = [SciMLSensitivity],
     clean = true, doctest = false, linkcheck = true,
-    warnonly = [:missing_docs],
-    format = Documenter.HTML(assets = ["assets/favicon.ico"],
-        canonical = "https://docs.sciml.ai/SciMLSensitivity/stable/"),
+    warnonly = true,
+    format = DocumenterVitepress.MarkdownVitepress(
+            repo="https://github.com/SciML/SciMLSensitivity.jl",
+            devbranch="master", devurl="dev",
+            deploy_url="https://docs.sciml.ai/SciMLSensitivity/", deploy_decision
```
avik-pal, line 25:
```suggestion
    format = DocumenterVitepress.MarkdownVitepress(;
            repo="https://github.com/SciML/SciMLSensitivity.jl",
            devbranch="master", devurl="dev",
            deploy_url="https://docs.sciml.ai/SciMLSensitivity/", deploy_decision
```

ChrisRackauckas:
Yes multidocumenter is essential. Notsure how this connects to it.

avik-pal:
It still isn't detecting the key. This is probably because buildkite disables forks from accessing secrets?

lazarusA:
apparently it builds without any issues. I will take this as a good sign 😄.  The multidocumenter thing is not just a super navigator tab for all repos?

ChrisRackauckas:
Yeah it demonstrates it's viable. It just needs multidocumenter and key things. @thazhemadam can do the key things, but we should wait until there's a plan with multidocumenter @pfitzner

ChrisRackauckas:
Closing as it is stale, but we are talking with @asinghvi17 about doing something to SciML hopefully down the line. I want to get SciML to vitepress but we need to think more about the whole multidocumenter setup and less about the single repos.

pfitzseb:
MultiDocumenter 0.8 should somewhat support vitepress generated docs, at least visually. You will get two top-bars and potential keybinding conflicts though.

PR #1067, opened 2024-06-09
Safety make_zero! on repeated Enzyme calls with caches

ChrisRackauckas:
This should ensure that the caches are always zero'd for the derivatives.

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -707,7 +707,8 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::EnzymeVJP, dgrad, 
     #if dy !== nothing
     #      tmp3 = dy
     #else
-    Enzyme.make_zero!(tmp3)
+    tmp3 .= 0
+    #Enzyme.make_zero!(tmp3)
     #end
 
     vec(tmp4) .= vec(λ)
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -715,16 +715,11 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::EnzymeVJP, dgrad, 
 
     isautojacvec = get_jacvec(sensealg)
 
-    if Core.Compiler.isconstType(_tmp6)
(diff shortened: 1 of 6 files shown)


src/derivative_wrappers.jl
```diff
@@ -692,6 +692,8 @@ function _vecjacobian!(dλ, y, λ, p, t, S::TS, isautojacvec::EnzymeVJP, dgrad,
 
     isautojacvec = get_jacvec(sensealg)
 
+    Enzyme.make_zero!(_tmp6)
```
wsmoses, line 695: above where you have the tmp2/3 .= 0, you may be able to replace that with make_zero! as well

src/gauss_adjoint.jl
```diff
@@ -476,6 +476,7 @@ function vec_pjac!(out, λ, y, t, S::GaussIntegrand)
         tmp3, tmp4, tmp6 = paramjac_config
         tmp4 .= λ
         out .= 0
+        Enzyme.make_zero!(tmp6)
```
wsmoses, line 479: also maybe out here/below

ChrisRackauckas:
It looks like it fails on a trivial closure?

https://github.com/SciML/SciMLSensitivity.jl/actions/runs/9501768875/job/26188120021?pr=1067#step:6:688

ChrisRackauckas:
For some reason this function fails `Enzyme.Compiler.guaranteed_const_nongen(typeof(_pf), nothing)` false and then in the next iterate it errors. It seems like `make_zero!` is not robust to active values in closures?

src/derivative_wrappers.jl
ChrisRackauckas: I had to revert those due to https://github.com/EnzymeAD/Enzyme.jl/issues/1611

Project.toml
```diff
@@ -61,7 +61,7 @@ DiffEqNoiseProcess = "5.19"
 Distributed = "1"
 Distributions = "0.25"
 EllipsisNotation = "1"
-Enzyme = "0.12"
+Enzyme = "0.12.21"
```
ChrisRackauckas, line 64:
```suggestion
Enzyme = "0.12.22"
```

PR #1080, opened 2024-07-13
docs: Make example code consistent with text

moble:
The text in the section ["Example: Continuous Adjoints on an Energy Functional"](https://docs.sciml.ai/SciMLSensitivity/stable/tutorials/adjoint_continuous_functional/#Example:-Continuous-Adjoints-on-an-Energy-Functional) does not agree with the code.  Specifically, it's describing the *energy functional* — in the title and the text — which I think we would all agree should be the sum of the squares.  In fact, the text actually gives that formula as displayed latex.  But the code consistently uses the square of the sum.  This PR corrects the code.

## Checklist


ChrisRackauckas (approved):
Thanks for diving into this.

PR #1085, opened 2024-07-26
Update Forward Mode to use SciMLStructures

DhairyaLGandhi:
## Checklist

## Additional context

Updates the forward mode dispatches to also use the SciMLStructures interface. As with elsewhere, only the tunable portion is respected while constructing the backpass. I have replaced the check for the compatibilty for `p` with that for SciMLStructures, but wanted feedback on whether this is correct. The previous check was trying to `p <: AbstractArray` so I think we should be covered if we get a SciMLStructure and canonically get the tunables from it.

Needs tests


DhairyaLGandhi:
Needs to be tested with https://github.com/SciML/ModelingToolkit.jl/pull/2929

PR #1087, opened 2024-07-30
Remove duplicate callbacks in ForwardDiff and ReverseDiff adjoints

jClugstor:
## Checklist

## Additional context
Fixes https://github.com/SciML/SciMLSensitivity.jl/issues/1081

In some of the Adjoint methods, if a problem that was constructed with a callback is passed to it, the callback was being called twice. This because the `kwargs` argument already contains the callback that was passed to the problem constructor. So when the problem is solved, it uses the callback in the problem itself, and the callback passed to the `solve`, effectively doing the same callback operation twice. 

This PR just makes it so that only the callback in the `kwargs` is used.

### DIFF
--- a/test/prob_kwargs.jl
+++ b/test/prob_kwargs.jl
@@ -35,23 +35,29 @@ a = ones(3)
 odef(du, u, p, t) = du .= u .* p
 prob = ODEProblem(odef, [2.0], (0.0, 1.0), [3.0])
 
-function f1(u0p)
-    condition(u, t, integrator) = t == 0.5
-    affect!(integrator) = integrator.p[1] += 0.1
-    cb = DiscreteCallback(condition, affect!)
-    prob = ODEProblem{true}(odef, u0p[1:1], (0.0, 1.0), u0p[2:2]; callback = cb)
-    sum(solve(prob, Tsit5(), tstops = [0.5], sensealg = ForwardDiffSensitivity()))
-end
+let callback_count1 = 0, callback_count2 = 0
+    function f1(u0p, adjoint_type)
+        condition(u, t, integrator) = t == 0.5
+        affect!(integrator) = callback_count1 += 1
+        cb = DiscreteCallback(condition, affect!)
+        prob = ODEProblem{true}(odef, u0p[1:1], (0.0, 1.0), u0p[2:2]; callback = cb)
+        sum(solve(prob, Tsit5(), tstops = [0.5], sensealg = adjoint_type))
+    end
 
-function f2(u0p)
-    condition(u, t, integrator) = t == 0.5
-    affect!(integrator) = integrator.p[1] += 0.1
-    cb = DiscreteCallback(condition, affect!)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Needs tests, make the problem from the issue into a test case.

test/prob_kwargs.jl
```diff
@@ -29,3 +29,28 @@ a = ones(3)
 @test Zygote.gradient(f, a)[1][1] ≈ Zygote.gradient(f2, a)[1][1]
 @test Zygote.gradient(f, a)[1][2] == Zygote.gradient(f2, a)[1][2] == 0
 @test Zygote.gradient(f, a)[1][3] == Zygote.gradient(f2, a)[1][3] == 0
+
+# callback in problem construction or in solve call should give same result
+odef(du, u, p, t) = du .= u .* p
```
ChrisRackauckas, line 34:
```suggestion
# callback in problem construction or in solve call should give same result
# https://github.com/SciML/SciMLSensitivity.jl/issues/1081
odef(du, u, p, t) = du .= u .* p
```

ChrisRackauckas:
Also handle ReverseDiffAdjoint() and TrackerAdjoint as mentioned in the issue.

jClugstor:
It errors out when using ReverseDiffAdjoint or TrackerAdjoint, since the callback tries to mutate the parameter array.

ChrisRackauckas:
Make some `@test_broken` and comment in the original issue. Maybe test another callback that doesn't require mutating?

jClugstor:
Turns out TrackerAdjoint had the same problem.
I made the tests check that the callbacks are called the same number of times, which is the root of the issue.

test/prob_kwargs.jl
```diff
@@ -29,3 +29,37 @@ a = ones(3)
...
+        condition(u, t, integrator) = t == 0.5
+        affect!(integrator) = callback_count2 += 1
+        cb = DiscreteCallback(condition, affect!)
+        prob = ODEProblem{true}(odef, u0p[1:1], (0.0, 1.0), u0p[2:2])
+        sum(solve(prob, Tsit5(), tstops = [0.5], callback = cb, sensealg = adjoint_type))
+    end
+
+    @testset "Callback duplication check" begin
+        for adjoint_type in [
+            ForwardDiffSensitivity(), ReverseDiffAdjoint(), TrackerAdjoint()]
```
ChrisRackauckas, line 57: and loop through some adjoint methods too, all sensealgs is best

ChrisRackauckas: Except ZygoteAdjoint

jClugstor: `ZygoteAdjoint()` and `GaussAdjoint()` error out, besides those I've included all the rest that support callbacks

ChrisRackauckas: For the ones that error, label them as @test_broken. 

Gauss adjoint with callbacks is currently on the critical Todo.

ChrisRackauckas (approved):

PR #1101, opened 2024-08-30
test: use TRBDF2 for the steady state adjoints

avik-pal:
See https://github.com/SciML/SteadyStateDiffEq.jl/pull/82


avik-pal:
@ChrisRackauckas seems like the issue is fixed

https://github.com/SciML/SciMLSensitivity.jl/actions/runs/10635160969/job/29484143086?pr=1101#step:6:945 is the only remaining problematic one

PR #1124, opened 2024-09-26
refactor: cleanup deps and replace SparseDiffTools with SciMLJacobianOperators

avik-pal:


avik-pal:
I think we should drop the NonlinearSolve dep in DiffEqCallbacks a weakdep as a breaking change

avik-pal:
Tests for steadystateadjoint pass locally. We need to wait for the following:

- [ ] SciMLJacobianOperators to be registered (EOD today)
- [ ] OptimizationBase still does support Enzyme 0.13

avik-pal:
We are getting there, we need

`add SteadyStateDiffEq#avik-pal-patch-1 ModelingToolkit#master OptimizationBase#main`

ChrisRackauckas:
Looks like shadowing is failing?

avik-pal:
yeah, typo in refactor. Should be fixed now. The docs are broken but Vaibhav has a open PR for that so I am not touching it atm.

ChrisRackauckas:
Shit didn't mean to click that. There was still an error in shadowing?

avik-pal:
Could that be from any of the packages being unlocked from upgrading? The only change in the shadowing was `@unpack` to `(; `. Quite a few of the packages while allowed to update weren't being tested in CI due to compat issues.

PR #1159, opened 2025-01-31
Test master

ChrisRackauckas:


ChrisRackauckas:
It looks like Flux got more implicit with its parameters, so implicit that we cannot even pass in a vector? So it looks like we need to drop Flux.train!.

ChrisRackauckas:
@avik-pal do you have a minibatching data loader you'd recommend? Just cutting out the Flux stuff.

avik-pal:
`DataLoader` from `MLUtils` https://lux.csail.mit.edu/stable/tutorials/intermediate/1_NeuralODE#Loading-MNIST

PR #1167, opened 2025-03-03
refactor: change `state_values(sol)` to `sol.u`

AayushSabharwal:


ChrisRackauckas:
What's the reason for not doing this one?

AayushSabharwal:
I thought it'd fix things, but broke them instead. I'll take another stab at it later, since we got SciMLBase CI working without this and it isn't immediately blowing things up.

PR #1168, opened 2025-03-06
Feat: Handle Adjoints through Initialization

DhairyaLGandhi:
## Checklist

## Additional context

MTK and SciML construct an initialization problem before starting the time stepping to ensure the starting values of the unknowns and parameters adhere to any constraints needed for the system. This PR adds handling for adjoint sensitivities of the NonlinearProblem, NonlinearSquaresProblem, SCCNonlinearProblem etc.

I am opening this to get some feedback regarding how we can accumulate gradients correctly. I have also included a test case for a DAE which I will update to use the values out of SciMLSensitivity.

Currently the gradients get calculated but don't get accumulated, we need to be able to update the gradients for the parameters. Since this is a manual dispatch, the usual graph building in AD is bypassed, and we need to handle this manually. Ideally, we should make it so the cfg itself includes the initialization so we would not have gotten incorrect gradients in the first place 😅  We are also forced to use a LinearProblem instead of  `\` because it cannot handle singular jacobians. 

cc @ChrisRackauckas


DhairyaLGandhi:
I wanted to ask whether it is preferred to retain `\` for the smaller problems, which would be slower with LinearSolve, and if so, we will need to deal with LAPACK errors.

ChrisRackauckas:
> I wanted to ask whether it is preferred to retain \ for the smaller problems, which would be slower with LinearSolve, and if so, we will need to deal with LAPACK errors.

LinearSolve.jl should be faster across the board? It depends a bit on the CPU architecture since it depends on whether it guesses the right LU correctly,

ChrisRackauckas:
Note that with the latest MTK update, there is now an `Initials` section where the initial `u0` live. That should fix a few things.

src/steadystate_adjoint.jl
```diff
@@ -103,15 +102,18 @@ end
     else
         if linsolve === nothing && isempty(sensealg.linsolve_kwargs)
             # For the default case use `\` to avoid any form of unnecessary cache allocation
```
ChrisRackauckas, line 104: Yeah I don't know about that comment. I think it's just old. (a) `\` always allocates because it uses `lu` instead of `lu!`, so it's re-allocating the while matrix which is larger than any LinearSolve allocation, and (b) we have since 2023 setup tests on StaticArrays, so the immutable path is non-allocating. I don't think (b) was true when this was written.

src/concrete_solve.jl
```diff
@@ -425,6 +425,21 @@ function DiffEqBase._concrete_solve_adjoint(
             save_end = true, kwargs_fwd...)
     end
 
+    # Get gradients for the initialization problem if it exists
+    igs = if _prob.f.initialization_data.initializeprob != nothing
+        iprob = _prob.f.initialization_data.initializeprob
+        ip = parameter_values(iprob)
+        itunables, irepack, ialiases = canonicalize(Tunable(), ip)
+        igs, = Zygote.gradient(ip) do ip
```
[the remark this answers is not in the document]
DhairyaLGandhi: Not yet. These gradients are currently against the parameters of the initialization problem, not the system exactly. And the mapping between the two is ill defined, so we cannot simply `accum`

I spoke with @AayushSabharwal about a way to map, it seems `initialization_data.intializeprobmap` might have some support to return the correctly shaped vector,  but there are cases where we cannot know the ordering of `dp` either.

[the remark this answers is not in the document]
DhairyaLGandhi: There's another subtlety. I am not sure we haven't missed some part of the cfg by manually handling accumulation of gradients. Or any transforms we might need to calculate gradients for. The regular AD graph building typically took care of these details for us, but in this case we would need to worry about incorrect gradients manually

src/steadystate_adjoint.jl
DhairyaLGandhi: So glad we can remove this branch altogether.

src/concrete_solve.jl
```diff
@@ -425,6 +425,21 @@ function DiffEqBase._concrete_solve_adjoint(
             save_end = true, kwargs_fwd...)
     end
 
+    # Get gradients for the initialization problem if it exists
+    igs = if _prob.f.initialization_data.initializeprob != nothing
```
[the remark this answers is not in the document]
DhairyaLGandhi: How can I indicate to `solve` to avoid running initialization?

src/concrete_solve.jl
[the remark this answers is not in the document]
ChrisRackauckas: Oh yes, you need to use the initializeprobmap https://github.com/SciML/SciMLBase.jl/blob/master/src/initialization.jl#L268 to map it back to the shape of the initial parameters. 

> but there are cases where we cannot know the ordering of dp either.

p and dp just need the same ordering, so initializeprobmap should do the trick.

> There's another subtlety. I am not sure we haven't missed some part of the cfg by manually handling accumulation of gradients. Or any transforms we might need to calculate gradients for. The regular AD graph building typically took care of these details for us, but in this case we would need to worry about incorrect gradients manually

This is the only change to `(u0,p)` before solving, so this would account for it, given `initializeprobmap` is just an index map so an identity function.

src/concrete_solve.jl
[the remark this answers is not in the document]
ChrisRackauckas: `initializealg = NoInit()`. Should probably just do `CheckInit()` for safety but either is fine.

DhairyaLGandhi:
Trying to use the initialization end to end caused gradients against parameters to get dropped. https://github.com/DhairyaLGandhi/SciMLBase.jl/tree/dg/nonlinear is a WIP branch which adds adjoints to the `getindex` calls, which does capture the expected gradients, but we still end up dropping gradients somewhere in the chain. I am looking into whether we are doing so in the custom adjoints, because that was an issue I had identified for the ODE case.

src/concrete_solve.jl
```diff
@@ -425,6 +426,25 @@ function DiffEqBase._concrete_solve_adjoint(
...
     end
 
+    # Get gradients for the initialization problem if it exists
+    igs = if _prob.f.initialization_data != nothing
+        Zygote.gradient(tunables) do tunables
+            new_prob = remake(_prob, p = repack(tunables))
+            new_u0, new_p, _ = SciMLBase.get_initial_values(new_prob, new_prob, new_prob.f, SciMLBase.OverrideInit(), Val(true);
+                                                            abstol = 1e-6,
+                                                            reltol = 1e-6,
+                                                            sensealg = SteadyStateAdjoint(autojacvec = ZygoteVJP()))
```
ChrisRackauckas, line 436: shouldn't default to ZygoteVJP. Should use the autojacvec of the ODE

DhairyaLGandhi:
I could use some understanding of how to handle initialization when MTK analytically solves the problem, and removes all the unknowns. In that case `u0`, `u` etc are empty, and we attempt to calculate the jacobian for it in `dgdgu_val` which seems weird, there are no unknowns to calculate the jacobian for. It would be equally incorrect to use the unknowns from the system itself. The best way to handle this might be to simply add a check for `isempty` and return. Is that reasonable?

DhairyaLGandhi:
Core5 fails because we try to call `canonicalize` on parameters constructed by Lux models (ie a NamedTuple). These are not SciMLStructures, what's the best course of action here @AayushSabharwal

AayushSabharwal:
Use a `ComponentArray`?

src/adjoint_common.jl
```diff
@@ -379,11 +391,11 @@ function get_paramjac_config(autojacvec::ReverseDiffVJP, p, f, y, _p, _t;
         if !isRODE
             __p = p isa SciMLBase.NullParameters ? _p :
                   SciMLStructures.replace(Tunable(), p, _p)
-            tape = ReverseDiff.GradientTape((y, __p, [_t])) do u, p, t
+            tape = ReverseDiff.GradientTape((y, _p, [_t])) do u, p, t
```
ChrisRackauckas, line 394: are you sure this is the right one?

src/concrete_solve.jl
```diff
@@ -412,16 +414,42 @@ function DiffEqBase._concrete_solve_adjoint(
...
     isq = sensealg isa QuadratureAdjoint
+
+    igs, new_u0, new_p = if _prob.f.initialization_data !== nothing
+        local new_u0
+        local new_p
+        iy, back = Zygote.pullback(tunables) do tunables
+            new_prob = remake(_prob, p = repack(tunables))
+            new_u0, new_p, _ = SciMLBase.get_initial_values(new_prob, new_prob, new_prob.f, SciMLBase.OverrideInit(), Val(true);
+                                                            abstol = 1e-6,
+                                                            reltol = 1e-6,
```
ChrisRackauckas, line 425: These don't make sense.

src/adjoint_common.jl
DhairyaLGandhi: The tunables are replaced into the original structure (`p`) from `_p` and we repack the `_p` tunables again into `p` again when passing into the `f`

src/concrete_solve.jl
DhairyaLGandhi: Addressed in https://github.com/SciML/SciMLSensitivity.jl/pull/1168/commits/984c2ce1ff36a29a75c4bae759eba3e1bc6e2a49

src/adjoint_common.jl
DhairyaLGandhi: The difference is that we were passing just `tunables` into `f`, now we are passing the SciMLStructure, which is what the `f` expects. cc @AayushSabharwal for whether this makes sense

ChrisRackauckas: >The difference is that we were passing just tunables into f

Makes sense.

src/concrete_solve.jl
```diff
@@ -412,16 +415,40 @@ function DiffEqBase._concrete_solve_adjoint(
...
+
+        igs, new_u0, new_p
+    else
+        nothing, u0, p
+    end
+    _prob = remake(_prob, u0 = new_u0, p = new_p)
+
     if sensealg isa BacksolveAdjoint
-        sol = solve(_prob, alg, args...; save_noise = true,
+        sol = solve(_prob, alg, args...; initializealg = SciMLBase.NoInit(), save_noise = true,
```
ChrisRackauckas, line 443: It should only noinit if the previous case was ran. Won't this right now break the brownbasic tests?

src/concrete_solve.jl
```diff
@@ -1688,26 +1719,56 @@ function DiffEqBase._concrete_solve_adjoint(
...
-            else
+            elseif Δ isa AbstractArray{<:AbstractArray} || Δ isa AbstractVectorOfArray || Δ isa AbstractArray
                 @. _out[_save_idxs] = Δ[_save_idxs]
+            elseif isnothing(_out)
+                _out
+            else
+                @. _out[_save_idxs] = Δ.u[_save_idxs]
             end
         end
         dp = adjoint_sensitivities(sol, alg; sensealg = sensealg, dgdu = df)
```
ChrisRackauckas, line 1748: When the new reverse ode is built it needs to drop the initial eqs but still keep the dae constraints. It can brownbasic?

src/concrete_solve.jl
```diff
@@ -412,16 +415,40 @@ function DiffEqBase._concrete_solve_adjoint(
         Base.diff_names(Base._nt_names(values(kwargs)),
         (:callback_adj, :callback))}(values(kwargs))
     isq = sensealg isa QuadratureAdjoint
+
+    igs, new_u0, new_p = if _prob.f.initialization_data !== nothing
```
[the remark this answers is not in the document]
DhairyaLGandhi: Got it. My understanding was that OverrideInit was what we strictly needed. We can check for BrownBasicInit/ defaults here.

src/concrete_solve.jl
DhairyaLGandhi: If there was no intialization data, it won't have ran the initialisation problem at all.

If I can genetically ignore handling `initializealg` and pass it directly to `get_initial_values`, that would be good.

src/concrete_solve.jl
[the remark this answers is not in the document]
DhairyaLGandhi: There doesn't seem to be a method which can take a `BrownFullBasicInit()`. I get a `MethodError`:

```julia
ERROR: MethodError: no method matching get_initial_values(::ODEProblem{…}, ::ODEProblem{…}, ::ODEFunction{…}, ::BrownFullBasicInit{…}, ::Val{…}; sensealg::SteadyStateAdjoint{…}, nlsolve_alg::Nothing)

Closest candidates are:
  get_initial_values(::Any, ::Any, ::Any, ::NoInit, ::Any; kwargs...)
   @ SciMLBase ~/Downloads/arpa/jsmo/t2/SciMLBase.jl/src/initialization.jl:282
  get_initial_values(::Any, ::Any, ::Any, ::SciMLBase.OverrideInit, ::Union{Val{true}, Val{false}}; nlsolve_alg, abstol, reltol, kwargs...)
   @ SciMLBase ~/Downloads/arpa/jsmo/t2/SciMLBase.jl/src/initialization.jl:224
  get_initial_values(::SciMLBase.AbstractDEProblem, ::SciMLBase.DEIntegrator, ::Any, ::CheckInit, ::Union{Val{true}, Val{false}}; abstol, kwargs...)
   @ SciMLBase ~/Downloads/arpa/jsmo/t2/SciMLBase.jl/src/initialization.jl:161
  ...

```

Only `CheckInit`, `NoInit`, and `OverrideInit` have dispatches.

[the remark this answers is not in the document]
DhairyaLGandhi: What would be the best course of action here? Seems like supporting BrownBasicInit is a dispatch that will automatically be utilised when it is moved into SciMLBase.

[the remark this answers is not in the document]
ChrisRackauckas: Yes I mean BrownBasicInit should not be taking this path. But that's a problem because then they will be disabled in the next stage below, and that needs to be accounted for. This dispatch is already built and setup for BrownBasicInit and there are tests on that.

src/concrete_solve.jl
ChrisRackauckas: No that is not correct. If there was no initialization data then it will use the built in initialization, defaulting to BrownBasicInit. It's impossible for a DAE solver to generally work without running initialization of some form, the MTK one is just a new specialized one but there has always been a numerical one in the solver. And if it hits that case, this code will now disable that. 

https://github.com/SciML/SciMLSensitivity.jl/blob/master/test/adjoint.jl#L952-L978 this code will hit that. I think it's not failing because it's not so pronounced here. You might want to change that test to https://github.com/SciML/SciMLSensitivity.jl/blob/master/test/adjoint.jl#L975C5-L975C69 `prob_singular_mm = ODEProblem(f, [1.0, 0.0, 1.0], (0.0, 100), p)` and it would pass before and fail now.

src/concrete_solve.jl
DhairyaLGandhi: Is there a way to drop the initial eqs after its solved? The assumption was since we run with `NoInit`, no initialization is run post the first call to `get_initial_values` and we accumulate those gradients independently of the adaptive solve.

ChrisRackauckas: But the reverse pass needs to run with some form of initialization or the starting algebraic conditions may not be satisfied. Don't run this one with NoInit(), that would be prone to hiding issue. For this one, at most CheckInit(), but I'm saying that BrownBasicInit() is likely the one justified here since the 0 initial condition is only true on the differential variables, while the algebraic variable initial conditions will be unknown, but the Newton solve will have zero derivative because all of the inputs are just Newton guesses, so BrownBasic will work out for the reverse. We should probably hardcode that since it's always the solution there.

src/concrete_solve.jl
DhairyaLGandhi: Both `BrownBasicInit` and `OrdinaryDiffEqCore.DefaultInit` require us to depend on a whole package for the default dispatch. Can it be exposed as a dispatch of `get_initial_values` instead?

src/concrete_solve.jl
DhairyaLGandhi: Ok, that will require us to add an OrdinaryDiffEqCore dep in this package. I will add that.

src/concrete_solve.jl
ChrisRackauckas: We can just take the dep. It's pretty much required anyways.

ChrisRackauckas:
```julia
using OrdinaryDiffEq

function rober(du, u, p, t)
    y₁, y₂, y₃ = u
    k₁, k₂, k₃ = p
    du[1] = -k₁ * y₁ + k₃ * y₂ * y₃
    du[2] = k₁ * y₁ - k₂ * y₂^2 - k₃ * y₂ * y₃
    du[3] = y₁ + y₂ + y₃ - 1
    nothing
end
function rober(u, p, t)
    y₁, y₂, y₃ = u
    k₁, k₂, k₃ = p
    return [-k₁ * y₁ + k₃ * y₂ * y₃,
        k₁ * y₁ - k₂ * y₂^2 - k₃ * y₂ * y₃,
        y₁ + y₂ + y₃ - 1]
end
M = [1.0 0 0
     0 1.0 0
     0 0 0]

f = ODEFunction{true}(rober, mass_matrix = M)
p = [0.04, 3e7, 1e4]

prob_singular_mm = ODEProblem(f, [1.0, 0.0, 1.0], (0.0, 100), p)
sol_singular_mm = solve(prob_singular_mm, FBDF(autodiff = false),
    reltol = 1e-12, abstol = 1e-12, initializealg = NoInit())
```

The solvers have gotten good enough that they end up fixing this in the first step 😅 but the initial condition is wrong. So a test to add here is:

```julia
sol_singular_mm = solve(prob_singular_mm, FBDF(autodiff = false),
    reltol = 1e-12, abstol = 1e-12, initializealg = BrownFullBasicInit())
@test all(u[1] + u[2] + u[3] ≈ 1 for u in sol_singular_mm.u)
```

This test `all(u[1] + u[2] + u[3] ≈ 1 for u in sol_singular_mm.u)` will fail without the fix, and harder DAEs will just fail to start solving.

ChrisRackauckas: Note that this is in the concrete solve part, so it would need to be tested by checking if the forward solution used in a loss function satisfies that property. The manual solve of that test should be okay since it won't go down this code path at all. So we might need to make another mass matrix DAE test with concrete solve if we don't have one.

ChrisRackauckas: So a way to do this same test is then in here https://github.com/SciML/SciMLSensitivity.jl/blob/master/test/mtk.jl#L37 have one where you set it to BrownFullBasicInit check whether `w ~ x + y + z + 2 * β` is satisfied at every step (with the issue potentially being at step 1)

DhairyaLGandhi: I printed the results with `BrownFullBasicInit`. We get  

```julia
dmtk, = Zygote.gradient(mtkparams) do p
    new_sol = solve(prob, Rosenbrock23(), p = p, initializealg = BrownFullBasicInit())
    Zygote.ChainRules.ChainRulesCore.ignore_derivatives() do
        @show extrema(new_sol[x + y + z + 2 * β - w])
        @test all(new_sol[x + y + z + 2 * β - w]  .≈ 0)
    end
    mean(abs.(new_sol[sys.x] .- gt))
end
```

```julia
extrema(new_sol[(x + y + z + 2β) - w]) = (-7.882583474838611e-15, 7.105427357601002e-15)
```

There are some errors with a `MethodError` for `initialize_dae!(::SDEIntegrator, ...)`

src/concrete_solve.jl
DhairyaLGandhi: Is the 0 derivative also applicable to parameters? Or only the unknowns?

src/concrete_solve.jl
DhairyaLGandhi: Seems like SDEs don't work with `BrownBasicInit`. I tried to get the "default `initializealg` for `solve`" from the `kwargs` and `prob.kwargs`, but those are both empty. And `prob.f.initialization_data` is still `nothing`.

src/concrete_solve.jl
ChrisRackauckas: Its applicable to all Newton guess values. There is no parameter init going on to reverse so it's only for algebraic conditions so it's only Newton guesses.

src/concrete_solve.jl
ChrisRackauckas: That will need to change the integrator to `Rodas5P`, Tsit5 will not be compatible with this form.

ChrisRackauckas: For SDEs, we will just need to make it compatible with BrownBasicInit.

DhairyaLGandhi: Ah, I was so confused why it worked out, I see the `InitialFailure` now.

ChrisRackauckas: Okay good. Yeah because MTK is too smart and makes lots of simple examples not DAEs 😅. But now you got the DAE, and if not running the built in init then you get the error I was expecting. The fix is that it needs to run brownbasic before solving for the same reason reverse needs to. Good we worked out a test for this

src/concrete_solve.jl
```diff
@@ -412,16 +415,61 @@ function DiffEqBase._concrete_solve_adjoint(
...
+    kwargs_init = kwargs_adj[Base.diff_names(Base._nt_names(kwargs_adj), (:initializealg,))]
+
+    if haskey(kwargs, :initializealg) || haskey(prob.kwargs, :initializealg)
+        initializealg = haskey(kwargs, :initializealg) ? kwargs[:initializealg] : prob.kwargs[:initializealg]
+    else
+        initializealg = DefaultInit()
+    end
+
+    default_inits = Union{OverrideInit, Nothing, DefaultInit}
+    igs, new_u0, new_p, new_initializealg = if (SciMLBase.has_initialization_data(_prob.f) && initializealg isa default_inits) || initializealg isa CheckInit
```
ChrisRackauckas, line 427:
```suggestion
    igs, new_u0, new_p, new_initializealg = if (SciMLBase.has_initialization_data(_prob.f) && initializealg isa default_inits)
```

I don't see why `CheckInit` would take this branch: if CheckInit then you effectively "do nothing" (after validating it's correct to do nothing)

DhairyaLGandhi: But if we must do something, then, does it run the equivalent of OverrideInit?

src/concrete_solve.jl
```diff
@@ -1688,26 +1740,56 @@ function DiffEqBase._concrete_solve_adjoint(
...
-            else
+            elseif Δ isa AbstractArray{<:AbstractArray} || Δ isa AbstractVectorOfArray || Δ isa AbstractArray
                 @. _out[_save_idxs] = Δ[_save_idxs]
+            elseif isnothing(_out)
+                _out
+            else
+                @. _out[_save_idxs] = Δ.u[_save_idxs]
             end
         end
         dp = adjoint_sensitivities(sol, alg; sensealg = sensealg, dgdu = df)
```
ChrisRackauckas, line 1769:
```suggestion
        dp = adjoint_sensitivities(sol, alg; sensealg = sensealg, dgdu = df, initializealg = BrownBasicInit())
```

test/mtk.jl
```diff
@@ -32,8 +33,73 @@ prob = ODEProblem(sys, u0, tspan, p, jac = true)
...
+sensealg = GaussAdjoint(; autojacvec = SciMLSensitivity.ZygoteVJP())
+dmtk_incorrectu0, = Zygote.gradient(mtkparams_incorrectu0) do p
+    new_sol = solve(prob_incorrectu0, Rodas5P(); p = p, initializealg = BrownFullBasicInit(), sensealg, abstol = 1e-6, reltol = 1e-3)
+    Zygote.ChainRules.ChainRulesCore.ignore_derivatives() do
+        @test new_sol.retcode == SciMLBase.ReturnCode.Success
+        @test all(isapprox.(new_sol[x + y + z + 2 * β - w], 0, atol = 1e-12))
+        @test all(isapprox.(new_sol[x^2 + y^2 - w2^2], 0, atol = 1e-5, rtol = 1e0))
+    end
+    mean(abs.(new_sol[sys.x] .- gt))
+end
```
ChrisRackauckas, line 85:
```suggestion
dmtk_overrideinit_incorrectu0, = Zygote.gradient(mtkparams_incorrectu0) do p
    new_sol = solve(prob_incorrectu0, Rodas5P(); p = p, initializealg = OverrideInit(), sensealg, abstol = 1e-6, reltol = 1e-3)
    Zygote.ChainRules.ChainRulesCore.ignore_derivatives() do
        @test new_sol.retcode == SciMLBase.ReturnCode.Success
        @test all(isapprox.(new_sol[x + y + z + 2 * β - w], 0, atol = 1e-12))
        @test all(isapprox.(new_sol[x^2 + y^2 - w2^2], 0, atol = 1e-5, rtol = 1e0))
    end
    mean(abs.(new_sol[sys.x] .- gt))
end

dmtk_incorrectu0, = Zygote.gradient(mtkparams_incorrectu0) do p
    new_sol = solve(prob_incorrectu0, Rodas5P(); p = p, initializealg = BrownFullBasicInit(), sensealg, abstol = 1e-6, reltol = 1e-3)
    Zygote.ChainRules.ChainRulesCore.ignore_derivatives() do
        @test new_sol.retcode == SciMLBase.ReturnCode.Success
        @test all(isapprox.(new_sol[x + y + z + 2 * β - w], 0, atol = 1e-12))
        @test all(isapprox.(new_sol[x^2 + y^2 - w2^2], 0, atol = 1e-5, rtol = 1e0))
    end
    mean(abs.(new_sol[sys.x] .- gt))
end
```

src/concrete_solve.jl
ChrisRackauckas: @AayushSabharwal there isn't a custom CheckInit path right?

ChrisRackauckas: https://github.com/SciML/SciMLBase.jl/blob/master/src/initialization.jl#L192 there is, but the derivative is still guaranteed to be zero.

test/mtk.jl
```diff
@@ -32,8 +33,73 @@ prob = ODEProblem(sys, u0, tspan, p, jac = true)
...
+
+u0_correct = [D(x) => 2.0,
+    x => 1.0,
+    y => 0.0,
```
ChrisRackauckas, line 89:
```suggestion
    y => t,
```

This should stress the BrownBasicInit part that I suggested above, as using the MTK init would then start with a vastly wrong value of `y`

test/mtk.jl
```diff
@@ -32,8 +33,73 @@ prob = ODEProblem(sys, u0, tspan, p, jac = true)
...
+    D(z) ~ x * y - β * z,
+    w ~ x + y + z + 2 * β,
+    0 ~ x^2 + y^2 - w2^2
+]
+
+@mtkbuild sys = ODESystem(eqs, t)
+
+u0_incorrect = [D(x) => 2.0,
+    x => 1.0,
+    y => 0.0,
```
ChrisRackauckas, line 58:
```suggestion
    y => t,
```

ChrisRackauckas:
I added the BrownBasic for the reverse and setup the tests so that there's a version that would capture that issue of the MTK init in the reverse pass. It almost certainly needs the fix to https://github.com/SciML/ModelingToolkit.jl/issues/3570 to pass though, so it will likely fail at first and @AayushSabharwal this is a reason to prioritize getting that one completed.

But when that is merged and this passes, then I think this is good to go.

test/desauty_dae_mwe.jl
```diff
@@ -0,0 +1,56 @@
...
+
+tunables, repack, aliases = SS.canonicalize(SS.Tunable(), parameter_values(iprob))
+
+linsolve = LinearSolve.DefaultLinearSolver(LinearSolve.DefaultAlgorithmChoice.QRFactorization)
+sensealg = SciMLSensitivity.SteadyStateAdjoint(autojacvec = SciMLSensitivity.ZygoteVJP())
+# sensealg = SciMLSensitivity.SteadyStateAdjoint(autojacvec = SciMLSensitivity.ZygoteVJP(), linsolve = linsolve)
+igs, = Zygote.gradient(tunables) do p
+    iprob2 = remake(iprob, p = repack(p))
+    sol = solve(iprob2,
+                sensealg = sensealg
```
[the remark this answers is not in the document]
DhairyaLGandhi: `iprob` is the initialization problem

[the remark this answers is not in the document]
ChrisRackauckas: oh I see.

ChrisRackauckas (approved):

src/concrete_solve.jl
```diff
@@ -412,16 +415,61 @@ function DiffEqBase._concrete_solve_adjoint(
...
+        initializeprob = prob.f.initialization_data.initializeprob
+        iu0 = state_values(initializeprob)
+        isAD = if iu0 === nothing
+            AutoForwardDiff
+        elseif has_autodiff(alg)
+            OrdinaryDiffEqCore.alg_autodiff(alg) isa AutoForwardDiff
+        else
+            true 
+        end
+        nlsolve_alg = default_nlsolve(nothing, Val(isinplace(_prob)), iu0, initializeprob, isAD)
```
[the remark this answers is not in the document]
ChrisRackauckas: They have at least been used downstream here for a long while. We should comment that for sure.

test/parameter_initialization.jl
```diff
@@ -0,0 +1,99 @@
...
+    x => 1.0,
+    y => 0.0,
+    z => 0.0]
+
+p = [σ => 28.0,
+    ρ => 10.0,
+    β => 8 / 3]
+
+tspan = (0.0, 100.0)
+prob = ODEProblem(sys, u0, tspan, p, jac = true)
```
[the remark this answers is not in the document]
ChrisRackauckas: Yeah there's an older version of the test that was overwritten, this one should be reverted.

ChrisRackauckas (approved):

test/mtk.jl
```diff
@@ -32,8 +34,98 @@ prob = ODEProblem(sys, u0, tspan, p, jac = true)
...
+          
+          (prob_correctu0, mtkparams_correctu0, NoInit()), 
+          (prob_correctu0, mtkparams_correctu0, nothing), 
+]
+
+grads = map(setups) do setup
+    prob, ps, init = setup
+    @show init
+    Zygote.gradient(ps) do p
+        new_sol = solve(prob, Rodas5P(); p = p, initializealg = init, sensealg, abstol = 1e-6, reltol = 1e-3)
```
ChrisRackauckas, line 118:
```suggestion
        if init === nothing
            new_sol = solve(prob, Rodas5P(); p = p, sensealg, abstol = 1e-6, reltol = 1e-3)
        else
            new_sol = solve(prob, Rodas5P(); p = p, initializealg = init, sensealg, abstol = 1e-6, reltol = 1e-3)
        end
```

DhairyaLGandhi:
```julia
julia> prob, ps, init = setups[3];

julia> init
OrdinaryDiffEqCore.DefaultInit()

julia> new_sol = solve(prob, Rodas5P(); p = p, sensealg, initializealg = init, abstol = 1e-6, reltol = 1e-3)
┌ Warning: Initialization system is overdetermined. 1 equations for 0 unknowns. Initialization will default to using least squares. `SCCNonlinearProblem` can only be used for initialization of fully determined systems and hence will not be used here. To suppress this warning pass warn_initialize_determined = false. To make this warning into an error, pass fully_determined = true
└ @ ModelingToolkit ~/.julia/packages/ModelingToolkit/aau6A/src/systems/diffeqs/abstractodesystem.jl:1520
retcode: InitialFailure
Interpolation: specialized 4rd order "free" stiffness-aware interpolation
t: 1-element Vector{Float64
}:
 0.0
u: 1-element Vector{Vector{Float64}}:
 [2.0, 0.0, 0.0, 1.0, 0.0]
```

test/mtk.jl
```diff
@@ -32,8 +34,116 @@ prob = ODEProblem(sys, u0, tspan, p, jac = true)
...
+mtkparams_timedepu0 = SciMLSensitivity.parameter_values(prob_incorrectu0)
+
+u0_correct = [D(x) => 2.0,
+    x => 1.0,
+    y => 0.0,
+    z => 0.0,
+    w2 => -1.0,]
+prob_correctu0 = ODEProblem(sys, u0_correct, tspan, p, jac = true, guesses = [w2 => -1.0])
+mtkparams_correctu0 = SciMLSensitivity.parameter_values(prob_correctu0)
+prob_correctu0.u0[5] = -1.0
```
ChrisRackauckas, line 92: @AayushSabharwal note that this is the issue of not passing over the initial values and guess values to u0. Without this `CheckInit` will fail. It's a minor issue but it is something we should handle over time.

AayushSabharwal: Yeah I'm working on using the hook in DiffEqBase. It's easy enough to implement, but _somehow_ the returned problem's u0 isn't being respected

test/mtk.jl
```diff
@@ -32,8 +34,119 @@ prob = ODEProblem(sys, u0, tspan, p, jac = true)
 sol = solve(prob, Tsit5())
 mtkparams = SciMLSensitivity.parameter_values(sol)
 
-gt = rand(5501)
+gt = rand(1495)
```
ChrisRackauckas, line 37:
```suggestion
gt = rand(length(sol.u))
```

src/adjoint_common.jl
```diff
@@ -737,16 +749,16 @@ function out_and_ts(_ts, duplicate_iterator_times, sol)
...
+#     Zygote.@adjoint function Zygote.literal_getproperty(sol::AbstractTimeseriesSolution,
+#             ::Val{:u})
+#         function solu_adjoint(Δ)
+#             zerou = zero(sol.prob.u0)
+#             _Δ = @. ifelse(Δ === nothing, (zerou,), Δ)
+#             (SciMLBase.build_solution(sol.prob, sol.alg, sol.t, _Δ),)
+#         end
+#         sol.u, solu_adjoint
+#     end
+# end
```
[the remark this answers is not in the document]
DhairyaLGandhi: I made a mistake pushing https://github.com/SciML/SciMLSensitivity.jl/pull/1168/commits/cdaa2c70c7e89050f10cd3731b70311a4570b05c so have reverted it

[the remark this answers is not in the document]
DhairyaLGandhi: Apologies

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "SciMLSensitivity"
 uuid = "1ed8b502-d754-442c-8d5d-10ac956f44a1"
 authors = ["Christopher Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
-version = "7.77.0"
+version = "7.76.0"
```
ChrisRackauckas, line 4: ?

DhairyaLGandhi: Ah this was picked from the history, updated. Thanks

ChrisRackauckas (approved):

PR #1180, opened 2025-04-03
Fix gradients for parameters getting dropped

DhairyaLGandhi:
Needs https://github.com/SciML/SciMLBase.jl/pull/979 and potentially a change in DifEqBase to pass `sensealg` in the kwargs in https://github.com/SciML/DiffEqBase.jl/blob/0874ff4ef8df2c74c843ce534e64ab3cdd40efb4/src/solve.jl#L1126

Currently UDEs fail to converge with reverse mode. This is due to dropped gradients for the parameters while solving the adjoint problem. We also definitely need to reduce the number of places where the assumptions SciML adjoints makes differs from the underlying AD. An example is returning the `Vector{Vector{Float64}}` as the gradient for a `sol` object instead of the `Tangent` or `NamedTuple` (like in Zygote) for the struct we are calculating gradients for. 

## Checklist

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -415,9 +415,8 @@ function DiffEqBase._concrete_solve_adjoint(
         (:callback_adj, :callback))}(values(kwargs))
     isq = sensealg isa QuadratureAdjoint
 
-    igs, new_u0, new_p = if _prob.f.initialization_data !== nothing
-        local new_u0
-        local new_p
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Conflicts

DhairyaLGandhi:
Can largely be closed since the `inittype` and the parameter gradient accumulation was covered in #1168 (and improved upon).

Only change of note here is the removal of the `literal_getproperty`, and the way `back` is called, which can be factored in separately

PR #1192, opened 2025-04-30
Bump PreallocationTools compat bound

jClugstor:
## Checklist


ChrisRackauckas:
Doesn't look like it improved the test it was supposed to fix?

jClugstor:
Yeah turns out this only fixes the specific case where the Dual tags are Nothing, but sometimes they're the OrdinaryDiffEqTag so the `get_tmp` still makes a ReinterpretedArray unfortunately. I think any mismatched tag is going to cause a ReinterpretedArray from `get_tmp`. 

I think we'll need to use LBC with that in mind then.

PR #1208, opened 2025-05-19
Take derivative at correct spot in `SteadyStateAdjoint`

jClugstor:
## Checklist

## Additional context
I noticed that we don't take the derivative in the correct spot in `SteadyStateAdjoint`, which causes issues with derivatives of large problems: fixes
https://github.com/SciML/SciMLSensitivity.jl/issues/1207

Essentially, we were taking the derivative of `f` at 0 always, since `lambda` comes from `zero(y)`, and the `StatefulJacobianOperator` was constructed with `lambda`. This meant that the adjoint equation was being set up incorrectly. 

Should probably add some tests for this too.


ChrisRackauckas:
Lambda will be updated because in a Krylov iteration it is aliased to u, but yes that is still not right. It should always be the derivative at the solve point, which is `y`.

Add a test.

ChrisRackauckas (approved):

PR #1214, opened 2025-05-24
Try to make doc build more robust

ChrisRackauckas:


ChrisRackauckas:
@frankschae do you know why the shadowing can give these Infs/NaNs sometimes?

PR #1217, opened 2025-06-06
SteadyStateAdjoint fix for applying `dp` twice

jClugstor:
## Checklist

## Additional context

Fixes https://github.com/SciML/SciMLSensitivity.jl/issues/1210 ?


ChrisRackauckas:
Add a test?

jClugstor:
Note: the reason that the current tests didn't catch this is because this only occured when the loss function includes `getindex`. If the loss function just returns the solution, `Δ` is an array or a number, so it doesn't hit the problematic branch. With `getindex` in the loss function `Δ` is a structural tangent, so it goes through the problematic branch.

PR #1223, opened 2025-06-22
Fix use of remake_zero! with callback affect differentiation

ChrisRackauckas:


ChrisRackauckas:
Handled in https://github.com/SciML/SciMLSensitivity.jl/pull/1349

ChrisRackauckas-Claude:
## Analysis from rebase attempt

Rebased this onto current master as PR #1350 and ran CI. The callback tests (Callbacks1, Callbacks2) fail on all Julia versions.

The issue is **not** about `Enzyme.remake_zero!` being broken. It's a **type mismatch**: the pre-allocated `_tmp6` from `S.diffcache.paramjac_config` is typed as `Void{ODEFunction{...}}`, but for `CallbackSensitivityFunction` the `f` is a `CallbackAffectWrapper`, so `Enzyme.Duplicated(SciMLBase.Void(f), _tmp6)` fails because the two arguments have different parameterized types.

The workaround `_tmp6 = Enzyme.make_zero(SciMLBase.Void(f))` is necessary because it creates a shadow with the **correct matching type** each time.

The proper fix would be to initialize `_tmp6` in `callback_tracking.jl` using the `CallbackAffectWrapper`-typed function so the pre-allocated shadow matches the type used at call time. This would allow `remake_zero!` to work.

PR #1286, opened 2025-11-10
Enzyme: Fix runtime activity errors

wsmoses:
This fixes several errors where EnzymeVJP failed.

Specficially the way Julia base handles broadcasting causes alias issues and therefore runtime activity problems. A copyto fixes this.

Additionally we can support cases where runtime activity is required.

### DIFF
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -6,9 +6,6 @@
 const have_not_warned_vjp = Ref(true)
 const STACKTRACE_WITH_VJPWARN = Ref(false)
 
-using Enzyme
-Enzyme.Compiler.VERBOSE_ERRORS[] = true
-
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Lots of test failures, I'll need to fix master maybe.

src/concrete_solve.jl
```diff
@@ -28,8 +42,8 @@ function inplace_vjp(prob, u0, p, verbose, repack)
         end
         false
     end
-    if ez
-        return EnzymeVJP()
+    if erz
+      return EnzymeVJP(; mode=Enzyme.set_runtime_activity(Enzyme.Reverse))
```
ChrisRackauckas: why would we do that? We generally don't want to use runtime activity?

PR #1304, opened 2025-12-19
Enzyme: Fix runtime activity errors

ChrisRackauckas-Claude:
## Summary
This PR adds a fallback to use Enzyme with `set_runtime_activity(Reverse)` mode when regular Enzyme autodiff fails. This helps handle cases where Enzyme encounters runtime activity errors.

Changes:
- Add a second try block (`erz`) that attempts Enzyme with runtime activity mode as a fallback
- Return `EnzymeVJP(; mode=Enzyme.set_runtime_activity(Enzyme.Reverse))` when the fallback succeeds
- Updates to derivative_wrappers.jl, gauss_adjoint.jl, quadrature_adjoint.jl, and sensitivity_algorithms.jl to support the mode parameter

This is a rebased version of #1286 (by @wsmoses) onto the current master.

Supersedes #1286

### DIFF
--- a/test/automatic_sensealg_choice.jl
+++ b/test/automatic_sensealg_choice.jl
@@ -21,4 +21,4 @@ ts = Float32.(collect(0.0:0.01:tspan[2]))
 prob = ODEProblem(dxdt_, x0, tspan, θ)
 _, repack, _ = SciMLStructures.canonicalize(SciMLStructures.Tunable(), p)
 sensealg = SciMLSensitivity.automatic_sensealg_choice(prob, x0, θ, true, repack)
-@test sensealg isa InterpolatingAdjoint && sensealg.autojacvec isa EnzymeVJP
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
## Additional fixes in latest commit

1. **EnzymeVJP constructor** - Added single-argument constructor `EnzymeVJP(chunksize::Int)` for backwards compatibility. The test in `automatic_sensealg_choice.jl` expects `EnzymeVJP(0)` but the struct now requires both `chunksize` and `mode` arguments.

2. **gclosure4 autodiff call** - Fixed the Enzyme.autodiff call order for the out-of-place case in `quadrature_adjoint.jl`. The `Enzyme.Const` return type annotation was in the wrong position (after the function args instead of right after the function annotation).

The Enzyme autodiff signature should be:
```julia
Enzyme.autodiff(mode, func_annotation, ReturnType, args...)
```

PR #1411, opened 2026-04-09
Add Sensitivity Methods Review Paper to Docs

facusapienza21:
Hi developers and contributors! 

I though it will be a good idea to add the [review paper](https://arxiv.org/abs/2406.09699) that some of us wrote some time ago to the documentation of SciMLSensitivity, as it is super pertinent for this and can be a very useful resource for newcommers. 

I limited myself to add the reference in some strategic places, together with other existing references. I also think it will be a great idea to include some figures we currently have in the paper to add some visuals to the docs, which always help. I am in particularly thinking about the following two figures: 

image
image

I am more than happy to work on including these or any other change that will benefit the clarity of the docs, specially thinking on new users!!! Do you have any suggestions here @ChrisRackauckas @JordiBolibar ?

Cheers

## Checklist


ChrisRackauckas:
Yes the figures could be very helpful

JordiBolibar:
The decision tree to choose methods depending on the problem's characteristics might be useful as well, but it might become outdated at some point as the ecosystem moves forward.

facusapienza21:
Yes, I thought about that too @JordiBolibar . I am happy to include it, but I think the table is a better way to go. 

I made the changes @ChrisRackauckas , but please let me know if you would change something as you are far more familiar with the docs. I included the images in assets (let me know if this is wrong).

PR #1413, opened 2026-04-10
Fix BoundsError in InterpolatingAdjoint when checkpoint cpsol has 1 point

ChrisRackauckas-Claude:
## Summary
- `split_states` for `ODEInterpolatingAdjointSensitivityFunction` recomputes a forward sub-solve between two adjacent checkpoint times and uses `abs(cpsol_t[end] - cpsol_t[end - 1])` to seed the integrator's `dt`. When the previous checkpoint sub-solve only saved a single time point (e.g. because the adaptive forward solve took one step that already reached the interval end), `cpsol_t[end - 1]` becomes `cpsol_t[0]` and throws `BoundsError: attempt to access 1-element Vector at index [0]`.
- Adds a `checkpoint_dt(cpsol_t, interval)` helper that falls back to the interval width when fewer than two saved time points are available, and uses it from all three ODE branches in `split_states`.
- Guards the SDE branch's `cpsol_t[1] - cpsol_t[2]` access with the same length check.
- Adds `test/checkpoint_single_step.jl` (also wired into `runtests.jl` Core1) which reproduces the failure on master and passes with the fix.

## Repro
This is the exact failure surfacing in [SciML/Optimization.jl#1169 CI](https://github.com/SciML/Optimization.jl/actions/runs/23171032225/job/67322714897?pr=1169) (and on Optimization.jl `master` since at least 2026-04-08, run `24144818227`), where `test/minibatch.jl` calls `adjoint_sensitivities(...; sensealg = InterpolatingAdjoint())` against a forward solve produced with tightly-spaced `saveat`. Stack trace:

```
BoundsError: attempt to access 1-element Vector{Float32} at index [0]
  [3] split_states(du, u, t, S::ODEInterpolatingAdjointSensitivityFunction{...})
      @ SciMLSensitivity src/interpolating_adjoint.jl:241
```

Line 241 in v7.99.0 is `cpsol′ = solve(... dt = abs(cpsol_t[end] - cpsol_t[end - 1]) ...)` — same expression and same bug on current master (still at lines 243/253/261 of `src/interpolating_adjoint.jl`). Reproduced locally on Julia 1.11 with the patched and unpatched package and the new regression test fails on master and passes here.

## Test plan
- [x] `test/checkpoint_single_step.jl` reproduces the BoundsError on master, passes after the fix
- [x] `Optimization.jl/test/minibatch.jl` runs to completion against the patched SciMLSensitivity (was previously erroring out at the `loss_grad` call)
- [ ] Full SciMLSensitivity CI

### DIFF
--- a/src/interpolating_adjoint.jl
+++ b/src/interpolating_adjoint.jl
@@ -132,6 +132,19 @@ function choose_dt(dt, ts, interval)
     return dt
 end
 
+# Compute a safe initial dt from a checkpoint solution's time vector. The
+# checkpoint solution may contain only a single saved point (e.g. when an
+# adaptive forward solve takes one step that already reaches the interval
+# end), in which case `cpsol_t[end - 1]` would be a `BoundsError`. Fall back
+# to the interval width in that case.
+function checkpoint_dt(cpsol_t, interval)
+    if length(cpsol_t) >= 2
+        return abs(cpsol_t[end] - cpsol_t[end - 1])
+    else
+        return abs(interval[2] - interval[1])
+    end
+end
+
 # u = λ'
 # add tstop on all the checkpoints
 function (S::ODEInterpolatingAdjointSensitivityFunction)(du, u, p, t)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
Closing per maintainer feedback. The minibatch BoundsError is being addressed at the test level by switching to Float64 in SciML/Optimization.jl#1181, which makes the bad cpsol-shrinkage codepath unreachable from current SciML tests. The right fix is to make the test more robust, not the dt heuristic.

PR #1432, opened 2026-04-26
Bump compat for OrdinaryDiffEq v7 / SciMLBase v3 ecosystem

ChrisRackauckas-Claude:
## Summary

Bumps compat bounds and migrates breaking API usages for the OrdinaryDiffEq v7 / SciMLBase v3 ecosystem, as introduced in SciML/OrdinaryDiffEq.jl#3562 and SciML/OrdinaryDiffEq.jl#3565 and documented in [NEWS.md](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/NEWS.md).

### Compat bumps
- `DelayDiffEq`: `"5.43.2"` → `"5.43.2, 6"` (main `Project.toml`)
- `DelayDiffEq`: `"5"` → `"5, 6"` (`docs/Project.toml`)
- `StochasticDiffEq`: `"6.69"` → `"6.69, 7"` (main `Project.toml`)
- `StochasticDiffEq`: `"6.56"` → `"6.56, 7"` (`docs/Project.toml`)
- `OrdinaryDiffEq`: `"6.31"` → `"6.31, 7"` (`docs/Project.toml`; main already had v7)

### Source migrations

- `u_modified!(integrator, true)` → `derivative_discontinuity!(integrator, true)` in `src/adjoint_common.jl` and `src/backsolve_adjoint.jl`. `u_modified!` is deprecated in SciMLBase v3 (replaced by `derivative_discontinuity!`).
- `prob_func(prob, i, repeat)` → `prob_func(prob, ctx)` (using `ctx.sim_id`) in `test/ensembles.jl`, `test/distributed.jl`, `test/forward_remake.jl`, `test/sde_neural.jl`.
- `output_func(sol, i)` → `output_func(sol, ctx)` in `test/forward_remake.jl`.  
  The old 3-argument ensemble callback signatures are no longer supported in SciMLBase v3; `EnsembleContext` carries trajectory index as `ctx.sim_id`.

## Test plan

- [ ] CI passes on full test suite
- [ ] Verify `SciMLSensitivity` loads against new ecosystem versions
- [ ] Ensemble tests run correctly with new context-based callbacks
- [ ] Adjoint backpropagation tests pass with `derivative_discontinuity!`

References: SciML/OrdinaryDiffEq.jl#3562, SciML/OrdinaryDiffEq.jl#3565

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1441, opened 2026-05-08
VCC reverse: compose simultaneous fires through one wrapper VJP

ChrisRackauckas-Claude:
> [!NOTE]
> Draft — please ignore until @ChrisRackauckas reviews. Posted as the follow-up surgery requested in https://github.com/SciML/SciMLSensitivity.jl/pull/1432.

## Context

After #1432 landed, `TrackedAffect` for `VectorContinuousCallback` (VCC) under the new DiffEqBase v7 dispatch unpacked each fire's mask into one stored row per fired component. The reverse `PresetTimeCallback` then fired once per component, looking up a diffcache keyed by `(pos_neg, single-Int)` and replaying that single component's affect on top of `uleft`.

That replay strategy is correct only when components mutate disjoint state. For multi-component fires that touch overlapping state (\"events you get can be dependent on multiple firing at the same time\"), it produces the wrong Jacobian because each iteration uses `uleft` instead of the chained intermediate state.

## Change

- `TrackedAffect.event_idx` for VCC now stores the verbatim per-fire mask (`Vector{Int8}`) — one row per actual fire.
- `CallbackAffectWrapper` / `CallbackAffectPWrapper` iterate the stored mask forward and call the user's affect for each fired component in order. The VJP through that combined wrapper produces the correct composed Jacobian regardless of dependency between components.
- `cb_diffcaches` is keyed by per-fire index for VCC, since each fire's mask may differ.
- The Reactant and Mooncake callback-VJP kernels iterate the mask in their `cb_state_fn` / `cb_param_fn` closures so the same shape works there.
- For the implicit time-derivative correction (`dgdt`), all simultaneously firing conditions share the same τ, so the first fired component is used as the representative for ∇τ.

## Status

- The pre-existing GaussAdjoint / BacksolveAdjoint mismatch in `test/callbacks/vector_continuous_callbacks.jl` (\"callback with linear affect\" set, dstuff[3:4]) reproduces identically on plain `master` 4b425b6 — not introduced by this PR. The \"Test condition function that depends on time only\" set (4 tests) still passes.
- Other backends (Reactant, Mooncake) are not exercised by that test file.
- CI run will give the broader signal.


Project.toml
```diff
@@ -66,7 +66,7 @@ ChainRulesCore = "0.10.7, 1"
 ComponentArrays = "0.15.5"
 ConstructionBase = "1.5"
 DelayDiffEq = "5.43.2, 6"
-DiffEqBase = "6.175, 7"
+DiffEqBase = "6.175, 7.3.1"
```
ChrisRackauckas:
```suggestion
DiffEqBase = "7.4.0"
```

Issue #1415, opened 2026-04-10
Add Enzyme-native method for _init_originator_gradient (init-path AD dispatch)


ChrisRackauckas-Claude:
## Background

#1414 introduces `_init_originator_gradient(originator, f, tunables)`, an originator-dispatched helper for the gradient of `get_initial_values` w.r.t. tunables in the DAE/ODE init path inside `_concrete_solve_adjoint`. After that PR:

- **Default** method (in `src/concrete_solve.jl`) uses `Zygote.pullback` — handles `ChainRulesOriginator`, `EnzymeOriginator`, `ReverseDiffOriginator`, `TrackerOriginator` by fallthrough.
- **`MooncakeOriginator`** method (in `ext/SciMLSensitivityMooncakeExt.jl`) uses `Mooncake.build_rrule` + `Mooncake.value_and_gradient!!` natively.

So Enzyme-driven AD currently still routes through Zygote at this one step, even though Enzyme is a hard dep of SciMLSensitivity and the rest of `_concrete_solve_adjoint` already specializes on `EnzymeOriginator` (lines 944, 967, 974, etc.).

## Ask

Add an Enzyme-native method:

```julia
function _init_originator_gradient(::SciMLBase.EnzymeOriginator, f, tunables)
    # Enzyme.gradient(Enzyme.Reverse, f, tunables) or equivalent,
    # returning the gradient in the same shape/type as `tunables`.
end
```

It can live directly in `src/concrete_solve.jl` since `Enzyme` is already a hard dep — no extension needed.

## Blockers / things to verify before flipping the dispatch

The desauty test (`test/desauty_dae_mwe.jl`, the canonical case for #1414) currently marks `Enzyme through init` as `@test_broken` for both `use_scc=false` and `use_scc=true`, with this comment:

> Broken due to multiple upstream Enzyme/NonlinearSolve issues:
> - Julia 1.12: `NonlinearSolveBaseEnzymeExt` rules disabled (`VERSION < v\"1.12\"` guard) causing LLVM crash in GC invariant verifier
> - Julia 1.10: `EnzymeMutabilityException` from MTK's `remake` (mutable closure), `MixedReturnException` with default `PolyAlgorithm`, and `NamedTuple` broadcasting error in `NonlinearSolveBaseEnzymeExt` reverse rule with `MTKParameters`
>
> See: NonlinearSolve.jl#869, Enzyme.jl#2699

[...]

PR #1467, opened 2026-06-03
Enzyme: route MTK DAE init gradient through Zygote; enable Enzyme-outer mtk.jl test

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft, opened by an agent.

Makes **Enzyme reverse-mode on the outside of an MTK DAE solve** work end-to-end (the `test/mtk.jl` DAE-initialization consistency test), and fixes the underlying state-corruption.

## Root cause

`_init_originator_gradient(::EnzymeOriginator)` ran a **nested `Enzyme.gradient`** inside the outer Enzyme reverse pass. Nested Enzyme (under `set_runtime_activity`) corrupts Enzyme's global compilation/runtime state: the **first** outer `Enzyme.autodiff` through an MTK DAE-init problem succeeds, but **every subsequent call** throws `LinearAlgebra.SingularException` out of the semi-explicit DAE adjoint (`adjoint_common.jl`). Minimal repro — the *same* setup differentiated 4× in one process:

```
iter 1: OK    grad = [-88324.8, 46528.9, 99324.9]
iter 2: FAILED LinearAlgebra.SingularException
iter 3: FAILED LinearAlgebra.SingularException
iter 4: FAILED LinearAlgebra.SingularException
```

This is exactly the failure mode of the multi-setup `mtk.jl` test (each setup passes in isolation; the sweep fails).

## Change

- **`_init_originator_gradient(::EnzymeOriginator)` → `Zygote.pullback`** (the existing `ADOriginator` fallback). It runs as opaque primal code in the rrule backpass and is **not** differentiated by the outer Enzyme, so the nesting — and the corruption — is gone. (The init sub-VJP it calls can still be `EnzymeVJP`; only the *outer* init differentiation moves off Enzyme.) Undoes the deliberate Enzyme-native attempt from #1415 — flagging for review.
- **`test/mtk.jl`: `@test_broken` → `@test`.** All 14 init-algorithm setups (CheckInit / BrownFullBasicInit / DefaultInit / NoInit / OverrideInit / overdetermined × correct/incorrect/timedep/overdetermined u0) now give a consistent gradient.
- **Inner VJP `EnzymeVJP` → `ReverseDiffVJP`** in `mtk.jl`. `EnzymeVJP._vecjacobian!` runs its *own* nested `Enzyme.autodiff` on the MTK-generated DAE RHS, which both mishandles that RHS for some DAEs and is the *second* state-corruption source. A non-Enzyme inner VJP is the established workaround; **the outer AD is still Enzyme**, which is what the test exercises. (Fixing EnzymeVJP's nested Enzyme — and the upstream Enzyme nested-reverse state issue — are follow-ups.)
- **compat:** `NonlinearSolveBase 2.28 → 2.30.2` (the `inactive_kwarg`, `within_autodiff` wrap-skip, and `inactive_type` Enzyme init/solve_up fixes — `inactive_type` landed in NonlinearSolve#947). `DiffEqBase` already requires `≥ 7.5.5`.

## Verification (local, Julia 1.11)

- `parameter_initialization.jl` "Adjoint through Parameter Initialization": **3/3 pass** with the Zygote-init change (no regression).
- `mtk.jl` 14-setup sweep with Zygote-init + ReverseDiffVJP: all setups consistent (`[-88425.05, 46558.38, 99481.47]`), including every `DefaultInit` setup that previously `MethodError`'d, and the `nothing`/OverrideInit setups that previously `SingularException`'d under the multi-setup sweep.
- 4× repeat of one setup: all 4 pass (was iter-1-only before).

The `mtk.jl` consistency check compares init methods against each other (not vs ForwardDiff); the large values come from the near-chaotic tspan-100 Lorenz adjoint, but all methods agree.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLSensitivity"
 uuid = "1ed8b502-d754-442c-8d5d-10ac956f44a1"
-version = "7.111.1"
+version = "7.111.2"
 authors = ["Christopher Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
 
 [deps]
@@ -92,7 +92,7 @@ Mooncake = "0.5.24"
 Reactant = "0.2.22"
 NLsolve = "4.5.1"
 NonlinearSolve = "3.0.1, 4"
-NonlinearSolveBase = "2.28"
+NonlinearSolveBase = "2.30.2"
 SCCNonlinearSolve = "1"
 Optimization = "4, 5"
 OptimizationOptimisers = "0.3"
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -356,37 +356,26 @@ function _init_originator_gradient(::SciMLBase.ADOriginator, f, tunables)
     return back(one(iy))[1]
 end
 
-# Enzyme-native init-path gradient. `Const(f)` is required because the
-# closure built by the caller captures references to `_prob` / `repack` /
-# `kwargs_init`, which Enzyme would otherwise reject with
-# `EnzymeMutabilityException` ("Function argument passed to autodiff
-# cannot be proven readonly").
+# Init-path gradient for the `EnzymeOriginator` (outer-Enzyme) case.
 #
-# Note: this routes the init-step gradient through Enzyme's reverse mode
-# even when the user is only using Enzyme indirectly (e.g. via
-# `InterpolatingAdjoint(autojacvec = EnzymeVJP())` driven by Zygote). The
-# Mooncake-equivalent path is restricted to true `MooncakeOriginator`
-# callers; here we trust that anyone whose rrule was reached with
-# `originator::EnzymeOriginator` is in fact running Enzyme on the outside
-# and would prefer Enzyme-native errors over silently routing through
-# Zygote. Upstream Enzyme/NonlinearSolve issues affecting DAE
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
Local verification (Julia 1.11, `test/mtk.jl` run in full):

- **`test/mtk.jl` exits 0, no failures.** Main Enzyme DAE-initialization `@test` is consistent across all 14 init-algorithm setups (CheckInit / BrownFullBasicInit / DefaultInit / NoInit / OverrideInit / overdetermined × correct/incorrect/timedep/overdetermined u0) → gradient `[-88425.05, 46558.38, 99481.47]` for every setup. The Mooncake block (now `@test`) returns a correct nonzero gradient.
- **`parameter_initialization.jl` "Adjoint through Parameter Initialization": 3/3 pass** — the Zygote-routed init gradient is a no-regression change there.
- Root-cause MWE (the nested-Enzyme corruption this fixes): the *same* setup differentiated 4× in one process gives `iter 1 OK, iter 2+ SingularException` without this change, and `4/4 OK` with it.

Follow-ups (not in this PR): the inner `EnzymeVJP._vecjacobian!` runs its own nested `Enzyme.autodiff` on the MTK DAE RHS (the reason the inner VJP here is `ReverseDiffVJP`), and the underlying Enzyme nested-reverse state-corruption is an upstream issue.

PR #1483, opened 2026-06-11
Fix Enzyme.Duplicated type mismatch with ComponentVector views

ChrisRackauckas-Claude:
This PR fixes a `MethodError` in `Enzyme.Duplicated` when using `EnzymeVJP()` with `ComponentVector` parameters that are views (e.g. nested `ComponentVector`s).

Enzyme requires the primal and shadow values to have identical types. When `SciMLSensitivity` allocates a shadow buffer, it often creates a full `Vector`-backed `ComponentVector`, while the primal might be a `SubArray`-backed view. This mismatch triggers a `MethodError`.

The fix:
1. Refines the `trivial_repack` logic to correctly identify when a structure needs repacking to match its primal type.
2. Ensures type consistency by using `copy(p)` when a type mismatch is detected between the primal `p` and shadow `_sp` in the `EnzymeVJP` path.

This resolves compatibility issues in tutorials like `feedback_control.md` when using Enzyme.

Included a new test case: `test/enzyme_componentarrays.jl`.

Note: This PR should be ignored until reviewed by @ChrisRackauckas.

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -1,3 +1,12 @@
+# Wraps a pre-allocated dense primal buffer for use with Enzyme.Duplicated when the
+# parameter is a view-backed array (e.g. ComponentArray with SubArray data). Enzyme
+# requires primal and shadow to have the same concrete type; since the shadow is always
+# a fresh dense allocation, the primal must also be dense. We pre-allocate once here
+# and copyto! the current parameter values at each adjoint step.
+struct EnzymeViewPrimalBuffer{T}
+    buf::T
+end
+
 struct AdjointDiffCache{
         UF, PF, G, TJ, PJT, uType, JC, GC, PJC, JNC, PJNC, rateType, DG1,
         DG2, DI,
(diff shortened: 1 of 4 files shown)


ChrisRackauckas-Claude:
Updated the branch with a root cause fix (replacing the previous band-aid).

**Root cause**: When `p` is a `ComponentArray` view (SubArray-backed), `zero(p)` returns a `Vector`-backed `ComponentArray` — a different concrete type. `Enzyme.Duplicated(p, zero(p))` threw a `MethodError` because primal and shadow had different types.

**Fix**: Add `EnzymeViewPrimalBuffer` struct. At adjoint cache build time (`adjointdiffcache`), detect when `tunables` is a view-backed array via `typeof(tunables) !== typeof(zero(tunables))`. In that case, pre-allocate a dense primal buffer (`copy(tunables)`) and store it as `_shadow_p` (same slot used for non-`AbstractArray` SciML struct shadows). At each adjoint step in `_vecjacobian!`, `copyto!` the current `p` values into this buffer and use it as the `Enzyme.Duplicated` primal — leaving `tmp2 = zero(tunables)` (already dense) as the shadow. Both are then the same concrete type.

This is the same one-allocation-at-cache-build-time pattern already used for non-`AbstractArray` SciML structures via `_needs_shadow → repack(zero(tunables))`.

PR #1509, opened 2026-06-28
Fix Downgrade CI: bump stale AD-backend compat floors

ChrisRackauckas-Claude:

## Problem
The `Downgrade` workflow has been red on `master` for 100+ consecutive runs. It fails at the `julia-actions/julia-downgrade-compat` `--min=@deps` resolution step; Resolver.jl reports only the unhelpful `ERROR: LoadError: Unsatisfiable`.

## Diagnosis
I reproduced the floor-pinned resolution locally on Julia 1.10.11 by building the merged (`[deps]` + test `[extras]`) project, pinning every non-stdlib dependency to its compat floor (`=floor`), and running the standard `Pkg.resolve()` — which gives human-readable conflicts. Three mutually-inconsistent lower bounds surfaced, all in the AD-backend/test ecosystem:

1. **Mooncake `0.5.24` → `0.5.25`.** `ComponentArrays >= 0.15.34` requires `Mooncake >= 0.5.25` via its Mooncake weak-extension compat (`["0.15.34 - 0"] Mooncake = "0.5.25 - 0.5"`), but the floor was `0.5.24`.
2. **Reactant `0.2.22` → `0.2.203`.** `RecursiveArrayTools 4.x` requires `MLDataDevices >= 1.17.9` (MLDataDevices only permits RAT v4 from `1.17.9` onward: `["1.17.9 - 1"] RecursiveArrayTools = "3.8.0 - 4"`). The ancient `0.2.22` Reactant floor transitively capped MLDataDevices below that. `0.2.203` is the minimum Reactant that resolves with the rest of the floor set. (`>= 0.2.165` is independently required by `Enzyme = 0.13.163`.)
3. **StableRNGs `1` → `1.0.4`.** `Reactant >= 0.2.208` requires `StableRNGs = 1.0.4`, so the `"1"` floor left part of the declared Reactant range unresolvable at the floor. Bumping to `1.0.4` keeps the whole compat range internally consistent.

These are all optional/test AD-backend deps, so core downgrade coverage (DiffEqBase, OrdinaryDiffEq, SciMLBase, …) is unchanged.

## Verification
Locally on Julia 1.10.11, pinning every non-stdlib dep to its compat floor:
- **Before:** `Unsatisfiable` (matches CI).
- **After (these 3 bumps):** merged-project resolution **succeeds**.

This fixes the resolution step that the Downgrade job dies on. The downgrade test-execution phase (Core1 at floor versions) will be validated by CI on this PR.

## Note
`master` also has unrelated failures in the main `CI` workflow (several Core groups fail in `Stiff Adjoints` with a `BoundsError` originating in `SciMLBase`'s `ext/SciMLBaseZygoteExt.jl` `ODESolution_getindex_pullback`, plus an SDE2 numerical failure and Core1 timeouts). Those are separate from this resolution fix and are being investigated independently.


ChrisRackauckas-Claude:
Closing this — after deeper investigation the floor bumps here do **not** fix the Downgrade job, because the root cause is not in `[compat]`.

## Actual root cause: a Resolver.jl bug, not stale floors

`SciML/.github`'s `downgrade.yml` runs `julia-actions/julia-downgrade-compat`, which now resolves the minimum-version manifest with **Resolver.jl** (`StefanKarpinski/Resolver.jl`, the experimental SAT resolver) via `resolve.jl --min=@deps`.

I reproduced the CI flow locally on Julia 1.10.11 (built the same merged `[deps]`+test-`[extras]` project `create_merged_project` builds, ran the same `resolve.jl`). Findings:

- The job dies at the resolve step with the opaque `ERROR: LoadError: Unsatisfiable`. Instrumenting `resolve.jl`'s final check shows the unsatisfiable package is reported as **Mooncake**.
- But it is **not** a real version conflict. Running the **standard Pkg resolver** (`Pkg.resolve()`) on the *exact same merged project* succeeds — both at the floors and at latest (Mooncake 0.5.33, Reactant 0.2.269).
- **Resolver.jl reports the merged project `Unsatisfiable` at *every* version combination — including latest/`--max`** — where `Pkg.resolve` finds a valid assignment. So no `[compat]` floor change can make Resolver.jl resolve this graph; it cannot resolve it at any versions.

That is, the failure is a Resolver.jl defect on SciMLSensitivity's (extension-heavy) dependency graph, not a floor-bump problem. This matches the history: the Downgrade job has been red for 100+ consecutive runs since strict downgrade was re-enabled (#1474), and the earlier floor-bump PRs (#1481, #1491) did not turn it green either.

## Suggested paths (none of them this PR)
- Report the minimal repro to `julia-actions/julia-downgrade-compat` / `StefanKarpinski/Resolver.jl`.
- Or, in `SciML/.github`'s `downgrade.yml`, pin `julia-downgrade-compat` to a revision that uses the Pkg-based resolver instead of Resolver.jl (affects all SciML repos — needs a deliberate decision).
- Or add the extension-trigger weakdeps to the action's skip list if that sidesteps the Resolver.jl path.

Leaving the floors unchanged here since bumping them does not fix CI and would be unvalidated against the real (Pkg) resolver's minimums.

ChrisRackauckas-Claude:
(Correction: I see this was already merged before my analysis above landed — so the three floor bumps *are* now on master. They're harmless to the normal CI lanes, since those resolve every dep to latest regardless of the floor. But per the analysis above they will **not** turn the Downgrade job green — that requires addressing the Resolver.jl resolve-step bug at the workflow/tooling level, not `[compat]`.)

PR #1527, opened 2026-07-05
Mark p Const in EnzymeVJP _vecjacobian! when no dgrad is requested

ChrisRackauckas-Claude:
**Note: please ignore this PR until reviewed by @ChrisRackauckas.**

## What

The Gauss/Quadrature adjoint ODE rhs calls `vecjacobian!` with `dgrad === nothing` on every solver stage — the parameter gradient there is not wanted (it comes from the quadrature). But the in-place EnzymeVJP branch of `_vecjacobian!` chose `Duplicated(p, shadow)` based only on the shadow buffer's existence, so Enzyme zeroed a p-sized shadow and accumulated a full parameter VJP on every stage, then discarded it.

This PR skips that discarded accumulation by demoting `p` to `Const`, **gated on `runtime_activity(mode)`**:

- Static-mode users (default `EnzymeVJP()`) keep exactly the previous `Duplicated(p, shadow)` behavior — no behavior or performance change.
- Runtime-activity modes — **including the automatically selected `EnzymeVJP(mode = set_runtime_activity(Reverse))` from `automatic_sensealg_choice`, i.e. the default Zygote-over-`solve` path** — get the demotion.

The gate is required for correctness: the first commit demoted unconditionally, and CI showed that under static activity analysis an rhs that stores p-derived array references into active computation (e.g. a Lux layer reshaping a `ComponentArray`'s weights) either raises `EnzymeRuntimeActivityError` (Core3 `default_alg_diff`, Core8 repeated adjoint) or **silently drops gradient terms** (Callbacks1 parameter-changing callback). Forcing runtime activity on instead was measured to slow small right-hand sides, so the demotion only applies when the user's mode already has runtime activity.

## Measured

32-state / 1024-parameter dense linear ODE, `GaussAdjoint`, Julia 1.10.11, Enzyme 0.13.171 (min of 10 samples):

| mode | before | after |
|---|---|---|
| `EnzymeVJP()` static | 0.58 ms | 0.58 ms (unchanged) |
| `EnzymeVJP(set_runtime_activity(Reverse))` | 0.58 ms | **0.46 ms (−21%)** |

Small-p (Lotka-Volterra): unchanged in both modes. Gradients bit-identical to a `ReverseDiffVJP` reference (relerr ≤ 5e-15) on both problems and both modes. The saving is the discarded p-VJP accumulation × ~6 stage calls/step, so it scales with `length(p)` — NN/MTK-sized parameters benefit most.

## Tests added

`test/Core3/allocation_regression.jl` (wired into the Core3 group):
1. **Per-call allocation ceiling** for the warmed `GaussIntegrand` `vec_pjac!` with EnzymeVJP (measured 64 B — all inside the `Enzyme.autodiff` entry; ceiling 128 B). Skipped under coverage.
2. **Per-step allocation slope** of full `adjoint_sensitivities` solves (T=10 vs T=100): ≤ 64 B/step for Gauss+Enzyme (measured ~24), ≤ 8 B/step for Interp+Enzyme (measured 0) — catches any accidentally introduced per-step/stage allocation. Skipped under coverage.
3. **Two-sided NaN-sentinel** (runs under coverage too): with a runtime-activity mode the parameter shadow stays untouched (`Const(p)` reached Enzyme); with the static default the shadow is written (`Duplicated` kept). The first side fails on master; the second side would fail on the unconditional demotion.

## Local validation (all against the gated commit)

- `Callbacks1/discrete_callbacks.jl`: **354/354** (this suite caught the silent gradient corruption of the unconditional version)
- `Core8/enzyme_vjp_repeated_adjoint.jl`: 3/3, `Core8/parameter_initialization.jl`: 3/3
- `Core3/default_alg_diff.jl` (auto-selection exercises the demotion): passes
- `Core3/gauss_zygote_inplace.jl`: 15/15
- New allocation regression tests: pass
- Direct gradient checks vs independent VJP on two benchmark problems: bit-identical

Pre-existing on master (not from this PR): QA (julia 1), Core4/SDE3/Core5 (julia 1.11) failures.

Background: the full adjoint performance analysis with allocation attribution is in https://github.com/ChrisRackauckas/InternalJunk/pull/61.

https://claude.ai/code/session_018uzZhKZV2ameiUL8ouYknd

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -1033,9 +1033,18 @@ function _vecjacobian!(
     # When the caller does not request a parameter gradient (`dgrad === nothing`,
     # as in the Gauss/Quadrature adjoint rhs on every solver stage), mark `p`
     # `Const` so Enzyme skips the parameter-gradient accumulation entirely,
-    # matching the out-of-place wrapper above.
+    # matching the out-of-place wrapper above. This is only sound when the mode
+    # has runtime activity: under static activity analysis, an rhs that stores
+    # `p`-derived array references into active computation (e.g. a Lux layer
+    # reshaping a `ComponentArray`'s weights) raises `EnzymeRuntimeActivityError`
+    # — or worse, silently drops gradient terms — once `p` is `Const`. So the
+    # demotion is gated on `runtime_activity(mode)` (already enabled on the
+    # automatically selected `EnzymeVJP`) rather than forcing runtime activity
+    # on, which measurably slows small right-hand sides.
+    _skip_p_grad = dgrad === nothing &&
+        Enzyme.EnzymeCore.runtime_activity(isautojacvec.mode)
     _shadow_p = nothing
-    dup = if dgrad !== nothing && !(tmp2 isa SciMLBase.NullParameters)
+    dup = if !_skip_p_grad && !(tmp2 isa SciMLBase.NullParameters)
         # tmp2 .= 0
         Enzyme.remake_zero!(tmp2)
         if _cached_shadow isa EnzymeViewPrimalBuffer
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
CI on 312ed37 (the gated commit): Core3, Callbacks1, and Core8 — the three groups the first commit broke — are all green. Remaining failures classified: SDE1 (1.11) and Core2 (1.11) are self-hosted-runner communication losses (infra flakes, annotations say "runner lost communication with the server"); QA (julia 1), SDE3 (1.11), and Downgrade Core1 fail on master too (pre-existing). No failures attributable to this PR.

Issue #1534, opened 2026-07-08
SimpleChains docs example fails with Enzyme illegal type analysis


ChrisRackauckas-Claude:
During local docs testing for SciML/SciMLSensitivity.jl#1490, `docs/src/examples/neural_ode/simplechains.md` failed while Documenter was executing example blocks.

Observed locally on Julia 1.12.6 during the full docs command:

```sh
timeout 3600 env JULIA_DEPOT_PATH="$PWD/.julia_depot" /home/crackauc/.juliaup/bin/julia +1.12.6 --project -e 'println("--- Instantiating project"); using Pkg; Pkg.instantiate(); Pkg.activate("docs"); Pkg.instantiate(); push!(LOAD_PATH, @__DIR__); println("+++ Building documentation"); include("docs/make.jl")'
```

Documenter reported the failing example at:

```text
docs/src/examples/neural_ode/simplechains.md:70-87
```

The error was:

```text
IllegalTypeAnalysisException: Enzyme compilation failed due to illegal type analysis
```

The stack involved `SimpleChains.SimpleChain`, `VectorizationBase.align`, and the example optimizer using:

```julia
OPT.AutoEnzyme(mode = Enzyme.set_runtime_activity(Enzyme.Reverse))
```

This file was not changed by #1490, so it appears to be an existing docs/example failure rather than a regression from that PR. It should be isolated from the #1490 doctest cleanup.

ChrisRackauckas-Claude:
The SimpleChains blocker now has a draft upstream fix: PumasAI/SimpleChains.jl#220.

The extension installs a package-level Enzyme reverse rule and avoids differentiating through SimpleChains' stack-backed implementation. A multi-layer VJP in both regular reverse mode and runtime-activity reverse mode exactly matched the existing ChainRules pullback. Once that PR is merged and released, the SimpleChains example can be moved to the Enzyme backend without relying on an Enzyme compiler workaround.

ChrisRackauckas-Claude:
[...]

Issue #1533, opened 2026-07-08
second_order_neural docs example segfaults during Documenter example execution


ChrisRackauckas-Claude:
From SciML/SciMLSensitivity.jl#1490, `docs/src/examples/ode/second_order_neural.md` exits with code 139 during Documenter example execution.

Observed locally on Julia 1.12.6. Documenter is running `@example` blocks here (`doctest=false`), so this is not a text doctest mismatch.

Reproduction command:

```sh
timeout 3600 env JULIA_DEPOT_PATH="$PWD/.julia_depot" /home/crackauc/.juliaup/bin/julia +1.12.6 --project -e 'using Pkg; Pkg.activate("docs"); push!(LOAD_PATH, pwd()); using Documenter, SciMLSensitivity; ENV["GKSwstype"]="100"; using Plots; makedocs(; root="docs", sitename="target", modules=[SciMLSensitivity], clean=true, doctest=false, linkcheck=false, warnonly=[:missing_docs], pages=["target" => "examples/ode/second_order_neural.md"])'
```

I observed exit code 139 both with the PR Enzyme-converted page and after reverting this page back to the original Mooncake form. The PR patch reverts the Enzyme conversion so the current PR can avoid adding another changed failing example, but this page likely needs a separate baseline investigation.

Relevant changed line in the PR before revert:

```julia
adtype = OPT.AutoEnzyme(; mode = Enzyme.set_runtime_activity(Enzyme.Reverse))
```

Reverted form:

```julia
adtype = OPT.AutoMooncake(; config = Mooncake.Config(; friendly_tangents = true))
```

Issue #1532, opened 2026-07-08
second_order_adjoints docs example fails under Enzyme conversion


ChrisRackauckas-Claude:
From SciML/SciMLSensitivity.jl#1490, the `docs/src/examples/ode/second_order_adjoints.md` example fails after the PR converts the first optimization from Mooncake to Enzyme.

Observed locally on Julia 1.12.6 with a PR checkout before reverting this page. Documenter runs example blocks here (`doctest=false`), so this is an `@example` failure rather than a text doctest mismatch.

Reproduction command:

```sh
timeout 3600 env JULIA_DEPOT_PATH="$PWD/.julia_depot" /home/crackauc/.juliaup/bin/julia +1.12.6 --project -e 'using Pkg; Pkg.activate("docs"); push!(LOAD_PATH, pwd()); using Documenter, SciMLSensitivity; ENV["GKSwstype"]="100"; using Plots; makedocs(; root="docs", sitename="target", modules=[SciMLSensitivity], clean=true, doctest=false, linkcheck=false, warnonly=[:missing_docs], pages=["target" => "examples/ode/second_order_adjoints.md"])'\n```\n\nThe Enzyme-converted version fails with:\n\n```text\nEnzymeRuntimeActivityError: Detected potential need for runtime activity\n```\n\nThe failure occurs around the first `OptimizationFunction` using `OPT.AutoEnzyme()`. The PR patch reverts this page back to the Mooncake version so the Enzyme conversion can be handled separately.\n\nNote: after reverting to Mooncake, the isolated page still later exits with code 139 on my machine, apparently in the later second-order optimization section. That looks like a separate baseline docs/runtime issue and should not be conflated with the Enzyme runtime-activity error tracked here.

PR #1490, opened 2026-06-18
Docs testing part 1

wsmoses:

PR #1535, opened 2026-07-08
Revert failing second-order docs conversions

ChrisRackauckas-Claude:
This PR stacks on top of #1490 and reverts the two second-order docs conversions that currently fail locally, so #1490 can keep the passing conversions separate from follow-up failures.

Ignore this PR until reviewed by @ChrisRackauckas.

Changes:
- Revert `docs/src/examples/ode/second_order_adjoints.md` from Enzyme back to Mooncake.
- Revert `docs/src/examples/ode/second_order_neural.md` from Enzyme back to Mooncake.

Isolated follow-up issues opened:
- #1532: `second_order_adjoints.md` Enzyme conversion fails with `EnzymeRuntimeActivityError`.
- #1533: `second_order_neural.md` exits 139 during Documenter example execution.
- #1534: unrelated `simplechains.md` docs example fails with Enzyme illegal type analysis.

Local verification on Julia 1.12.6:

```sh
timeout 3600 env JULIA_DEPOT_PATH="$PWD/.julia_depot" /home/crackauc/.juliaup/bin/julia +1.12.6 --project -e 'using Pkg; Pkg.activate("docs"); push!(LOAD_PATH, pwd()); using Documenter, SciMLSensitivity; ENV["GKSwstype"]="100"; using Plots; makedocs(; root="docs", sitename="target", modules=[SciMLSensitivity], clean=true, doctest=false, linkcheck=false, warnonly=[:missing_docs], pages=["target" => "examples/ode/second_order_adjoints.md"])'
```

Result: exited 139 after reverting, so this is not a docs-green claim. The revert removes the PR-introduced Enzyme runtime-activity failure, but the original/Mooncake page still appears to have a separate crash.

Also ran `git diff --check`, which passed.

### DIFF
--- a/docs/src/examples/ode/second_order_adjoints.md
+++ b/docs/src/examples/ode/second_order_adjoints.md
@@ -24,7 +24,7 @@ import OrdinaryDiffEq as ODE
 import Plots
 import Random
 import OptimizationOptimJL as OOJ
-import Enzyme
+import Mooncake
 
 u0 = Float32[2.0; 0.0]
 datasize = 30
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
We want to see what it takes to get it enzyme.

ChrisRackauckas-Claude:
Closing this revert PR per direction in #1490: the goal is to keep moving these docs examples to Enzyme rather than revert them. The isolated follow-up issues remain open for the Enzyme failures.

PR #1539, opened 2026-07-10
Add SundialsAdjoint using the SUNDIALS CVODES C adjoint interface

ChrisRackauckas-Claude:

## Summary

Adds a `SundialsAdjoint` sensitivity algorithm which uses the SUNDIALS CVODES C adjoint interface (`CVodeAdjInit`/`CVodeF`/`CVodeB`), wrapped by Sundials.jl since SciML/Sundials.jl#543 (Sundials v6.2.3).

- **`SundialsAdjoint{CS, AD, FDT, VJP}`** (exported, documented, added to the docs `@docs` list): options for `autojacvec`, the CVODES checkpoint interval `steps` (`Nd` of `CVodeAdjInit`), and the checkpoint interpolation `interp` (`:hermite`/`:polynomial`).
- **`SciMLSensitivitySundialsExt`** package extension (weakdep on Sundials ≥ 6.2.3) implements `_adjoint_sensitivities(sol, ::SundialsAdjoint, ::Union{CVODE_BDF, CVODE_Adams})`:
  - forward pass re-integrated by CVODES with checkpointing (`CVodeF`),
  - backward adjoint pass integrated by CVODES against its checkpoint interpolation (`CVodeB`), with stops at each discrete cost time to apply the `λ += dg/du` jump via `CVodeReInitB`,
  - parameter gradient accumulated by the CVODES backward quadrature (`CVodeQuadInitB`/`CVodeGetQuadB`).
  - The vjps inside the backward RHS/quadrature reuse the existing `vecjacobian!`/`vec_pjac!` machinery (via an `ODEQuadratureAdjointSensitivityFunction` and a `GaussIntegrand` cache), so `ReverseDiffVJP`, `EnzymeVJP`, `ZygoteVJP`, and the `autojacvec=false` Jacobian paths all work unchanged.
  - Supports the `:Dense`, `:Band`, and `:GMRES` CVODE linear solvers and `method = :Functional`.
- **`_concrete_solve_adjoint` wiring**: `SundialsAdjoint` is added to the discrete-adjoint union in `concrete_solve.jl`, so reverse-mode AD (Zygote etc.) of `solve(prob, CVODE_BDF(); sensealg = SundialsAdjoint())` automatically drives the C adjoint.
- **Informative errors**: using `SundialsAdjoint` with a non-Sundials solver (or without `using Sundials`) errors with instructions; continuous cost functionals, mass matrices, out-of-place functions, and non-`Float64` state error informatively.
- Minor version bump 7.114.0 → 7.115.0 (new public API).

## Local validation

All run on Julia 1.10 with the local Sundials.jl v6.2.3 clone dev'ed (v6.2.3 was registered today and had not yet propagated to pkg servers):

- `test/Core2/sundials_adjoint.jl`: 22/22 pass. Verifies `du0`/`dp` against `ForwardDiff.gradient` (rtol 1e-5) on Lotka-Volterra for `ReverseDiffVJP`, `EnzymeVJP`, `autojacvec=false`, default vjp, GMRES linear solver, `CVODE_Adams` + `:polynomial` interpolation, endpoint-only cost, and the Zygote `solve` reverse-AD path for both `p` and `u0`; plus the informative-error paths.
- `GROUP=Core2` run of the full group: my testset passes inside the group; 4 other files errored only with `Package X not found` in the ad-hoc dev environment (Reactant etc. missing), not from this change.
- Brusselator (BrussScaling benchmark setup, `abstol=1e-5, reltol=1e-7`): gradients agree with `GaussAdjoint(autojacvec=EnzymeVJP())` + `Rodas5` to ~1e-4 relative at N=4 and N=8 for all four CVODES configs (Dense/GMRES × Enzyme/compiled ReverseDiff).
- Performance (Brusselator scaling, times in seconds; companion SciMLBenchmarks PR adds this to BrussScaling.jmd):

| method | 256 params | 576 | 900 | 1156 |
|---|---|---|---|---|
| GaussAdjoint EnzymeVJP (Rodas5) | 0.276 | 0.954 | 2.04 | 3.32 |
| CVODES Dense EnzymeVJP | 0.052 | 0.249 | 0.90 | 1.66 |
| CVODES GMRES EnzymeVJP | 0.019 | 0.050 | 0.089 | 0.111 |

## Notes for review

- CI will fail to resolve until Sundials v6.2.3 propagates to the Julia pkg servers (registration merged today: JuliaRegistries/General#160579).
- While testing I found that `GaussAdjoint` currently does not compose with Sundials solvers at all: the checkpointing re-solve produces a solution with a different interpolation type than `GaussIntegrand`'s concretely-typed `sol` field accepts (`setproperty!` conversion error). That pre-existing issue is out of scope here; can file separately.
- Scope limitations (each errors informatively): discrete cost functionals only (which is what reverse-mode AD of `solve` uses), no callbacks/events, in-place `ODEProblem`s with `Float64` state, no mass matrices.

## Process notes

Implemented per the plan: struct + fallback error in the main package, full CVODES drive in the extension, `concrete_solve` wiring, tests, docs entry. Runic was run on all touched files.


ChrisRackauckas-Claude:
Addressed both review points in c292ad2a:

- **Standard autovjp**: the `autojacvec = nothing` fallback now calls the same `inplace_vjp` heuristic that `adjoint_sensitivities` uses, instead of hard-coding `ReverseDiffVJP()`. (This path only triggers on direct `_adjoint_sensitivities` calls — the top-level `adjoint_sensitivities` already applies `setvjp` before dispatching here.)

- **Continuous loss support**: `dgdu_continuous`/`dgdp_continuous` and scalar `g(u, p, t)` are now supported, including mixed discrete + continuous. The CVODES slots map directly: the `-dg/du` term is added to the backward RHS (`fB = -J'λ - dg/du`) and the `-dg/dp` term to the backward quadrature (`fQB = -(λ' df/dp + dg/dp)`), both through the existing `accumulate_cost!` machinery so the `g`-via-AD form comes for free. The one unsupported form is discrete scalar `g` with `t` (errors, pointing at `dgdu_discrete`).

New tests (all passing locally, 28 total now): continuous cost via `dgdu_continuous + dgdp_continuous`, via scalar `g`, and mixed discrete + continuous, each verified against a ForwardDiff reference computed with an augmented-quadrature-state ODE at rtol 1e-5.

Also re-ran the Brusselator check: all configs still agree with `GaussAdjoint` to ~1e-4.

One correction to the PR description: `GaussAdjoint` *does* work with Sundials solvers when the forward solve is dense (`dense = true`); only the checkpointing path taken for `saveat` solutions hits the `GaussIntegrand.sol` interpolation-type mismatch. The updated SciMLBenchmarks PR uses this for a same-solver comparison (CVODE_BDF dense/GMRES with both `SundialsAdjoint` and `GaussAdjoint`, plus FBDF): with the solver held fixed the two adjoints are within ~20% of each other (GaussAdjoint slightly faster at the largest sizes), confirming the big gaps in the first table were dominated by the forward solver choice (CVODE_BDF+GMRES vs Rodas5+dense LU), not the C-vs-Julia adjoint machinery.

Project.toml
```diff
@@ -126,6 +128,7 @@ StaticArraysCore = "1.4.4"
 Statistics = "1.10"
 SteadyStateDiffEq = "2.10.1"
 StochasticDiffEq = "7"
+Sundials = "6.2.3"
```
ChrisRackauckas:
```suggestion
Sundials = "6.3.0"
```

ChrisRackauckas-Claude:
Note on the red master CI for the merge commit (1ebcdea): all 41 failures are `Unsatisfiable requirements detected for package Sundials` — the pkg servers were still serving a registry snapshot without Sundials v6.2.3 when CI ran (15:36Z). The General registry itself has had 6.2.3 since 01:25Z (and 6.3.0 has since been tagged as well), so this resolves itself with propagation; the jobs just need a re-run (I don't have workflow re-run rights on this repo). No code changes needed.

PR #1542, opened 2026-07-10
Add DAEProblem support to SundialsAdjoint via the SUNDIALS IDAS C adjoint interface

ChrisRackauckas-Claude:

Follow-up to #1539: extends `SundialsAdjoint` to fully implicit `DAEProblem`s using the SUNDIALS **IDAS** C adjoint interface (`IDAAdjInit`/`IDASolveF`/`IDACreateB`/`IDAInitB`/`IDACalcICB`/`IDAQuadInitB`/`IDASolveB`), mirroring the CVODES structure merged there. Uses the IDAS adjoint wrappers already present in the registered Sundials.jl v6.2.3 (validated end-to-end by SciML/Sundials.jl#546), so no Sundials.jl release is needed.

## Design

New method `_adjoint_sensitivities(sol, ::SundialsAdjoint, ::Sundials.IDA)` in `ext/SciMLSensitivitySundialsExt.jl` for in-place Float64 `DAEProblem`s `F(du, u, p, t) = 0` with discrete cost functionals.

**Adjoint formulation** (Cao–Li–Petzold / IDAS user guide), backward problem in residual form:

```
resB = (∂F/∂du)ᵀ ypB − (∂F/∂u)ᵀ yB
```

with backward quadrature `qB' = (∂F/∂p)ᵀ λ`, `qB(tf) = 0`, so `dG/dp = qB(t0) = −∫ λᵀ (∂F/∂p) dt`, and `du0 = (∂F/∂du)ᵀ λ |_{t0}` (the gradient w.r.t. the differential components of `u0`; algebraic entries are zero for semi-explicit systems). The sign conventions were pinned against the analytic gradients of the low-level Sundials.jl IDAS adjoint test (`test/idas_adjoint.jl` from SciML/Sundials.jl#546) and then validated numerically against ForwardDiff (below).

**Key assumption (documented prominently in the docstring):** the residual must be **linear in `du` with a constant `∂F/∂du`** (i.e. `F = M du − h(u, p, t)` with a constant mass matrix), so the `d/dt (∂F/∂du)ᵀ λ` term of the adjoint DAE vanishes. This covers mass-matrix-style and semi-explicit index-1 DAEs written implicitly — the overwhelmingly common case. A best-effort runtime screen compares the `(∂F/∂du)ᵀ v` vjp at two different `(du, u, t)` points and errors when a state/time-dependent `∂F/∂du` is detected, so du-nonlinear residuals fail loudly instead of returning silently wrong gradients.

**Cost jumps and consistent backward initialization:** `(∂F/∂du)ᵀ` is formed once as a constant matrix `Mt` (n vjps, valid by the constancy assumption). Discrete cost jumps are transferred onto the adjoint by solving `(∂F/∂du)ᵀ Δλ = gu` via `pinv(Mt)`, with a residual check that errors informatively when the cost gradient has components w.r.t. algebraic variables (that case needs an index-1 constraint-transfer term not implemented here — erroring beats a silently wrong gradient; `gu` on algebraic components is not in the range of `(∂F/∂du)ᵀ`). After `IDAInitB` and after each interior-stop `IDAReInitB`, `IDACalcICB` (enabled by `IDASetIdB` from `differential_vars`) computes the consistent algebraic λ components and differential λ′ components; the forward `(y, y′)` that `IDACalcICB` needs at interior stops comes from `IDAGetAdjY`. `differential_vars` is therefore required (informative error otherwise). Empirically `IDACalcICB` converged immediately on all test problems, including with the zero `λ′` guess.

**VJPs:** the existing `vecjacobian!`/`vec_pjac!` caches are ODE-`f(du,u,p,t)`-shaped, so the extension builds ReverseDiff `GradientTape`s over `(du, u, p, t)` directly: one tape forward pass per callback evaluation with two reverse passes (seed `yB` → `(∂F/∂u)ᵀ` and `(∂F/∂p)ᵀ` pulls; seed `ypB` → `(∂F/∂du)ᵀ` pull). `ReverseDiffVJP(true)` caches a compiled tape; the uncompiled default re-records at each evaluation point (same convention as `vecjacobian!`). Only `ReverseDiffVJP` is supported for DAEs in this first version; other `autojacvec` choices error informatively. `inplace_vjp` now returns `ReverseDiffVJP()` early for `AbstractDAEProblem` (its probes are 4-arg-ODE-shaped and previously all failed through to `false` with noisy warnings).

**Reverse-AD of `solve`:** the ODE/SDE/RODE `_concrete_solve_adjoint` method in `src/concrete_solve.jl` is widened to `AbstractDAEProblem`, guarded so that non-`SundialsAdjoint` continuous adjoints on DAEs get an informative error (previously a `MethodError`). `SciMLBase.sensitivity_solution` has no `DAESolution` method, so a local `_sensitivity_solution` helper builds the subset solution via `SciMLBase.build_solution` for DAE solutions and delegates unchanged for everything else. Zygote-over-`solve` works for both `p` and `u0`.

## Supported / deferred

Supported (DAE path): discrete costs (`t` + `dgdu_discrete`/`dgdp_discrete`), `no_start`, jump loop with `IDAReInitB` + `IDAQuadReInitB` + `IDACalcICB`, `:Dense`/`:Band`/`:GMRES` linear solvers, `:hermite`/`:polynomial` checkpoint interpolation, `steps`, `quad_error_control`, `abstol`/`reltol`/`maxiters`, reverse-mode AD of `solve` (Zygote tested).

Deferred with informative errors (possible follow-ups):

- continuous cost functionals (`g`/`dgdu_continuous`/`dgdp_continuous`) for DAEs;
- cost gradients w.r.t. algebraic variables (needs the index-1 constraint-transfer term);
- vjp choices other than `ReverseDiffVJP` (an Enzyme path would need a DAE-residual vjp cache);
- du-nonlinear residuals (needs the `d/dt (∂F/∂du)ᵀ λ` term).

## Validation (all run locally, Julia 1.10)

Analytic index-1 DAE (`F1 = y1' + a y1`, `F2 = y2 − y1`, `G = y1(tf)`): `du0 = exp(-a tf)` and `dp = -tf exp(-a tf)` matched to ~1e-9 absolute; the algebraic `du0` component is exactly 0.

Robertson DAE (index-1, stiff, `ts = [10, 50, 100]`, `G = Σ_ts (u1+u2)`) vs `ForwardDiff.gradient` through the equivalent mass-matrix ODE formulation (`Rodas5P`, 1e-12 tolerances), forward solve `solve(daeprob, IDA(); saveat = ts, abstol = 1e-10, reltol = 1e-10)`:

```
du0 rel err:       [5.9e-9, 1.8e-5]   (2nd component magnitude ~2.8e-5)
dp  rel err:       [1.8e-9, 5.2e-9, 5.2e-9]
Zygote dp rel err: [1.8e-9, 5.2e-9, 5.2e-9]
```

Local test-suite runs:

```
Sundials IDAS DAE Adjoint | 32 pass   (new test/Core2/sundials_idas_adjoint.jl)
Sundials CVODES Adjoint   | 28 pass   (existing suite re-run)
Literal Adjoint           |  1 pass   (ODE-path regression check for the
                                       concrete_solve.jl changes)
```

## Notes for review

- One existing-test change: the `occursin("CVODES-compatible solver", …)` assertion in `test/Core2/sundials_adjoint.jl` was updated to `occursin("CVODES/IDAS", …)` because the `SundialsAdjoint` fallback error message now mentions both interfaces.
- Version bumped 7.115.0 → 7.116.0 (new public capability, minor bump).
- `SundialsAdjoint` docstring updated (overview + SciMLProblem Support) with DAE support, the linear-in-`du` assumption, `du0` semantics, and the Cao–Li–Petzold reference; the docs `@docs` entry already exists.
- A `sensitivity_solution(::DAESolution, u, t)` method in SciMLBase would let us drop the local `_sensitivity_solution` helper; can be done upstream as a follow-up.
- While re-running the existing CVODES suite locally, one run (out of five) segfaulted in `N_VScale` inside `libsundials_core` during the `adjoint_sensitivities interface` testset — a code path this PR does not touch. This looks like an intermittent GC-rooting issue in the #1539 CVODES extension code (prime suspect: SUNDIALS solver objects/handles held only in dead locals outside the `GC.@preserve` block); it is being investigated separately and any fix will come as its own PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLSensitivity"
 uuid = "1ed8b502-d754-442c-8d5d-10ac956f44a1"
-version = "7.115.0"
+version = "7.116.0"
 authors = ["Christopher Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
 
 [deps]
--- a/ext/SciMLSensitivitySundialsExt.jl
+++ b/ext/SciMLSensitivitySundialsExt.jl
@@ -885,6 +885,16 @@ function _adjoint_sensitivities(
     ncheck = Ref{Cint}(0)
     which = Ref{Cint}(0)
 
+    # The forward/backward linear solvers and matrices are raw SUNDIALS pointers
+    # that IDAS uses for the whole integration; they carry no finalizer, so they
+    # must stay bound (pre-declared here, assigned inside the `try`) and be freed
+    # explicitly in the `finally` after `IDAFree` (via `empty!(mem)`) and before
+    # `SUNContext_Free`, otherwise they leak. `:GMRES` returns a `nothing` matrix.
+    local LSf = nothing
+    local Af = nothing
+    local LSB = nothing
+    local AB = nothing
+
     GC.@preserve data ufwd_nv dufwd_nv yret_nv ypret_nv yinterp_nv ypinterp_nv λ_nv λp_nv qB_nv id_nv begin
         try
             # Forward pass with checkpointing. The provided `(u0, du0)` are
@@ -1099,6 +1109,10 @@ function _adjoint_sensitivities(
             has_p && (dp .+= qB)
         finally
             empty!(mem)
+            LSf === nothing || Sundials.SUNLinSolFree(LSf)
+            LSB === nothing || Sundials.SUNLinSolFree(LSB)
+            Af === nothing || Sundials.SUNMatDestroy(Af)
+            AB === nothing || Sundials.SUNMatDestroy(AB)
             Sundials.SUNContext_Free(ctx)
         end
     end
--- a/ext/SciMLSensitivitySundialsExt.jl
+++ b/ext/SciMLSensitivitySundialsExt.jl
@@ -4,12 +4,12 @@ using SciMLSensitivity: SciMLSensitivity, SciMLBase, SciMLLogging,
     SundialsAdjoint, QuadratureAdjoint, GaussAdjoint,
     ODEQuadratureAdjointSensitivityFunction, GaussIntegrand,
     vecjacobian!, vec_pjac!, accumulate_cost!, inplace_vjp,
-    ReverseDiffVJP, mutable_zeros,
+    ReverseDiffVJP, mutable_zeros, compile_tape, ReverseDiff,
     canonicalize, Tunable, isscimlstructure, unwrapped_f
 import SciMLSensitivity: _adjoint_sensitivities
-using SciMLBase: ODEProblem, isinplace
+using SciMLBase: ODEProblem, DAEProblem, isinplace
 using Sundials: Sundials, N_Vector, NVector, realtype
-using LinearAlgebra: I, UniformScaling
+using LinearAlgebra: I, UniformScaling, pinv, norm, mul!
 
 _cvodes_lmm(::Sundials.CVODE_BDF) = Sundials.CV_BDF
 _cvodes_lmm(::Sundials.CVODE_Adams) = Sundials.CV_ADAMS
@@ -189,7 +189,10 @@ function _adjoint_sensitivities(
 
     prob = sol.prob
     prob isa ODEProblem ||
-        error("SundialsAdjoint only supports `ODEProblem`s.")
+        error(
+        "SundialsAdjoint with `CVODE_BDF`/`CVODE_Adams` only supports `ODEProblem`s. " *
+            "For a `DAEProblem`, use `IDA()` as the solver instead."
+    )
     isinplace(prob) ||
         error("SundialsAdjoint currently only supports in-place (mutating) `ODEProblem`s.")
     mm = prob.f.mass_matrix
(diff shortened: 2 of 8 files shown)


ChrisRackauckas-Claude:
Follow-up push (`b163aeb1`): CI-timing diagnosis + a resource-lifetime fix.

**Why the earlier CI runs were all red.** Every test job failed at the `Pkg` step with `Unsatisfiable requirements ... Sundials [c3572dad] ... possible versions are: 2.0.0 - 6.2.2 ... restricted to versions 6.3.0 - 6`. This is not a code problem: the `Sundials = "6.3.0"` compat (inherited from the merged CVODES PR #1539) requires Sundials.jl v6.3.0, which registered in General only at 2026-07-10T15:05Z (JuliaRegistries/General#160709). The CI run at 16:34Z still hit a stale Julia PkgServer registry snapshot that topped out at 6.2.2. v6.3.0 is now registered and merged, so this re-triggered run should resolve. The `6.3.0` pin is correct and matches master; no compat change was made.

**Resource-lifetime fix.** The IDAS path allocated the forward and backward SUNDIALS linear solvers and dense/band matrices but never freed them, leaking one set of raw SUNDIALS objects per adjoint call. They are now pre-declared before the `try` and freed in the `finally` (after `IDAFree` via `empty!(mem)`, before `SUNContext_Free`), mirroring the CVODES resource-lifetime pattern in #1543. `:GMRES` returns a `nothing` matrix and is guarded.

**Local validation (Julia 1.10, Sundials.jl v6.3.0):** all pass after the fix.

```
analytic index-1 DAE          | 4  pass
Robertson DAE vs ForwardDiff  | 12 pass
reverse-mode AD of solve      | 3  pass
informative errors            | 13 pass
--------------------------------------
new test/Core2/sundials_idas_adjoint.jl | 32 pass

test/Core2/sundials_adjoint.jl (CVODES regression) | 28 pass
```

PR #1545, opened 2026-07-10
Add adjoint sensitivity support for fully implicit DAEProblems (index-1 and Hessenberg index-2)

ChrisRackauckas-Claude:
> [!IMPORTANT]
> **This PR should be ignored until reviewed by @ChrisRackauckas.**

## Summary

Adds adjoint sensitivity analysis for fully implicit DAEs (`DAEProblem`, `F(du, u, p, t) = 0`) with `InterpolatingAdjoint`, including structural hooks for index-1 and Hessenberg index-2 systems. Previously `adjoint_sensitivities` errored with "Continuous adjoint sensitivities are only supported for ODE/SDE/RODE problems" and `solve`-level reverse AD routed `DAEProblem` to whole-solver AD fallbacks.

Both the direct interface and the `solve`-level AD interface now work:

```julia
prob = DAEProblem(rober_dae!, du0, u0, tspan, p, differential_vars = [true, true, false])
sol  = solve(prob, DFBDF(), abstol = 1e-10, reltol = 1e-10)

# direct
du0grad, dpgrad = adjoint_sensitivities(sol, FBDF(); t = ts, dgdu_discrete = dg,
    sensealg = InterpolatingAdjoint(autojacvec = ReverseDiffVJP()))

# via solve + Zygote
Zygote.gradient(p -> loss(solve(prob, DFBDF(); p, saveat = ts,
    sensealg = InterpolatingAdjoint(autojacvec = ReverseDiffVJP()))), p)
```

## Mathematical formulation

Follows Cao, Li, Petzold & Serban, *Adjoint sensitivity analysis for differential-algebraic equations: The adjoint DAE system and its numerical solution*, SIAM J. Sci. Comput. 24(3), 2003 (already cited in the `BacksolveAdjoint` docstring references). The adjoint of `F(u̇, u, p, t) = 0` is

```
d/dt(F_du' λ) − F_u' λ + g_u' = 0,   dG/du₀ = (F_du' λ)(t₀),   dG/dp = ∫ (g_p' − F_p' λ) dt
```

To avoid the total time derivative of `F_du` along the trajectory, the augmented variable `w := F_du' λ` is introduced. The naive augmented system `ẇ = F_u'λ − g_u'`, `0 = w − F_du'λ` is index-2 even for index-1 DAEs, so it is **structurally index-reduced** using `differential_vars`: for algebraic variables `j` (zero columns of `F_du`), `w_j ≡ 0` and the differentiated constraint `0 = [F_u'λ − g_u']_j` is used as the algebraic equation determining the remaining adjoint components. With the parameter quadrature appended, the augmented adjoint state is `z = [w; λ; grad]`:

| rows | equation |
|---|---|
| `j` differential var | `ẇ_j = [F_u'λ]_j − g_u[j]` |
| `j` algebraic var | `0 = w_j` |
| `n+j`, `j` differential | `0 = w_j − [F_du'λ]_j` |
| `n+j`, `j` algebraic | `0 = [F_u'λ]_j − g_u[j]` |
| grad | `grad' = F_p'λ − g_p'` |

For `F = M u̇ − f(u, p, t)` this reduces exactly to the transposed-mass-matrix adjoint that `ODEAdjointProblem` already uses for semi-explicit DAEs. All vector-Jacobian products of the residual are computed in a single reverse pass (`dae_vecjacobian!`), with `ReverseDiffVJP`, `ZygoteVJP` (out-of-place), and `autojacvec = false` (ForwardDiff Jacobians) supported.

## Index-1 / index-2 handling

The DAE structure is classified once at the terminal time (`dae_adjoint_structure`): algebraic equations = zero rows of `∂F/∂(du)`, algebraic variables = `differential_vars` (or zero columns as fallback), and the algebraic block `dhda = F_u[algeqs, algvars]` decides index-1 (`dhda` nonsingular) vs Hessenberg index-2 (`dhda ≡ 0` with `F_u[algeqs, diffvars] · F_du[diffeqs, diffvars]⁻¹ · F_u[diffeqs, algvars]` nonsingular). Higher-index / mixed systems get an informative error pointing at index reduction.

Discrete-cost jumps in `w` must respect the adjoint (hidden) constraints. Both indices share one correction path (a multiplier `Δλa` supported on the algebraic equations, mirroring the existing semi-explicit `ReverseLossCallback` machinery):

- **index-1:** `Δλa = −dhda' \ gᵤ[algvars]` — allows discrete/continuous cost on algebraic variables;
- **index-2:** `Δλa = −(N·dhdd') \ (N·gᵤ[diffvars])` with `N = (F_du[diffeqs,diffvars] \ F_u[diffeqs,algvars])'`, which projects the jump onto the adjoint constraint manifold while preserving the pairing with admissible (constraint-tangent) state perturbations. Discrete cost on index-2 *algebraic* variables is rejected (it corresponds to derivatives of delta distributions).

In both cases `w[diffvars] += gᵤ[diffvars] + dhdd'·Δλa` and the parameter-gradient correction `dp += Δλa' · ∂F_algeqs/∂p` is accumulated after the reverse solve via the same `Δλas` mechanism the mass-matrix code uses — this is exercised in the tests by a parameter-dependent algebraic constraint (index-1) and a parameter-dependent index-2 constraint.

Adjoint consistency at the terminal time and after each jump is restored by `BrownFullBasicInit` (whose initialization system *is* the Cao–Petzold consistent-terminal-condition solve). For index-2 that system is singular, so `NoInit` is used and the BDF discretization determines the index-2 multiplier from the second step on (its terminal value is never used by the gradient extraction). The `initializealg` is attached to the adjoint problem's kwargs so callers can override it.

## Reverse solver flexibility

`DAEAdjointProblem` emits the augmented adjoint in two forms depending on the adjoint algorithm:

- **mass-matrix `ODEProblem`** for ODE algorithms (`FBDF`, `Rodas5P`, ...);
- **fully implicit `DAEProblem`** (residual wrapper `DAEAdjointResidual`, `M ż − G(z)`) for DAE algorithms (`DFBDF`, ...), which is what makes the `solve`-level path work — `concrete_solve` reuses the forward algorithm for the reverse solve. Backwards-in-time DAE solves need an explicit initial `dt` (automatic init-dt fails for reverse DAE tspans), seeded from the forward solution's last step.

## Requirements and current limitations

- Forward solution must be dense with an interpolant supporting `sol(t, Val{1})` (native Julia DAE solvers like `DFBDF`; fixed-step `DImplicitEuler` for index-2, since adaptive BDF error control is unstable on index-2 systems).
- `InterpolatingAdjoint` only; `QuadratureAdjoint`/`GaussAdjoint`/`BacksolveAdjoint` throw an informative error suggesting `InterpolatingAdjoint` or the mass-matrix form. Checkpointing, callbacks, `EnzymeVJP`/`MooncakeVJP`/`ReactantVJP`, and Functors.jl parameters are not yet supported (informative errors).
- `du0` (the `DAEProblem` derivative initial condition) is treated as non-differentiable in the `solve`-level path; `u0` gradients are returned for the differential components (algebraic components are constrained, gradient 0), consistent with the semi-explicit mass-matrix convention.
- Default `sensealg` selection for `DAEProblem` is unchanged (discrete-AD fallbacks); the continuous adjoint is opt-in via an explicit `sensealg`.

## Follow-ups (separate PRs)

- SciMLBase: add a `sensitivity_solution(::AbstractDAESolution, u, t)` method so the local `_sensitivity_solution` shim in `concrete_solve.jl` can be dropped.
- `QuadratureAdjoint`/`GaussAdjoint` DAE variants, checkpointing, `EnzymeVJP`, callback (event) support, and making `InterpolatingAdjoint` the default for `DAEProblem` in `automatic_sensealg_choice`.

## Tests

`test/Core3/dae_adjoint.jl` (registered in the Core3 group), all compared against `ForwardDiff` through the DAE solve or analytic gradients:

- **Index-1** fully implicit Robertson: discrete cost including the algebraic variable, `du0` gradient in consistent directions, adjoint solved both as mass-matrix (`FBDF`) and fully implicit (`DFBDF`), default vjp choice, Zygote `solve`-level `p` and `u0` gradients — agreement to `1e-4` rtol (observed ~1e-8 relative).
- **Index-1 with p-dependent algebraic constraint** (exercises the `Δλas` dp correction).
- **Index-0** (implicit-form Lotka–Volterra) vs the ODE adjoint: `ReverseDiffVJP()`, `ReverseDiffVJP(true)`, `autojacvec=false`, out-of-place + `ZygoteVJP`, discrete and continuous costs (`dgdu_continuous` and AD of `g`).
- **Hessenberg index-2** with p-dependent constraint vs analytic gradients (first-order convergence in the forward `dt` verified during development).
- Informative-error tests for `QuadratureAdjoint` and checkpointing.

Local run: 28/28 pass. Full `GROUP=Core3` run also passing locally (see PR comments for logs).

## Process notes

Implementation notes: the augmented formulation, row layout, jump-correction derivations, and initialization requirements are documented in a long header comment in `src/dae_adjoint.jl`. Development was validated incrementally against ForwardDiff on Robertson (index-1), implicit Lotka–Volterra (index-0), and an analytic Hessenberg index-2 problem before consolidation into the test suite; `DFBDF` was confirmed unstable on index-2 forward solves (both the pendulum and a linear Hessenberg-2 system), which is why the index-2 test uses fixed-step `DImplicitEuler`.


ChrisRackauckas-Claude:
## Local test runs (documentation of verification process)

**New test file** (`test/Core3/dae_adjoint.jl`), run standalone against the final tree (Julia 1.12.4):

```
Test Summary:                     | Pass  Total     Time
Index-1: fully implicit Robertson |    9      9  2m18.3s
Index-1: parameter-dependent algebraic constraint |    1      1  26.2s
Index-0: implicit-form ODE vs ODE adjoint |   10     10  2m30.5s
Index-2 Hessenberg with p-dependent constraint |    6      6  16.3s
informative errors |    2      2  6.4s
```

**Full Core3 group** (all seven Core3 files run under `@safetestset` against this branch):

```
Test Summary: | Pass  Broken  Total      Time
Core 3        |  270       6    276  50m57.5s
```

(The 6 broken are pre-existing `@test_broken` in `Core3/adjoint.jl`, unchanged.)

Note on local `Pkg.test()`: at the time of development, `GROUP=Core3 Pkg.test()` could not resolve locally because master's new compat bound `Sundials = "6.3.0"` (from #1539) was registered in General only ~1h earlier (JuliaRegistries/General#160709, merged 2026-07-10 14:49 UTC) and the Pkg-server registry mirror had not yet propagated. That is registration lag, not a breakage — CI with a fresh registry should resolve fine. The Core3 run above was therefore executed by replicating the group's `@safetestset` includes in an environment with the Core3 test dependencies (Sundials is not used by any Core3 file).

Development validation before consolidating the tests: index-1 Robertson dp/du0 matched ForwardDiff-through-`DFBDF` to ~8 significant digits; index-2 gradients matched analytic values with first-order convergence in the forward `dt` (errors 1.8e-3 → 3.7e-4 as dt went 1e-4 → 2e-5), confirming the residual discrepancy is forward-discretization error, not adjoint error. `DFBDF` was confirmed unstable (retcode `Unstable`) on index-2 forward solves for both the index-2 pendulum and the linear Hessenberg-2 test system, which is why the index-2 test uses fixed-step `DImplicitEuler`.

ChrisRackauckas-Claude:
CI status note: all test-group jobs in the first run failed at dependency resolution with `Unsatisfiable requirements ... Sundials` — master's new compat bound `Sundials = "6.3.0"` (#1539) is registered in General (JuliaRegistries/General#160709, merged 2026-07-10 15:05 UTC) but the Julia pkgservers had not propagated it as of ~19:20 UTC. This affects master CI identically and is unrelated to this PR; jobs should go green on a re-run once the pkgservers catch up. The typos-check failure (journal abbreviation "Comput.") is fixed in e7957826.

ChrisRackauckas-Claude:
CI round 3 triage: after the registry propagated, 46/50 checks passed. The two remaining failures were (a) QA: ExplicitImports flagging the two non-public SciMLBase names used by the `_sensitivity_solution` shim — added to the existing ignore-until-public allowlist in `test/QA/aqua.jl` (alongside `sensitivity_solution` itself), verified locally with `GROUP=QA Pkg.test()`: 19/19 pass; and (b) Downgrade (lts, Core1), which is failing on master and on all recent branches (pre-existing, a `downgrade-sundials-no-promote` branch is already addressing it) — though this PR still tightens the new `OrdinaryDiffEqBDF` test-dep compat to "2.3" since 2.0.0 does not precompile against current dependency minima.

ChrisRackauckas-Claude:
CI round 4 triage: QA now passes. New failures appeared in Core1/Core2/SDE1/SDE2 (Julia 1.11) and Core3 (Julia 1) that were green in round 3 with identical source; the failing round resolved SciMLBase v3.34.0, which existed for only a few hours before v3.35.0. The Core3 failure (maxiters interruption inside the reverse DAE solve of the Zygote-path tests) does not reproduce locally on the updated stack (SciMLBase v3.35.0, OrdinaryDiffEq v7.1.2): gradients match FD to 9 digits. Retriggered CI to resolve the current releases. Downgrade remains the known master-wide failure.

ChrisRackauckas-Claude:
## Architecture unification + #1542 incorporation (commits bf07912d, ad15e73f)

Per review direction, the DAE support is now integrated with the existing package machinery the way SDE/RODE support is, instead of a parallel code path, and #1542 (IDAS `SundialsAdjoint`) is merged into this PR and rebased onto the shared dispatches:

- **`vecjacobian!` is DAE-robust across backends**: optional `du`/`ddu` arguments thread the residual form through `ReverseDiffVJP`, **`EnzymeVJP`** (new for DAEs — one `autodiff` call with a `du` shadow), `ZygoteVJP`, `TrackerVJP`, and `autojacvec = false`, with an informative-error fallback so the Mooncake/Reactant extensions degrade gracefully.
- **`AdjointDiffCache`/`adjointdiffcache`** carry the DAE structure (`algeeq_idxs`/`diffeq_idxs`/`isdae`/`index2`) and build the 4-input residual tapes/Enzyme configs; **one `ReverseLossCallback`** now applies the index-corrected cost jumps for mass-matrix and fully implicit forms.
- **All quadrature-capable adjoints extended to index-1 DAEs**: `QuadratureAdjoint`, `GaussAdjoint`, `GaussKronrodAdjoint` (adjoint state `[w; λ]`, parameter gradient from the residual p-vjp via a shared `dae_vec_pjac!` in both integrands, `Δλa` jump corrections included).
- **Index-2 stays `InterpolatingAdjoint`-only, by measurement**: the index-2 adjoint multiplier is impulsive at discrete cost times and its dp contribution concentrates at quadrature-interval endpoints. Probing the analytic Hessenberg-2 problem: quadrature-method error is tolerance-independent (1.09e-2 at both 1e-8 and 1e-10) while the state-augmented accumulation converges (2.2e-5, forward-dt-limited). The quadrature methods reject index-2 with this explanation.
- **IDAS route upgraded via the shared dispatches**: tapes from `get_dae_paramjac_config`, `(∂F/∂du)ᵀ` from `dae_du_jacobian`, and the cost-jump transfer upgraded from range-check-and-error to the shared index-1 correction — the "no cost gradients on algebraic variables" and (implicit) "no p-dependent constraints" limitations of #1542 are now lifted and tested. Hessenberg index-2 is rejected on that route (`IDACalcICB` cannot initialize it).

Also answered in review discussion: reinit (Brown/`IDACalcICB`) restores *consistency* of the algebraic adjoint components after a jump, but cannot recover the index-1 constraint-transfer term `dhdd'Δλa` in the differential jump nor the `Δλa'∂F_alg/∂p` dp term — the correction is a no-op exactly when the discrete cost doesn't touch algebraic variables.

### Local verification (final tree, Julia 1.12.4)

```
native DAE suite (test/Core3/dae_adjoint.jl):        53/53
IDAS suite (test/Core2/sundials_idas_adjoint.jl):    36/36
GROUP=Core3 Pkg.test():   295 pass / 6 pre-existing broken   (79m51s)
GROUP=Core2 Pkg.test():   193 pass / 13 pre-existing broken  (68m59s)
GROUP=QA    Pkg.test():   19/19                              (3m03s)
```

The native suite covers all four methods × both adjoint problem forms (mass-matrix `FBDF` / fully implicit `DFBDF`), EnzymeVJP, per-method `Δλa`-dp-correction coverage on a p-dependent constraint, index-0/1/2, and Zygote solve-level gradients.

ChrisRackauckas-Claude:
Thanks — addressed in e1b00ffb. Point by point:

**1. `VERSION >= v"1.8-"` check** — right, always true now (min supported is 1.10). This was pre-existing code I relocated during the refactor; removed the branch, kept the `@view`.

**2. `DAEAdjointSensitivityFunction` (R170-246) — dead code?** — it's live, not dead: it's the *shared* augmented-adjoint RHS used by all the continuous DAE methods. `_dae_adjoint_problem` builds one for `InterpolatingAdjoint` (state `[w; λ; grad]`, `withgrad=true`), `QuadratureAdjoint`, and `GaussAdjoint`/`GaussKronrodAdjoint` (state `[w; λ]`, `withgrad=false`), and `DAEAdjointResidual` wraps it for the fully-implicit-`DAEProblem` reverse-solve form. It can't reuse `ODEInterpolatingAdjointSensitivityFunction` etc. because the augmented DAE state has a genuinely different layout and RHS (the `w`/`λ` split with the algebraic-row rearrangement `0 = w_j` / `0 = w_j − [F_du'λ]_j`), which the ODE adjoint RHS `dλ = −F_u'λ − g_u` doesn't have. If you'd prefer it live physically alongside the other sensitivity functions rather than in `dae_adjoint.jl`, I'm happy to move it.

**3. index-2 init = ShampineCollocationInit (BDF1)?** — good thought, and I tried it, but it's *not* equivalent here and gives the wrong gradient. `ShampineCollocationInit`'s collocation step modifies **all** variables (differential + algebraic) to satisfy the DAE. In the augmented adjoint, the differential components `w`/`grad` at the terminal time and after each jump are **prescribed boundary conditions** (`w(T)=0`, and the jump sets them), not values to be made "consistent" — so Shampine perturbs the boundary data and returns garbage (measured `dp[2] = −0.223` vs the correct `0.060` on the analytic index-2 problem). `NoInit` is correct here precisely because `z(T)=0` is already consistent for the homogeneous augmented system, and after each jump `derivative_discontinuity!` lets the BDF discretization pick up the index-2 multiplier on the following step. I put this reasoning in a comment.

**4. Doubled `_vecjacobian!` dispatch missing MooncakeVJP** — fixed. The separate 13-arg DAE `_vecjacobian!` methods are gone; `du`/`ddu` are now optional trailing args on the existing per-backend `_vecjacobian!` methods, each branching internally (`du === nothing || return _dae_vecjacobian!(...)`) — so DAE support lives exactly where each backend already does. `MooncakeVJP`/`ReactantVJP` now carry the same optional args and route to the generic `_dae_vecjacobian!` fallback, which errors in-method (`"MooncakeVJP is not currently supported for DAEProblem adjoints…"`) rather than being silently absent from a parallel dispatch. Implementing Mooncake/Reactant DAE residual vjps would be a follow-up in those extensions (they'd need `(du, u, p, t)` pullback caches).

Re-ran the native DAE suite green after the fold; the full `GROUP=Core3`/`Core2`/`QA` runs are in progress and I'll post the numbers.

ChrisRackauckas-Claude:
Full local `Pkg.test` regression on the final tree (fold + trim, Julia 1.12.4), all green:

```
GROUP=Core3:  295 pass /  6 pre-existing broken  (77m)
GROUP=Core2:  199 pass / 13 pre-existing broken  (71m)   ← includes both Sundials suites
GROUP=QA:      19 pass                            (3m)
```

CI note: the red 1.11 jobs are self-hosted-runner communication drops (and one earlier 2h timeout), not test failures — Core3-on-LTS runs the identical tests green in ~50m on the same CI runs, and the same 1.11 runners are dropping across unrelated groups (SDE2/SDE3/Callbacks2/Core1/Core7). Documentation is a docs-tutorial segfault that's red on master too, and Downgrade is the known master-wide failure. I trimmed the redundant adjoint-form coverage in the heaviest DAE testset (both forms once via `InterpolatingAdjoint`, mass-matrix form for the other three methods — per-method and per-form coverage preserved) to give the 1.11 runners headroom regardless.

PR #1556, opened 2026-07-17
Provide exact adjoint Jacobian for GaussAdjoint/QuadratureAdjoint implicit adjoint solves (~2-3.5x on stiff Brusselator adjoints)

ChrisRackauckas-Claude:
> [!IMPORTANT]
> This PR should be ignored until reviewed by @ChrisRackauckas.

## Summary

Investigating GaussAdjoint performance on the 2D Brusselator (N=16, 512 states) showed that for **stiff adjoint solves without a user-provided analytical Jacobian**, the implicit adjoint solver differentiates the vjp-based adjoint rhs itself to get its Jacobian. That costs a full vjp (plus a forward-solution interpolation) per Jacobian column — and with `ReverseDiffVJP` it re-records the tape with `Dual` numbers on **every** Jacobian evaluation.

But the adjoint rhs of the state-only adjoints (`GaussAdjoint`, `QuadratureAdjoint`) is linear in `λ`: its Jacobian is exactly `-(df/du)^T` evaluated at the interpolated forward solution, independent of `λ`. #1378 already passed the user's analytical `jac` through to the adjoint solver; this PR extends that to the no-user-jac case by differentiating the **forward** rhs directly (ForwardDiff/FiniteDiff per the sensealg's `autodiff` setting) and negate-transposing.

The shared closure construction now lives in `build_adjoint_jac` (adjoint_common.jl), used by both `GaussAdjoint` and `QuadratureAdjoint`.

**Gating for the AD fallback** (conservative):
- no user `jac`, no `jac_prototype` (dense only; a sparse prototype without user jac keeps the existing colored-AD-of-adjoint-rhs path)
- `SciMLBase.forwarddiffs_model(alg)` — the adjoint solver would push `Dual`s through the vjp rhs anyway, so requiring a ForwardDiff-compatible forward rhs does not narrow support; explicit solvers (no Jacobian use) and finite-difference implicit solvers keep current behavior
- in-place ODE, mutable `AbstractFloat` state

## Also fixes a latent bug

With a **sparse user `jac_prototype` and a continuous cost**, the sparse `ftranspose!` adjoint-jac closure was passed to the adjoint solver *without* its sparse prototype (`adjoint_jac_prototype` was gated on `sense.discrete`), so the adjoint solver allocated a dense `J` and `ftranspose!` threw:

```
MethodError: no method matching ftranspose!(::Matrix{Float64}, ::SparseMatrixCSC{Float64, Int64}, ::typeof(-))
```

The adjoint state is just `λ` for these methods, so the transposed prototype is correct for continuous costs too; the gate is removed.

## Benchmarks

2D Brusselator, N=16 (512 states), KenCarp47 forward + adjoint, tol 1e-6, discrete sum cost at 23 time points, `adjoint_sensitivities`. Forward solve: 0.45 s.

| sensealg | master | this PR | speedup |
|---|---|---|---|
| `GaussAdjoint(EnzymeVJP)` | 0.97–1.17 s, 8.05M allocs, 703 MiB | 0.45–0.50 s, 6.5k allocs, 8.8 MiB | **~2.2×** |
| `GaussAdjoint(ReverseDiffVJP(true))` | 5.2–6.9 s, 2.4 GiB | 1.5–1.7 s, 14.7 MiB | **~3.5×** |
| `QuadratureAdjoint(EnzymeVJP)` | 0.98–1.88 s, 705 MiB | 0.64 s, 10.6 MiB | **~1.7×** |

The adjoint solve now costs about the same as the forward solve. Gradients are identical to master (verified digit-for-digit on the benchmark, and against ForwardDiff references in the new tests).

Explicit-solver adjoints are unaffected (profiled: ~88% of time is inside Enzyme vjp calls, which this PR does not touch).

## Tests

Added to `test/Core3/user_vjp.jl`:
- unit test that the fallback Jacobian matches `-(df/du)^T` analytically (both `autodiff=true` and `autodiff=false`)
- gating tests (no fallback for `Tsit5`, no fallback for `Rodas5P(autodiff=AutoFiniteDiff())`)
- end-to-end `adjoint_sensitivities` gradients vs ForwardDiff with `Rodas5P` adjoint solves for `GaussAdjoint(EnzymeVJP)`, `GaussAdjoint(ReverseDiffVJP)`, `QuadratureAdjoint(EnzymeVJP)`
- regression test for the sparse-jac + continuous-cost `ftranspose!` bug (sparse vs dense prototype must agree)

Ran locally: `test/Core3/user_vjp.jl` (all 34 pass), plus full `GROUP=Core1`, `GROUP=Core2` (stiff adjoints), and `GROUP=Core3` package test groups — all pass.

## Process notes

Investigation: cloned repo, benchmarked GaussAdjoint vs Interpolating/Quadrature on Brusselator with ROCK4 (explicit) and KenCarp47 (stiff), profiled both vjp backends, identified the AD-of-vjp-rhs Jacobian as the dominant stiff-path cost (8M allocations, 703 MiB per gradient), implemented the exact-Jacobian fallback, and A/B benchmarked master vs branch in separate environments.

### DIFF
--- a/src/adjoint_common.jl
+++ b/src/adjoint_common.jl
@@ -815,6 +815,79 @@ function getprob(S::SensitivityFunction)
 end
 inplace_sensitivity(S::SensitivityFunction) = isinplace(getprob(S))
 
+"""
+    build_adjoint_jac(sol, sensealg, alg, f, tspan)
+
+Construct a `jac(J_adj, λ, p, t)` for the adjoint ODE of the state-only adjoint
+methods (`GaussAdjoint`, `QuadratureAdjoint`, whose adjoint state is just `λ`).
+The adjoint rhs `-(∂f/∂u)ᵀλ` is linear in `λ`, so its Jacobian is exactly
+`-(∂f/∂u)ᵀ` evaluated at the interpolated forward solution — independent of `λ`.
+
+When the user supplies an analytical `jac` (with a `jac_prototype`), evaluate it
+at `y(t)` and negate-transpose. Otherwise, for dense in-place ODEs solved by a
+ForwardDiff-based implicit method, fall back to differentiating the *forward*
+rhs (ForwardDiff/FiniteDiff per `sensealg`'s `autodiff` setting) and
+negate-transposing. Without this, the implicit adjoint solver differentiates the
+vjp-based adjoint rhs itself, which costs a full vjp (plus a forward-solution
+interpolation) per Jacobian column — and with `ReverseDiffVJP` re-records the
+tape with `Dual` numbers on every Jacobian evaluation.
+
+Returns `nothing` when no cheap exact Jacobian can be built; the adjoint solver
+then falls back to its own AD/FD of the adjoint rhs.
+"""
+function build_adjoint_jac(sol, sensealg, alg, f, tspan)
+    jac_prototype = sol.prob.f.jac_prototype
+    u0 = state_values(sol.prob)
+    if SciMLBase.has_jac(sol.prob.f) && jac_prototype !== nothing
+        _fwd_jac_cache = copy(jac_prototype)
+        _fwd_jac_fn = sol.prob.f.jac
+        _fwd_sol = sol
+        if jac_prototype isa SparseArrays.AbstractSparseMatrixCSC
+            return function (J_adj, _λ, _p, _t)
+                _y = _fwd_sol(_t, continuity = :right)
+                _fwd_jac_fn(_fwd_jac_cache, _y, _p, _t)
+                SparseArrays.ftranspose!(J_adj, _fwd_jac_cache, -)
+                return nothing
+            end
+        else
+            return function (J_adj, _λ, _p, _t)
+                _y = _fwd_sol(_t, continuity = :right)
+                _fwd_jac_fn(_fwd_jac_cache, _y, _p, _t)
+                transpose!(J_adj, _fwd_jac_cache)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas-Claude:
Additional A/B datapoint on a quiet machine — same Brusselator benchmark but with `FBDF()` as forward + adjoint solver (787 forward steps, forward solve ~1.0 s):

| | master | this PR |
|---|---|---|
| `GaussAdjoint(EnzymeVJP)` adjoint | 3.05–3.41 s, 2.70M allocs, 239.7 MiB | 2.84–2.87 s, 17.8k allocs, 9.4 MiB |

FBDF reuses W/Jacobians much more aggressively than KenCarp47, so the wall-time win is smaller (~1.1×), but allocations drop ~150× and memory ~25×. Gradients agree to within the adjoint solve tolerance (the exact Jacobian slightly changes the adjoint solver's Newton/step-size trajectory, so results are not bit-identical with FBDF, unlike the KenCarp47 run; both match the tight-tolerance ForwardDiff reference).

ChrisRackauckas-Claude:
CI note: the `tests / QA (julia 1)` failure is the Aqua **Persistent tasks** check, which fails identically on current master (e.g. [master run 29443609993](https://github.com/SciML/SciMLSensitivity.jl/actions/runs/29443609993)) — it is pre-existing and unrelated to this PR. A separate investigation of that master failure is underway.

PR #1557, opened 2026-07-17
Fix `Mooncake` over `ReverseDiffAdjoint`/`TrackerAdjoint`/`ForwardSensitivity`

### LINKED ISSUE #1510: Core1 timeout: Mooncake stacked over another AD is unsupported (won't-fix, Mooncake#1208)
## Summary
The `Core1` CI group times out (2h, previously 6h) on the **Mooncake-over-another-AD** testsets in `test/Core1/concrete_solve_derivatives.jl` (added in #1420). Failing since ~2026-05-08 (the 6h→2h cap change only altered how it surfaces); **not** caused by the recent (2026-06-28) Enzyme/SDE work.

This is **unsupported and won't be fixed**: it's a documented Mooncake known-limitation — **chalk-lab/Mooncake.jl#1208** (closed won't-fix; "overlay-blind inference", workaround is a hand-written `rrule!!`). Mooncake stacked over another AD's adjoint is not supported.

## Affected testsets (all skipped permanently in #1511)
- `Mooncake with ReverseDiffAdjoint` (`:450`)
- `Mooncake with TrackerAdjoint` (`:455`)
- `Mooncake with ForwardSensitivity (p-only)` (`:494`)

## Mechanism
These differentiate a `solve` with Mooncake stacked over another AD. The `DiffEqBase.solve` overlay returns an `ODESolution{TrackedReal/Dual,…}` under those sensealgs, but Mooncake's inference is overlay-blind and records `ODESolution{Float64,…}`, so `build_rrule`'s typeassert fails. It surfaces as a fast `TypeError`, or — under coverage instrumentation (the CI config) — a multi-hour compile/inference hang that times out Core1 (reproduced locally at 746 s / no-return, Julia 1.12.6).

## Resolution
- #1511 skips the three testsets permanently (`@test_skip`) with a note citing chalk-lab/Mooncake.jl#1208.
- Optional cleanup: the `MooncakeOriginator` delegations at `src/concrete_solve.jl:2146` and `:2387` (#1420) can be dropped, since Mooncake-over-other-AD is unsupported.

This issue can be closed once #1511 merges (kept only as the rationale record for the skip); there is no SciMLSensitivity-side functional fix to track.

AstitvaAggarwal:
Fixes #1510 for real instead of the skip-only patch in #1511.

The problem: `solve` tries to tell Mooncake it's the one differentiating via `set_mooncakeoriginator_if_mooncake`, but that never actually works : `MooncakeOriginator`/`ChainRulesOriginator` are zero-field structs, so Julia just constant-folds the whole overlay away before Mooncake's rule dispatch ever sees it (chalk-lab/Mooncake.jl#1208).

Instead of fixing the detection, this adds three `Mooncake.rrule!!` methods for `solve_up`, specific to `ReverseDiffAdjoint`/`TrackerAdjoint`/ `ForwardSensitivity`, that just build `MooncakeOriginator()` directly since these rules only ever fire under Mooncake anyway. They're narrower than the generic `@from_rrule` rule in `DiffEqBaseMooncakeExt.jl` so dispatch picks them automatically.
 
Un-skips the 3 tests #1511 marked permanently broken.
Tested with explicit alg, auto-alg selection, and alg + legacy positional args (`timeseries_init`/`ts_init`) across all three sensealgs - 8/8 pass against ForwardDiff reference gradients. Full `concrete_solve_derivatives.jl` run passes as well.

cc: @ChrisRackauckas

### DIFF
--- a/ext/SciMLSensitivityMooncakeExt.jl
+++ b/ext/SciMLSensitivityMooncakeExt.jl
@@ -5,7 +5,8 @@ using Mooncake: Mooncake
 import SciMLSensitivity: get_paramjac_config, get_cb_paramjac_config, mooncake_run_ad,
     MooncakeVJP, MooncakeLoaded,
     DiffEqBase, MooncakeAdjoint, _init_originator_gradient,
-    ReverseDiffAdjoint, TrackerAdjoint, ForwardDiffSensitivity
+    ReverseDiffAdjoint, TrackerAdjoint, ForwardDiffSensitivity,
+    EnzymeAdjoint
 using SciMLSensitivity: SciMLBase, SciMLStructures, canonicalize, Tunable, isscimlstructure,
     SciMLStructuresCompatibilityError, convert_tspan,
     has_continuous_callback,
@@ -256,7 +257,7 @@ end
 # AbstractSensitivityAlgorithm}` signature, dispatch prefers these three
 # sensealgs and falls back to the generic rule for everything else.
 const _MooncakeOverAnotherADSensealg = Union{
-    ReverseDiffAdjoint, TrackerAdjoint, ForwardDiffSensitivity,
+    ReverseDiffAdjoint, TrackerAdjoint, ForwardDiffSensitivity, EnzymeAdjoint
 }
 
 function _solve_up_mooncake_over_another_ad(prob, sensealg, u0, p, alg_and_kwargs...; kwargs...)
--- a/ext/SciMLSensitivityMooncakeExt.jl
+++ b/ext/SciMLSensitivityMooncakeExt.jl
@@ -287,6 +287,7 @@ function Mooncake.rrule!!(
         sensealg::Mooncake.CoDual{<:_MooncakeOverAnotherADSensealg},
         u0::Mooncake.CoDual, p::Mooncake.CoDual, alg_and_rest::Mooncake.CoDual...,
     )
+    sensealg isa EnzymeAdjoint && error("EnzymeAdjoint currently is not supported inside of Mooncake autodiff")
     fargs = (f, prob, sensealg, u0, p, alg_and_rest...)
     primals = Mooncake.tuple_map(Mooncake.primal, fargs)
     lazy_rdata = Mooncake.tuple_map(Mooncake.lazy_zero_rdata, primals)
--- a/ext/SciMLSensitivityMooncakeExt.jl
+++ b/ext/SciMLSensitivityMooncakeExt.jl
@@ -4,7 +4,8 @@ using SciMLSensitivity: SciMLSensitivity, FakeIntegrator
 using Mooncake: Mooncake
 import SciMLSensitivity: get_paramjac_config, get_cb_paramjac_config, mooncake_run_ad,
     MooncakeVJP, MooncakeLoaded,
-    DiffEqBase, MooncakeAdjoint, _init_originator_gradient
+    DiffEqBase, MooncakeAdjoint, _init_originator_gradient,
+    ReverseDiffAdjoint, TrackerAdjoint, ForwardSensitivity
 using SciMLSensitivity: SciMLBase, SciMLStructures, canonicalize, Tunable, isscimlstructure,
     SciMLStructuresCompatibilityError, convert_tspan,
     has_continuous_callback,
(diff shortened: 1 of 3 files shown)


AstitvaAggarwal:
a bump on this @ChrisRackauckas, thanks.

ext/SciMLSensitivityMooncakeExt.jl
```diff
@@ -239,4 +240,102 @@ function SciMLBase._concrete_solve_adjoint(
...
+# any rule dispatch happens. These hand-written `solve_up` rules sidestep
+# detection entirely -- they only ever run under Mooncake, so they construct
+# `MooncakeOriginator()` directly and dispatch into the existing
+# `MooncakeOriginator` methods added in #1420 (src/concrete_solve.jl), which
+# re-solve with a plain `Float64` primal Mooncake can build a `CoDual` for.
+# Being more specific than the generic rule's `Union{Nothing,
+# AbstractSensitivityAlgorithm}` signature, dispatch prefers these three
+# sensealgs and falls back to the generic rule for everything else.
+const _MooncakeOverAnotherADSensealg = Union{
+    ReverseDiffAdjoint, TrackerAdjoint, ForwardSensitivity,
```
ChrisRackauckas:
```suggestion
    ReverseDiffAdjoint, TrackerAdjoint, ForwardDiffSensitivity,
```

ext/SciMLSensitivityMooncakeExt.jl
```diff
@@ -4,7 +4,8 @@ using SciMLSensitivity: SciMLSensitivity, FakeIntegrator
 using Mooncake: Mooncake
 import SciMLSensitivity: get_paramjac_config, get_cb_paramjac_config, mooncake_run_ad,
     MooncakeVJP, MooncakeLoaded,
-    DiffEqBase, MooncakeAdjoint, _init_originator_gradient
+    DiffEqBase, MooncakeAdjoint, _init_originator_gradient,
+    ReverseDiffAdjoint, TrackerAdjoint, ForwardSensitivity
```
ChrisRackauckas:
```suggestion
    ReverseDiffAdjoint, TrackerAdjoint, ForwardDiffSensitivity
```

ChrisRackauckas (approved):

ChrisRackauckas:
Sorry missed the bump. It needed a few tweaks but should be good to go.

AstitvaAggarwal:
@ChrisRackauckas I don't think `EnzymeAdjoint` under `Mooncake` actually works yet, it was added to `_MooncakeOverAnotherADSensealg` in https://github.com/SciML/SciMLSensitivity.jl/pull/1557/commits/7a62a7f17a6694b4d26dbf598f744b44d77ae08c, but there's no test exercising `EnzymeAdjoint` with `Mooncake` as the outer AD (the existing `EnzymeAdjoint` testset explicitly excludes `Mooncake` as an outer backend). 

I had confirmed by direct testing on #1571 that this combination throws a `TypeError` from a type mismatch inside Mooncake's rule machinery (same symptom reported in #1578). Might be worth pulling EnzymeAdjoint back out of the union here until that's actually fixed.

ext/SciMLSensitivityMooncakeExt.jl
```diff
@@ -239,4 +241,102 @@ function SciMLBase._concrete_solve_adjoint(
...
+    )
+end
+
+function Mooncake.rrule!!(
+        f::Mooncake.CoDual{typeof(DiffEqBase.solve_up)},
+        prob::Mooncake.CoDual{<:SciMLBase.AbstractDEProblem},
+        sensealg::Mooncake.CoDual{<:_MooncakeOverAnotherADSensealg},
+        u0::Mooncake.CoDual, p::Mooncake.CoDual, alg_and_rest::Mooncake.CoDual...,
+    )
+    fargs = (f, prob, sensealg, u0, p, alg_and_rest...)
```
ChrisRackauckas, line 291:
```suggestion
    )
    sensealg isa EnzymeAdjoint && error("EnzymeAdjoint currently is not supported inside of Mooncake autodiff")
    fargs = (f, prob, sensealg, u0, p, alg_and_rest...)
```

ChrisRackauckas:
Okay that seemed like an omission because it was labelled as being "the autodiff methods", while leaving one out. Instead I made the error explicit. I think we can go with this.

PR #1559, opened 2026-07-18
Update optimization_sde doc test

wsmoses:


wsmoses:
@ChrisRackauckas the optimization_sde failure is from the other [still zygote] example, not this one -- which passes, so I'm going to merge

PR #1567, opened 2026-07-23
QA: retry the persistent-tasks wrapper precompile (fix flaky Aqua failure)

### LINKED ISSUE #1554: Aqua persistent-task QA misclassifies retryable wrapper precompile failure
## Failure

Current default-branch QA fails in the Aqua persistent-task check:

- Current master run: https://github.com/SciML/SciMLSensitivity.jl/actions/runs/29395878784/job/87291188165
- Same failure on the LinearSolve 5 compat PR: https://github.com/SciML/SciMLSensitivity.jl/actions/runs/29289163457/job/86948781214
- Same failure on follow-up PR CI: https://github.com/SciML/SciMLSensitivity.jl/actions/runs/29400248355/job/87304080406
- Last known green master QA: https://github.com/SciML/SciMLSensitivity.jl/actions/runs/29181902897/job/86621098845

Aqua reports:

```
done.log was not created, but precompilation exited
```

This is the misleading failure mode already described by https://github.com/JuliaTesting/Aqua.jl/issues/315.

## Hidden subprocess failure reproduced locally

I recreated the package that `Aqua.test_persistent_tasks` generates: a fresh wrapper package with `using SciMLSensitivity`, followed by writing `done.log`. I ran its first precompile in a clean writable depot on Julia 1.12.6 with `JULIA_NUM_PRECOMPILE_TASKS=4`.

With the current dependency set, the first `Pkg.precompile()` exits with status 0 but marks the wrapper `?` and reports:

```
Module NonlinearSolveBaseChainRulesCoreExt ... is missing from the cache.
1 dependencies failed but may be precompilable after restarting julia
```

The wrapper source never reaches the status write, so `done.log` is absent and Aqua reports a persistent-task failure. An immediate second `Pkg.precompile()` takes about 21 seconds, precompiles the wrapper successfully, and creates `done.log`. This demonstrates a retryable parallel-precompile/cache-ordering failure, not a persistent task in SciMLSensitivity.

## Controlled results

All cases used a new first depot and the same current package sources:

| Resolved versions | First precompile | done.log |
| --- | --- | --- |
| LinearSolve 5.0.0, NonlinearSolveBase 2.35.0 | wrapper retryable-failed, exit 0 (4m30s) | absent |
| LinearSolve 4.3.0, NonlinearSolveBase 2.35.0 | identical retryable failure (4m54s) | absent |
| LinearSolve 4.3.0, NonlinearSolveBase 2.33.0 | 66 dependencies succeeded (4m37s) | present |
| LinearSolve 5.0.0, NonlinearSolveBase 2.34.0 | 66 dependencies succeeded (2m51s) | present |
| LinearSolve 5.0.0, NonlinearSolveBase 2.34.1 | 66 dependencies succeeded (2m45s) | present |

The 2.34.1 result is important: the July 13 CI failure resolved the same NonlinearSolveBase 2.34.1. The outcome is therefore scheduling-sensitive rather than a deterministic semantic regression at one source commit.

The CI dependency history also changed across the green/red boundary:

- Green July 12: NonlinearSolveBase 2.33.0, SciMLBase 3.35.1, OrdinaryDiffEqCore 4.6.0, Enzyme 0.13.182, LinearSolve 4.3.0.
- Red July 13: NonlinearSolveBase 2.34.1, SciMLBase 3.36.0, OrdinaryDiffEqCore 4.7.0, Enzyme 0.13.184, LinearSolve 5.0.0.
- Current red: current versions including NonlinearSolveBase 2.35.0.

A source-only interval initially pointed at the LinearSolve compat merge, but the controlled LinearSolve 4.3 / NonlinearSolveBase 2.35 run disproves a LinearSolve-only fix. I tested and discarded that compat revert; the SciMLSensitivity worktree remains unchanged.

## Suggested direction

The robust fix belongs in the wrapper precompile handling: preserve and surface the first precompile output, then retry when Pkg reports that the wrapper may be precompilable after restarting. The persistent-task timeout should still run after the wrapper successfully reaches the status write. Disabling or marking the persistent-task test broken would hide real failures and is not an acceptable workaround.

No SciMLSensitivity code change is proposed because the package precompiles successfully on retry, and current evidence does not justify an arbitrary dependency pin.

ChrisRackauckas-Claude:
**Please ignore until reviewed by @ChrisRackauckas.** Draft, opened by an agent.

Closes #1554.

## Root cause (confirmed by reproduction)

The QA group's Aqua persistent-tasks check fails intermittently on master with `done.log was not created, but precompilation exited`. It is **not** a persistent task in SciMLSensitivity — it's a retryable cold-cache parallel-precompile failure being misclassified, exactly as #1554 hypothesized.

Aqua's `test_persistent_tasks` spawns a throwaway wrapper package that `using`s SciMLSensitivity and writes a `done.log` sentinel from inside the precompile process; a package that never lets that process exit within `tmax` is flagged as holding a persistent `Task`. The wrapper's freshly-resolved dependency set precompiles **cold and in parallel**, and one module hits a *retryable* `"… is missing from the cache … 1 dependencies failed but may be precompilable after restarting julia"` failure. `Pkg.precompile` then exits without compiling the wrapper, so `done.log` is never written → misreported as a persistent task. The culprit module is scheduling-dependent (I reproduced it as `Accessors`; the issue saw `NonlinearSolveBaseChainRulesCoreExt`), which is why it's flaky rather than a deterministic regression. A `tmax` bump does **not** help — `tmax` governs the wait *after* `done.log` appears; here the process exits *before* writing it. The correct fix is a **retry**.

## Fix

Disable Aqua's single-shot `persistent_tasks` sub-check and run a local `has_persistent_tasks_with_retry` that faithfully reproduces Aqua's wrapper (its public API only exposes the asserting single-shot `test_persistent_tasks`; `has_persistent_tasks` is non-public, so per the public-API rule the small helper is reproduced rather than reached into) but **retries when the child precompile exits without writing `done.log`** (up to 3 attempts). Adds `Pkg` as a test dependency (the helper calls it directly).

**Not a cop-out:** it does not skip, `@test_broken`, or loosen anything. A genuine persistent task writes `done.log` and *then* keeps the process alive past `tmax`, which is still detected on every attempt; the retry only fires on the missing-`done.log` path, which is unreachable for a real background task.

## Verification (local, Julia 1.12.6)

- **3× with a wiped compile cache**: the retryable failure fired on all three cold runs and the retry recovered every time (`pass=true`).
- **1× end-to-end `GROUP=QA`** (`Pkg.test()`): `Quality Assurance | 22 pass / 22 total`, exit 0.
- Runic reports no changes.

## Note on placement

This is the repo-local unblock for the master red. The **durable** fix belongs upstream in `SciMLTesting`'s `with_clean_persistent_tasks_sources` / `run_qa` (or JuliaTesting/Aqua.jl#315), so the retry is fleet-wide and every SciML repo can drop the local copy. Happy to prepare that as a follow-up and revert this to a one-liner once it lands — flagging for your call.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLSensitivity"
 uuid = "1ed8b502-d754-442c-8d5d-10ac956f44a1"
-version = "7.116.3"
+version = "7.116.4"
 authors = ["Christopher Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
 
 [deps]
@@ -109,6 +109,7 @@ OrdinaryDiffEqNonlinearSolve = "2"
 OrdinaryDiffEqRosenbrock = "2"
 OrdinaryDiffEqSDIRK = "2"
 OrdinaryDiffEqStabilizedRK = "2"
+Pkg = "1.10"
 PreallocationTools = "1.2"
 QuadGK = "2.11.3"
 Random = "1.10"
@@ -164,6 +165,7 @@ OrdinaryDiffEqNonlinearSolve = "127b3ac7-2247-4354-8eb6-78cf4e7c58e8"
 OrdinaryDiffEqRosenbrock = "43230ef6-c299-4910-a778-202eb28ce4ce"
 OrdinaryDiffEqSDIRK = "2d112036-d095-4a1e-ab9a-08536f3ecdbf"
 OrdinaryDiffEqStabilizedRK = "358294b1-0aab-51c3-aafe-ad5ab194a2ad"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 SciMLTesting = "09d9d899-5365-40a9-917a-5f67fddea283"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Superseded by the durable upstream fix: **SciML/SciMLTesting.jl#32** puts the race-tolerant persistent-tasks retry into `run_qa` itself, fleet-wide.

Closing this repo-local draft. Once SciMLTesting 2.4.1 registers, SciMLSensitivity needs **no change** — its existing `run_qa(SciMLSensitivity; …)` in `test/QA/aqua.jl` picks up the retry automatically (the `SciMLTesting = "2.1"` compat already resolves to 2.4.1), so the QA flake is fixed without carrying the 88-line local reproduction of Aqua's wrapper.

Why not the alternatives, for the record:
- **Serial precompile** (the first "clean" idea): verified broken — it removes the parallel race but is ~6.6× slower, writing `done.log` ~837s before a cold serial tree finishes and tripping Aqua's `tmax=10s`. Failed 3/3 cold.
- **Repo-local retry** (this PR): works (validated 3/3 cold), but reproduces Aqua internals per-repo. #32 is the same retry, centralized.
- The truly-upstream home is Aqua itself (JuliaTesting/Aqua.jl#315).

PR #1571, opened 2026-07-27
native `MooncakeAdjoint` `solve_up` rule + explicit HVP errors for other `sensealgs`

AstitvaAggarwal:
This is about getting Mooncake's HVP (`value_and_hvp!!`) actually working for SciMLSensitivity, and explicitly erroring out where not supported. closes/addresses SciML/SciMLSensitivity.jl#1427, the following PR gives a baseline for future work into supporting other AD backends as well.

## `MooncakeAdjoint` sensealg

`MooncakeAdjoint`'s gradient was going through the generic `@from_rrule`-based `solve_up` primitive, which round-trips through `mooncake_tangent`/`to_cr_tangent` for ChainRules compatibility. That conversion has no case for something as deeply nested as an `ODESolution`, so it throws.

Since `MooncakeAdjoint`'s gradient never actually needs to leave Mooncake's native tangent representation, this adds a dedicated `rrule!!` for `solve_up` restricted to `MooncakeAdjoint`, built around a nested `Mooncake.build_rrule` call instead of going through ChainRules at all. Gradient and HVP both work correctly now, verified against ForwardDiff on a neural ODE (see `"Mooncake with MooncakeAdjoint"` in `test/Core1/concrete_solve_derivatives.jl`).

## Other `sensealgs`

`value_and_hvp!!` only actually works for `MooncakeAdjoint`. Before this PR, every other sensealg either crashed with a confusing internal stacktrace or, worse, could silently give the wrong HVP. Each one now fails with an explicit, accurate error instead:

- `ReverseDiffAdjoint`/`TrackerAdjoint` hand their gradient off to ReverseDiff.jl/Tracker.jl, and Mooncake can't safely forward-differentiate through another AD tool's internals. Plain gradients are untouched and still work (via the existing `MooncakeOriginator`-based dispatch from #1420/#1510).
- `ForwardSensitivity` doesn't delegate to another AD package, but forward-mode tracing into its own ForwardDiff.jl-based Jacobian machinery hits a Mooncake type-prediction limitation (chalk-lab/Mooncake.jl#1209). Its plain gradient still works too.
- `EnzymeAdjoint`/`ZygoteAdjoint` never actually worked under Mooncake at all, even for a plain gradient, confirmed by direct testing -- for two unrelated reasons. They now say so instead of crashing.

Also removed a dead forward-mode `solve_up` rule that turned out to be unreachable: neither `MooncakeAdjoint`'s nor the other sensealgs' reverse rules ever call back into `solve_up` internally during HVP composition, so a forward-mode primitive sitting there could never fire. And swapped a dependency on Mooncake's internal `__call_rule` for a small local equivalent -- identical on Julia >= 1.11, and keeps the same Julia 1.10 compiler-bug workaround Mooncake's own version has.

## Tests

Added/updated coverage in `test/Core1/concrete_solve_derivatives.jl` for all five sensealgs' behavior under Mooncake described above, including un-skipping a few plain-gradient tests that turned out to already be passing. This now includes the `"Mooncake with MooncakeAdjoint"` gradient+HVP test itself (both checked against an independent nested-`ForwardDiff` ground truth), previously staged separately.

Also fixes the QA/ExplicitImports ignore list for the new Mooncake symbols this PR's rewrite introduces (`mooncake_tangent`, `to_fwds`, the explicitly-imported rule-writing API, etc.) -- this was the direct cause of the QA CI failure on earlier pushes.

## Dependencies

The `"Mooncake with MooncakeAdjoint"` HVP test needs three prerequisite PRs fixing separate forward-mode gaps it exercises (HVP is forward-over-reverse, so it hits code paths a plain reverse-mode gradient never touches):
- SciML/OrdinaryDiffEq.jl#4201 (`promote_f` as an opaque Mooncake primitive) -- **merged**
- SciML/OrdinaryDiffEq.jl#4202 (Mooncake primitive rule for `initialize_saveat`) -- **merged**
- SciML/FunctionWrappersWrappers.jl#84 (forward-mode rules for calling a `FunctionWrappersWrapper`) -- open, mergeable, blocked only by a QA ignore-list gap (fix pushed)

chalk-lab/Mooncake.jl#1259 (three tangent-construction crashes hit while debugging this) has also landed and is merged.

Until #84 merges and is released, the `MooncakeAdjoint` HVP test may still fail in CI on a `FunctionWrapper`/forward-mode gap; the plain-gradient tests and the other four sensealgs' explicit-error tests do not depend on it.

### DIFF
--- a/ext/SciMLSensitivityMooncakeExt.jl
+++ b/ext/SciMLSensitivityMooncakeExt.jl
@@ -2,11 +2,13 @@ module SciMLSensitivityMooncakeExt
 
 using SciMLSensitivity: SciMLSensitivity, FakeIntegrator
 using Mooncake: Mooncake
+import Mooncake: MinimalCtx, ForwardMode, Dual, CoDual, frule!!, rrule!!, @is_primitive,
+    @zero_derivative
 import SciMLSensitivity: get_paramjac_config, get_cb_paramjac_config, mooncake_run_ad,
     MooncakeVJP, MooncakeLoaded,
     DiffEqBase, MooncakeAdjoint, _init_originator_gradient,
     ReverseDiffAdjoint, TrackerAdjoint, ForwardSensitivity,
-    EnzymeAdjoint
+    EnzymeAdjoint, ZygoteAdjoint
 using SciMLSensitivity: SciMLBase, SciMLStructures, canonicalize, Tunable, isscimlstructure,
     SciMLStructuresCompatibilityError, convert_tspan,
     has_continuous_callback,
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Rebase?

AstitvaAggarwal:
I will have some followup PRs for fixing the CI errors for Julia lts and v1.11

PR #1581, opened 2026-08-02
Fix Mooncake ForwardSensitivity dispatch

ChrisRackauckas-Claude:
## Summary

- dispatch Mooncake-over-another-AD solves for `ForwardSensitivity`, matching the rule body and existing regression
- cover both the default and `autodiff = false` ForwardSensitivity configurations for p-only differentiation

## Root cause

The original Mooncake rule change imported and dispatched on `ForwardSensitivity`, but a later review commit changed both sites to `ForwardDiffSensitivity` while the implementation, comments, and test continued to use `ForwardSensitivity`. The p-only Mooncake call therefore fell through to the generic rule and produced an invalid cotangent type.

This restores the intended `ForwardSensitivity` dispatch. No public API changes are involved.

## Validation

- exact clean-master Core1 baseline: 234 passed, 4 errored, 5 broken
- combined patched Core1 with SciML/OrdinaryDiffEq.jl#4101: 239 passed, 0 errored, 5 broken (244 total), 41m15.6s
- whole-repository `Runic --check .` passed

The other three clean-master Core1 errors are the independent Rosenbrock tracked-state regression addressed by SciML/OrdinaryDiffEq.jl#4101.

Ignore this PR until reviewed by @ChrisRackauckas.

### DIFF
--- a/ext/SciMLSensitivityMooncakeExt.jl
+++ b/ext/SciMLSensitivityMooncakeExt.jl
@@ -5,7 +5,7 @@ using Mooncake: Mooncake
 import SciMLSensitivity: get_paramjac_config, get_cb_paramjac_config, mooncake_run_ad,
     MooncakeVJP, MooncakeLoaded,
     DiffEqBase, MooncakeAdjoint, _init_originator_gradient,
-    ReverseDiffAdjoint, TrackerAdjoint, ForwardDiffSensitivity,
+    ReverseDiffAdjoint, TrackerAdjoint, ForwardSensitivity,
     EnzymeAdjoint
 using SciMLSensitivity: SciMLBase, SciMLStructures, canonicalize, Tunable, isscimlstructure,
     SciMLStructuresCompatibilityError, convert_tspan,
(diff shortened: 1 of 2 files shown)


AstitvaAggarwal:
Thanks for this @ChrisRackauckas, I think I should have named that union better since it dint quite convey why `ForwardSensitivity` was there which caused this confusion (also im unable to open/respond to your comment here for some reason https://github.com/AstitvaAggarwal/SciMLSensitivity.jl/commit/e17b48119769316e6639999e375efa6c0665377b). 
I had checked while working on #1557 that only `ForwardSensitivity` broke (`ForwardDiffSensitivity` was  fine), reason being that `ForwardSensitivity` uses `ForwardDiff.jl` internally for local per-step Jacobians, which makes its `ODESolution` type deep enough to trip up `Mooncake`'s generic auto-derived rule.

Putting `ForwardSensitivity` back into `_MooncakeOverAnotherADSensealg` is fine here as it then uses the `rrule!!` from #1557 again instead of the generic path. Maybe we can also rename the union through this PR?

Also note: PR #1571 touches this same union (predates #1557) so it still has ForwardSensitivity in the union. I'll rebase it onto master once this PR merges.

ChrisRackauckas:
ForwardDiffSensitivity works though?

AstitvaAggarwal:
yeah

PR #1606, opened 2026-08-20
fix: don't discard the caller's repack in automatic_sensealg_choice

### LINKED ISSUE #1605: `automatic_sensealg_choice` discards the caller's repack and silently falls back to numerical VJPs for MTK problems
**Describe the bug 🐞**

When `solve` is differentiated without an explicit `sensealg`,
`SciMLBase._concrete_solve_adjoint(..., sensealg::Nothing, ...)` canonicalizes
`p::MTKParameters` into `(tunables, repack)` and calls
`automatic_sensealg_choice(prob, u0, tunables, verbose, repack, p)`
(concrete_solve.jl:477–484, v7.118.0). Inside `automatic_sensealg_choice`,
the `p` parameter is now the **tunables `Vector`**, and since
`isscimlstructure(::Vector) == true`, the function re-canonicalizes it
(concrete_solve.jl:219):

```julia
elseif isscimlstructure(p)
    tunables, repack, _ = canonicalize(Tunable(), p)   # p is already the tunables Vector here
```

`canonicalize(Tunable(), ::Vector)` returns a `SciMLStructures.ArrayRepack`,
which **overwrites the caller's MTKParameters repack** with an identity-like
one. Every VJP probe in `inplace_vjp` then calls the RHS with the bare
tunables `Vector` instead of the reconstructed `MTKParameters`.

For any MTK system whose generated RHS reads a portion beyond the tunables —
a non-tunable (constant) parameter, or the `Nonnumeric` portion (which is
where ModelingToolkitNeuralNets stores the Lux chain, i.e. **every MTKNN
UDE**) — the probe *primal* throws
`BoundsError: attempt to access Tuple{Vector{Float64}, Nothing} at index [3]`.
All AD probes (EnzymeVJP, EnzymeVJP w/ runtime activity, ReverseDiffVJP) fail
on this non-AD error, and the chooser returns
`GaussAdjoint(autodiff = false, autojacvec = false)` — numerical
(finite-difference) VJPs.

This is silent: the probe warnings are suppressed whenever the solve-level
`verbose` routes to `false` (common in downstream libraries), so users who set
up Enzyme/Zygote end-to-end train on FD VJPs without any indication. The only
visible symptom in the wild was Lux's ArrayInterfaceReverseDiffExt
"TrackedReal → TrackedArray" warning fired by the dying ReverseDiff tape
probe.

**Expected behavior**

`automatic_sensealg_choice` should probe with the caller's `repack` (or the
original `p`) so the probes exercise the RHS with the parameter object it will
actually receive. With the correct repack, the Enzyme probe passes and the
choice would be `GaussAdjoint(autojacvec = EnzymeVJP())` here.

**Minimal Reproducible Example 👇**

Without MRE, we would only be able to help you to a limited extent, and attention to the issue would be limited. to know more about MRE refer to [wikipedia](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and [stackoverflow](https://stackoverflow.com/help/minimal-reproducible-example).

```julia
using ModelingToolkit
using SciMLBase
using SciMLSensitivity
using SciMLStructures
using SciMLStructures: Tunable, canonicalize

# > 100 tunables so automatic_sensealg_choice skips the ForwardDiffSensitivity
# short-circuit and actually reaches the VJP probes. `c` is non-tunable, so the
# generated RHS must read a portion of MTKParameters beyond the tunables.
@parameters k[1:128] = 0.01 .* ones(128)
@parameters c = 3.0 [tunable = false]
@variables x(t) = 1.0
@named sys = System([D(x) ~ -c * sum(k) * x], t)
sys = mtkcompile(sys)
prob = ODEProblem(sys, [], (0.0, 1.0))

tunables, repack, _ = canonicalize(Tunable(), prob.p)
f = SciMLBase.unwrapped_f(prob.f)
du = zero(prob.u0)

println("n_tunables = ", length(tunables))

# The RHS works with correctly repacked parameters:
f(du, copy(prob.u0), repack(copy(tunables)), 0.0)
println("RHS with repack(tunables)::", typeof(repack(tunables)).name.name, ": OK")

# ... but the probes inside automatic_sensealg_choice call it with the bare
# Vector, because the function re-canonicalizes its (already canonical) `p`
# argument and overwrites the caller's repack with SciMLStructures.ArrayRepack:
try
    f(du, copy(prob.u0), copy(tunables), 0.0)
    println("RHS with bare tunables Vector: OK (bug NOT reproduced)")
catch e
    println("RHS with bare tunables Vector: ", sprint(showerror, e))
end

# What the probes SHOULD do (caller's repack): the Enzyme probe passes.
vjp_correct = SciMLSensitivity.inplace_vjp(prob, prob.u0, copy(tunables), true, repack)
println("inplace_vjp with the caller's repack chooses: ", vjp_correct)

# What actually happens (this is the exact call from _concrete_solve_adjoint,
# concrete_solve.jl:484): the repack argument is shadowed at :219 and every
# probe dies on the BoundsError above → numerical fallback. With verbose
# still on you can see the probe warnings; with the default solve verbosity
# routed in (e.g. `verbose = false` from a library) it is completely silent.
SciMLSensitivity.STACKTRACE_WITH_VJPWARN[] = true
chosen = SciMLSensitivity.automatic_sensealg_choice(
    prob, prob.u0, tunables, true, repack, prob.p)
println("automatic_sensealg_choice returns: ", chosen)
@assert chosen isa GaussAdjoint && chosen.autojacvec === false
println("=> silent finite-difference VJPs despite Enzyme (and ReverseDiff) handling this RHS fine")
```

**Error & Stacktrace ⚠️**

```julia

```

**Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia
Status `~/work/dev/upstream_bugs/scimlsens_repack_shadowing/Project.toml`
  [7da242da] Enzyme v0.13.199
  [961ee093] ModelingToolkit v11.39.1
  [b1df2697] OrdinaryDiffEqTsit5 v2.1.3
  [0bca4576] SciMLBase v3.49.1
  [1ed8b502] SciMLSensitivity v7.118.0
  [53ae85a6] SciMLStructures v1.10.4
```

  - Output of `using Pkg; Pkg.status(; mode = PKGMODE_MANIFEST)`

```julia

```

  - Output of `versioninfo()`

```julia
Julia Version 1.12.7
Commit 6d172b025e4 (2026-08-15 08:05 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × Intel(R) Core(TM) i9-14900K
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, alderlake)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 32 virtual cores)
```

**Additional context**

SebastianM-C:
Fixes #1605.

## What was wrong

`SciMLBase._concrete_solve_adjoint(..., sensealg::Nothing, ...)` canonicalizes `p` (e.g. `MTKParameters`) into `(tunables, repack)` and calls `automatic_sensealg_choice(prob, u0, tunables, verbose, repack, p)`. Inside, the ODE/SDE method re-canonicalized its `p` argument — which at that point is already the bare tunables `Vector`, and `isscimlstructure(::Vector) == true` — so the caller's repack was overwritten with `SciMLStructures.ArrayRepack` (identity-like).

Every VJP probe in `inplace_vjp` then called the RHS with the bare tunables `Vector` instead of the reconstructed parameter object. For any MTK system whose RHS reads a portion beyond the tunables (a non-tunable parameter, or the `Nonnumeric` portion — where ModelingToolkitNeuralNets stores the Lux chain, i.e. every MTKNN UDE), the probe *primal* throws `BoundsError`, all AD probes "fail", and the chooser silently returns `GaussAdjoint(autodiff = false, autojacvec = false)` — numerical (finite-difference) VJPs. Silent whenever solve-level `verbose` routes to `false`, which is common in downstream libraries.

## The fix

- In the ODE/SDE `automatic_sensealg_choice` method, trust the caller's `(p, repack)` when a `repack` is provided (i.e. `p` is already the canonical tunables), and only self-canonicalize when `repack === nothing` (that entry path is exercised by `test/Core8/parameter_initialization.jl`). This matches what the `ConcreteNonlinearProblem` method already does.
- Pass `tunables` (not `p`) to `inplace_vjp` in the in-place branch, so the self-canonicalizing path also probes with a `(tunables, repack)` pair that actually match.

## Verification

The MWE from #1605 now selects `GaussAdjoint(autojacvec = EnzymeVJP())` instead of `GaussAdjoint(autodiff = false, autojacvec = false)`. Added it (condensed) as a regression test in `test/Core3/automatic_sensealg_choice.jl`: it asserts the chosen `autojacvec` is not a `Bool`, i.e. no silent numerical fallback, without pinning which AD VJP wins the probe.

### DIFF
--- a/test/Core3/automatic_sensealg_choice.jl
+++ b/test/Core3/automatic_sensealg_choice.jl
@@ -61,29 +61,3 @@ let N = 150
     @test sensealg_sparse isa GaussAdjoint
     @test sensealg_sparse.autojacvec isa EnzymeVJP
 end
-
-# Regression test for #1605: `automatic_sensealg_choice` used to re-canonicalize the
-# already-canonical tunables vector, overwriting the caller's repack (e.g. for
-# `MTKParameters`) with an identity-like `ArrayRepack`. Every VJP probe then called the
-# RHS with the bare tunables vector; any RHS reading a non-tunable portion threw in the
-# probe primal, so the chooser silently fell back to numerical VJPs
-# `GaussAdjoint(autodiff = false, autojacvec = false)`. More than 100 tunables are needed
-# to get past the `ForwardDiffSensitivity` short-circuit and reach the probes.
-using ModelingToolkit
-using ModelingToolkit: t_nounits as t, D_nounits as D
-
-let
-    @parameters k[1:128] = 0.01 .* ones(128)
-    @parameters c = 3.0 [tunable = false]
-    @variables x(t) = 1.0
-    @named sys = System([D(x) ~ -c * sum(k) * x], t)
-    sys = mtkcompile(sys)
-    prob = ODEProblem(sys, [], (0.0, 1.0))
-
-    tunables, repack, _ = SciMLStructures.canonicalize(SciMLStructures.Tunable(), prob.p)
-    sensealg_mtk = SciMLSensitivity.automatic_sensealg_choice(
-        prob, prob.u0, tunables, true, repack, prob.p
-    )
-    @test sensealg_mtk isa GaussAdjoint
-    @test !(sensealg_mtk.autojacvec isa Bool)
-end
--- a/test/Core3/automatic_sensealg_choice.jl
+++ b/test/Core3/automatic_sensealg_choice.jl
@@ -61,3 +61,29 @@ let N = 150
     @test sensealg_sparse isa GaussAdjoint
     @test sensealg_sparse.autojacvec isa EnzymeVJP
 end
+
+# Regression test for #1605: `automatic_sensealg_choice` used to re-canonicalize the
+# already-canonical tunables vector, overwriting the caller's repack (e.g. for
+# `MTKParameters`) with an identity-like `ArrayRepack`. Every VJP probe then called the
+# RHS with the bare tunables vector; any RHS reading a non-tunable portion threw in the
+# probe primal, so the chooser silently fell back to numerical VJPs
+# `GaussAdjoint(autodiff = false, autojacvec = false)`. More than 100 tunables are needed
+# to get past the `ForwardDiffSensitivity` short-circuit and reach the probes.
+using ModelingToolkit
+using ModelingToolkit: t_nounits as t, D_nounits as D
+
+let
+    @parameters k[1:128] = 0.01 .* ones(128)
+    @parameters c = 3.0 [tunable = false]
+    @variables x(t) = 1.0
+    @named sys = System([D(x) ~ -c * sum(k) * x], t)
+    sys = mtkcompile(sys)
+    prob = ODEProblem(sys, [], (0.0, 1.0))
+
+    tunables, repack, _ = SciMLStructures.canonicalize(SciMLStructures.Tunable(), prob.p)
+    sensealg_mtk = SciMLSensitivity.automatic_sensealg_choice(
+        prob, prob.u0, tunables, true, repack, prob.p
+    )
+    @test sensealg_mtk isa GaussAdjoint
+    @test !(sensealg_mtk.autojacvec isa Bool)
+end
--- a/src/concrete_solve.jl
+++ b/src/concrete_solve.jl
@@ -215,6 +215,14 @@ function automatic_sensealg_choice(
 
     if p === nothing || p isa SciMLBase.NullParameters
         tunables, repack = p, identity
+    elseif repack !== nothing
+        # The caller has already canonicalized: `p` is the tunables and `repack`
+        # reconstructs the original parameter object. Canonicalizing again would
+        # replace the caller's repack with an identity-like one (e.g.
+        # `SciMLStructures.ArrayRepack` for `MTKParameters`), making every VJP
+        # probe call the RHS with the bare tunables vector instead of the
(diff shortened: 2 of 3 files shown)


test/Core3/automatic_sensealg_choice.jl
```diff
@@ -61,3 +61,29 @@ let N = 150
...
+    sys = mtkcompile(sys)
+    prob = ODEProblem(sys, [], (0.0, 1.0))
+
+    tunables, repack, _ = SciMLStructures.canonicalize(SciMLStructures.Tunable(), prob.p)
+    sensealg_mtk = SciMLSensitivity.automatic_sensealg_choice(
+        prob, prob.u0, tunables, true, repack, prob.p
+    )
+    @test sensealg_mtk isa GaussAdjoint
+    @test !(sensealg_mtk.autojacvec isa Bool)
+end
```
ChrisRackauckas:
```suggestion
```

PR #1629, opened 2026-08-29
Support structured parameters and untracked problem callbacks in callback adjoints

pankgeorg:
Fixes the SciMLSensitivity side of SciML/ModelingToolkit.jl#5018 (companion: SciML/ModelingToolkit.jl#5042, required by the new tests).

The callback VJP path assumed `p` is a plain `Vector`; it now goes through SciMLStructures for structured parameters (plain arrays unchanged), `FakeIntegrator` implements the SymbolicIndexingInterface accessors, tracked callbacks keep the newer SciMLBase fields, `reset_p` accepts a bare callback, and `adjoint_sensitivities` uses the problem's callbacks when tracked or throws an explanatory `ArgumentError` when not (instead of a `FieldError` or a silently uncorrected gradient).

Tested: new `test/Core8/mtk_events.jl` (47 pass, 1 `@test_broken`: compiled ReverseDiff tape with an affect writing a non-tunable parameter); `Callbacks1` 361/361, `Callbacks2` 211/211. Core8 needs the MTK companion released; compat bump to follow.

### DIFF
--- a/docs/src/manual/differential_equation_sensitivities.md
+++ b/docs/src/manual/differential_equation_sensitivities.md
@@ -179,29 +179,13 @@ and DDEs. The continuous adjoint sensitivities `BacksolveAdjoint`, `Interpolatin
 the event terminates the time evolution and several states are saved. Currently,
 the continuous adjoint sensitivities do not support multiple events per time point.
 
-#### Structured parameters and events
-
-Events are supported for problems whose parameter object is a
-[SciMLStructure](https://github.com/SciML/SciMLStructures.jl) rather than a plain
-vector — most notably the parameter object of a
-[ModelingToolkit.jl](https://docs.sciml.ai/ModelingToolkit/stable/) system with
-`continuous_events` or `discrete_events`. The gradient is taken with respect to
-the tunable portion of the parameters, and an affect that assigns to a parameter
-is handled as long as the parameter object is a mutable SciMLStructure.
-
 #### Events with `adjoint_sensitivities`
 
 The adjoint of a hybrid system needs the event times and the state and parameter
 values just before each event. Differentiating through `solve` records them
-automatically:
-
-```julia
-Zygote.gradient(u0 -> loss(solve(remake(prob; u0), alg; sensealg)), u0)
-```
-
-When calling [`adjoint_sensitivities`](@ref) directly, the forward solve has to be
-run with a *tracked* callback set, which `SciMLSensitivity.track_callbacks` builds
-from the callbacks of the problem:
+automatically; when calling [`adjoint_sensitivities`](@ref) directly, the forward
+solve has to be run with a *tracked* callback set, which
+`SciMLSensitivity.track_callbacks` builds from the callbacks of the problem:
 
 ```julia
 cb = SciMLSensitivity.track_callbacks(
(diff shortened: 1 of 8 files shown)


docs/src/manual/differential_equation_sensitivities.md
```diff
@@ -179,6 +179,48 @@ and DDEs. The continuous adjoint sensitivities `BacksolveAdjoint`, `Interpolatin
...
 
+#### Structured parameters and events
+
+Events are supported for problems whose parameter object is a
+[SciMLStructure](https://github.com/SciML/SciMLStructures.jl) rather than a plain
+vector — most notably the parameter object of a
+[ModelingToolkit.jl](https://docs.sciml.ai/ModelingToolkit/stable/) system with
+`continuous_events` or `discrete_events`. The gradient is taken with respect to
+the tunable portion of the parameters, and an affect that assigns to a parameter
+is handled as long as the parameter object is a mutable SciMLStructure.
```
ChrisRackauckas: Useless comment

docs/src/manual/differential_equation_sensitivities.md
```diff
@@ -179,6 +179,48 @@ and DDEs. The continuous adjoint sensitivities `BacksolveAdjoint`, `Interpolatin
...
+_prob = remake(prob; kwargs = merge(values(prob.kwargs), (; callback = cb)))
+sol = solve(_prob, alg; save_everystep = true)
+du0, dp = adjoint_sensitivities(sol, alg; t = ts, sensealg, dgdu_discrete)
+```
+
+`adjoint_sensitivities` picks the tracked callbacks up from the problem, so
+passing `callback = cb` to it as well is optional. Handing it a solution whose
+callbacks were not tracked throws, because the events would otherwise be ignored
+silently; pass `callback` explicitly to run the adjoint without any event
+correction anyway.
```
ChrisRackauckas: Useless as well, if we just support everything, we don't need to list every option as supported.

src/callback_tracking.jl
```diff
@@ -141,14 +155,102 @@ function (f::TrackedAffect)(integrator, event_idx = nothing)
     end
 end
 
+# A `ContinuousCallback` may be built with `affect_neg! = nothing`, in which case
```
ChrisRackauckas: It can also only have affect_neg

src/callback_tracking.jl
```diff
@@ -141,14 +155,102 @@ function (f::TrackedAffect)(integrator, event_idx = nothing)
...
+
+Whether every affect in `cb` is a [`TrackedAffect`](@ref) (or `nothing`), i.e. the
+callbacks were run through [`track_callbacks`](@ref) on the forward pass and hence
+carry the event times, left limits and event-time correction the adjoint needs.
+"""
+function all_callbacks_tracked(cb)
+    cbs = CallbackSet(cb)
+    return all(is_tracked_callback, cbs.continuous_callbacks) &&
+        all(is_tracked_callback, cbs.discrete_callbacks)
+end
```
ChrisRackauckas, line 197: why is this needed?

src/gauss_adjoint.jl
```diff
@@ -411,82 +420,8 @@ end
...
-            end
-        end
-    else
-        p = deepcopy(CBS.discrete_callbacks[perm].affect!.pleft[getindex.(ts, 1)[perm]])
-    end
-
-    return p
-end
+# `Gaussreset_p` was a verbatim copy of `reset_p`; keep the name for callers.
+const Gaussreset_p = reset_p
```
ChrisRackauckas: just use the new name.

ChrisRackauckas (approved):

Issue #760, opened 2022-12-27
How to 


iliailmer:
I was trying to adapt [this example from tutorials](https://docs.sciml.ai/SciMLSensitivity/dev/ode_fitting/optimization_ode/) so that I can use a more customized data sample. For example, if I have states `x1, x2, x3`, I collect the following sample:

```julia
mq = [y1~x1, y2~x2+x3]
```
then I am trying to use loss as 
```julia
function loss(p)
    sol = solve(prob, Tsit5(), p = p, saveat = tsteps)
    data_true = [data_sample[v.rhs] for v in mq]
    data = [sol[v.rhs] for v in mq] # this is where the code fails
    loss = sum(sum((data[i] .- data_true[i]) .^ 2 for i in eachindex(data)))
    return loss, sol
end
```
but I am getting ` ArgumentError: invalid index: x1(t) of type Term{Real, Base.ImmutableDict{DataType, Any}}` at the line
```julia
 data = [sol[v.rhs] for v in mq]
```

What can be done to fix this?

ChrisRackauckas:
This boils down to a simple issue that observed variables currently do not have adjoints defined. This had a PR that went stale: https://github.com/SciML/SciMLBase.jl/pull/85. We should revive it, and given we need to finish https://github.com/SciML/SciMLBase.jl/pull/342 when Yingbo comes back, @YingboMa let's take a day to solve these two before continuing to other things.

iliailmer:
@ChrisRackauckas @YingboMa Thank you! For now, using numeric indices solves the issue.

BernhardAhrens:
@iliailmer 
Could you give some more details for your workaround?

> For now, using numeric indices solves the issue.

I think I have a similar use case. Thank you!

iliailmer:
@BernhardAhrens sure, if states are `[x1, x2]` and measured quantities are `[x1, x2^2]` then I replace the line

```julia
data = [sol[v.rhs] for v in mq] # this is where the code fails
```

with
 
```julia
data = [sol[1] sol[2]^2]
```

ChrisRackauckas:
https://github.com/SciML/SciMLBase.jl/pull/479 Handles a lot of cases, but `sol[sym]` is still missing an overload.

PR #1662, opened 2026-09-13
Add the observed parameter cotangent in the adjoint backpasses

### LINKED ISSUE #1661: Adjoints drop the parameter cotangent of observed variables
Follow-up to #760. The Zygote rule for `sol[obs]` returns the parameter part of the cotangent as `prob = (p = ...,)`, but the time-dependent backpasses only read `Δ.u`, so a parameter that enters the loss only through an observed variable gets a zero gradient. Only the steady-state backpass reads `Δ.prob.p`.

```julia
using ModelingToolkit, OrdinaryDiffEq, SciMLSensitivity, Zygote, ForwardDiff
using SymbolicIndexingInterface: setp_oop
@variables x(t) y(t)
@parameters α β
@mtkcompile sys = System([D(x) ~ -β * x, y ~ α * x], t)
prob = ODEProblem(sys, [x => 1.0, α => 1.2, β => 2.1], (0.0, 1.0))
set_p = setp_oop(prob, [α, β])
loss(ps) = sum(solve(remake(prob; p = set_p(prob, ps)), Tsit5(); saveat = 0.1, abstol = 1e-8, reltol = 1e-8)[y])
ForwardDiff.gradient(loss, [1.2, 2.1])   # [4.7554, -1.7503]
Zygote.gradient(loss, [1.2, 2.1])[1]     # [0.0, -1.7503], same with InterpolatingAdjoint(autojacvec = ReverseDiffVJP(true))
```

SciMLSensitivity 7.119.7, SciMLBase 3.53.3, ModelingToolkit 11.42.1, Zygote 0.7.13, Julia 1.12.5.

utkuyilmaz1903:
Remaining part of #760; fixes #1661.

SciMLBase's rule for `sol[obs]` returns the observed function's parameter cotangent as `Δ.prob.p`, but the time-dependent backpasses only read `Δ.u`, so a parameter that enters the loss only through an observed variable got a zero gradient. `adjoint_sensitivity_backpass` and the ForwardDiffSensitivity backpass now add it, like `steadystatebackpass` already does; the tangent is written into the parameter structure and projected with `canonicalize(Tunable(), ...)`, which covers the `DespecializedParameters` wrapper `solve` adds. Nothing changes when `Δ.prob.p` is absent.

Still reading only `Δ.u`: the Tracker, ReverseDiff, Enzyme and ForwardSensitivity backpasses, the `diff_tunables = Val(false)` branch, and `sol[obs, j]`, whose rule returns a solution-shaped cotangent instead of a `Tangent`.


ChrisRackauckas-Claude:
Thanks for tracking this down — the `Δ.prob.p` channel is indeed the right place for this. I've taken your commit forward in #1669 (rebased onto master) with the cotangent plumbing consolidated into a shared helper and the test extended to state/mixed/multi-symbol losses across all five sensitivity algorithms. Your authorship is preserved on the original commit.

🤖 Comment by an AI agent (Devin CLI 3000.10.27, model SWE-2 Max) on behalf of Chris Rackauckas.

PR #1669, opened 2026-09-16
Add the observed parameter cotangent in the adjoint backpasses

### LINKED ISSUE #1661: Adjoints drop the parameter cotangent of observed variables

ChrisRackauckas-Claude:
Builds on @utkuyilmaz1903's #1662 (closes #1661), rebased onto master with the cotangent-plumbing consolidated.

## Problem

A loss that reads a parameter only through an observed solution variable (`sol[obs]`) gets a zero gradient: SciMLBase's `sol[obs]` rule puts the observed function's parameter cotangent in `Δ.prob.p`, but the time-dependent backpasses only read `Δ.u`.

## Summary

- Accumulate the explicit parameter cotangent from `Δ.prob.p` into `dp` in `adjoint_sensitivity_backpass` and the `ForwardDiffSensitivity` backpass.
- `tunable_cotangent(p, Δp)` (src/utils.jl) projects a structural parameter cotangent onto the tunable portion of `p`: handles `Tangent`/`NamedTuple`/thunk/`AbstractZero` shapes, `DespecializedParameters` `(params = …)` wrapping, SciML structures (incl. plain arrays), and `Functors` parameters. `solution_parameter_cotangent(Δ)` reads `Δ.prob.p` defensively — a solution-shaped cotangent (e.g. from `sol[x, j]`) carries a `prob` holding parameter *values*, which must not be mistaken for a gradient.
- Consolidates the same projection that was inlined at three existing backpasses (steadystate + the two optimization adjoints) into the shared helper.
- Regression test `test/Core8/observed_parameter.jl`: observed-only, state-only, mixed, and multi-symbol (`sol[[x,y]]`, `sol[[x,y],:]`) losses × default/Interpolating/Quadrature/Gauss/ForwardDiffSensitivity, all checked against ForwardDiff.

## Not covered (same as the original PR)

Tracker, ReverseDiff, and Enzyme backpasses, `ForwardSensitivity`, and `diff_tunables = Val(false)` still read only `Δ.u`.

Found while testing, pre-existing on master and out of scope here:

- `Zygote.gradient` of `sol[sym]` losses on `NonlinearProblem` fails in `adjointdiffcache` (`similar(::Nothing, ::Nothing, ::Nothing)`) — the `NonlinearSolution` indexing rule returns `Δ.u === nothing` (observed) or a solution-shaped cotangent (state), neither of which `df` handles. The default nonlinear adjoint also returns `-0.0` for both. Under investigation separately.
- A mixed loss on one solution object (`sol[x] + sol[y]`) can't be differentiated: the two pullbacks return differently-shaped cotangents that `Zygote.accum` cannot add. `sol[[x, y]]` is the supported way to mix channels.

#### Test plan

- [x] `test/Core8/observed_parameter.jl`: 25/25 pass (5 losses × 5 sensealgs vs ForwardDiff)
- [x] `test/Core8/parameter_initialization.jl` (steadystate backpass regression): 3/3 pass
- [x] `GROUP=QA`: 20/20 pass (Aqua, ExplicitImports, JET)
- [x] `typos` clean on the diff
- [ ] Docs build skipped: internal docstrings only, no `docs/` or public-API changes

🤖 Generated with Devin (Devin CLI 3000.10.27, model: SWE-2 Max). Session transcript (local, no public URL): /home/crackauc/.local/share/devin/cli/summaries/history_ac7d11e5bd2442f6.md


ChrisRackauckas-Claude:
Follow-up filed for the two pre-existing nonlinear-solution gaps noted in the PR body: https://github.com/SciML/SciMLSensitivity.jl/pull/1670 (stacked on this branch) fixes `adjointdiffcache`'s `u0 === nothing` sizing for MTK `NonlinearProblem`s, the `Δ.u === nothing` handling in the steady-state backpasses, and the dropped `Δ.prob.p` in the NLLS backpass. The remaining piece — the silent `-0.0` end-to-end — is upstream: NonlinearSolveBase's `solve_up_adjoint` strips `Tangent{Any}` cotangents to `∂sol.u` before calling the backpass; companion fix at https://github.com/SciML/NonlinearSolve.jl/pull/1294. Merge order: this PR, then #1670, then the upstream one-liner.

